PR #5, opened 2026-03-20
refactor(bridge): replace catch-all proxy with explicit route handlers

AlexDochioiu:
## Summary

- Replace the `ProxyHandler` catch-all with 17 explicit, fully-typed route handlers — every mobile→bridge request is now handled explicitly, unknown routes return 404
- Add 12 new `BridgePlugin` interface methods with typed models (`PluginSessionStatus`, `PluginAgent`, `PluginPendingQuestion`, `PluginPromptPart`)
- Add 11 new `OpenCodeApi` HTTP methods + `PendingQuestion` model in the OpenCode plugin
- Remove deprecated `proxyRequest` from `BridgePlugin` interface and `OpenCodePlugin` implementation
- 60+ new test cases (TDD) covering all 12 new handlers

## Routes Now Explicitly Handled

| Method | Path | Handler |
|--------|------|---------|
| POST | /session | CreateSessionHandler |
| PATCH | /session/:id | UpdateSessionArchiveStatusHandler |
| DELETE | /session/:id | DeleteSessionHandler |
| GET | /session/:id/children | GetChildSessionsHandler |
| GET | /session/status | GetSessionStatusesHandler |
| POST | /session/:id/prompt_async | SendPromptHandler |
| POST | /session/:id/abort | AbortSessionHandler |
| GET | /agent | GetAgentsHandler |
| GET | /question | GetPendingQuestionsHandler |
| POST | /question/:id/reply | ReplyToQuestionHandler |
| POST | /question/:id/reject | RejectQuestionHandler |
| GET | /project/current | GetCurrentProjectHandler |

## Breaking Changes

- `proxyRequest` removed from `BridgePlugin` interface (was already `@Deprecated`)
- Unknown routes now return 404 instead of being proxied

## Verification

- `make analyze` — 0 issues across all 3 bridge modules
- `make test` — 259 tests pass (65 opencode plugin + 194 app)


bridge/app/lib/src/bridge/routing/create_session_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  @override
+  Future<RelayResponse> handle(
+    RelayRequest request, {
+    required Map<String, String> pathParams,
+    required Map<String, String> queryParams,
+    String? fragment,
+  }) async {
+    final worktree = findHeader(request.headers, "x-opencode-directory");
+    if (worktree == null || worktree.isEmpty) {
```
AlexDochioiu: This should work based on session id. Create a session for a given session id. The id should be part of the req body. Create a new shared model for the body.

bridge/app/lib/src/bridge/routing/create_session_handler.dart
```diff
@@ -0,0 +1,53 @@
...
+        null => null,
+      },
+      summary: switch (created.summary) {
+        PluginSessionSummary(:final additions, :final deletions, :final files) => SessionSummary(
+          additions: additions,
+          deletions: deletions,
+          files: files,
+        ),
+        null => null,
+      },
```
AlexDochioiu, line 62: since this handler creates a new session, just put this as `null` to avoid sending pointless stuff.

bridge/app/lib/src/bridge/routing/delete_session_handler.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  @override
+  Future<RelayResponse> handle(
+    RelayRequest request, {
+    required Map<String, String> pathParams,
+    required Map<String, String> queryParams,
+    String? fragment,
+  }) async {
+    final sessionId = pathParams[_idParam]!;
+    await _plugin.deleteSession(sessionId);
```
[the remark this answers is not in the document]
AlexDochioiu: We should treat this 404 as success, since it means the session was not found, so for the sake of the goal, "session is deleted"

bridge/app/lib/src/bridge/routing/get_child_sessions_handler.dart
```diff
@@ -0,0 +1,59 @@
...
+  GetChildSessionsHandler(this._plugin) : super(HttpMethod.get, "/session/:$_idParam/children");
+
+  @override
+  Future<RelayResponse> handle(
+    RelayRequest request, {
+    required Map<String, String> pathParams,
+    required Map<String, String> queryParams,
+    String? fragment,
+  }) async {
+    final sessionId = pathParams[_idParam]!;
```
AlexDochioiu: instead of force casting, add a line after where we throw if this is null or empty.

bridge/app/lib/src/bridge/routing/get_current_project_handler.dart
```diff
@@ -0,0 +1,42 @@
...
+import "request_handler.dart";
+
+/// Handles `GET /project/current` — returns project for a given worktree.
+class GetCurrentProjectHandler extends RequestHandler {
+  final BridgePlugin _plugin;
+
+  GetCurrentProjectHandler(this._plugin) : super(HttpMethod.get, "/project/current");
+
+  @override
+  Future<RelayResponse> handle(
```
AlexDochioiu: Is this handler and `/project/current` used by mobile? If not, we don't need a handler and a plugin method for `_plugin.getCurrentProject`

bridge/app/lib/src/bridge/routing/get_pending_questions_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "request_handler.dart";
+
+/// Handles `GET /question` — returns all pending question prompts.
+class GetPendingQuestionsHandler extends RequestHandler {
+  final BridgePlugin _plugin;
+
+  GetPendingQuestionsHandler(this._plugin) : super(HttpMethod.get, "/question");
```
AlexDochioiu, line 15: for what session? How is opencode knowing which session this is for?

bridge/app/lib/src/bridge/routing/get_session_statuses_handler.dart
```diff
@@ -0,0 +1,41 @@
...
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "request_handler.dart";
+
+/// Handles `GET /session/status` — returns statuses for sessions.
+class GetSessionStatusesHandler extends RequestHandler {
+  final BridgePlugin _plugin;
+
+  GetSessionStatusesHandler(this._plugin) : super(HttpMethod.get, "/session/status");
```
AlexDochioiu, line 14: for what session?

bridge/app/lib/src/bridge/routing/reject_question_handler.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import "request_handler.dart";
+
+const _idParam = "id";
+
+/// Handles `POST /question/:id/reject` — rejects a pending question.
+class RejectQuestionHandler extends RequestHandler {
+  final BridgePlugin _plugin;
+
+  RejectQuestionHandler(this._plugin) : super(HttpMethod.post, "/question/:$_idParam/reject");
```
AlexDochioiu, line 14: do we care about the active session/etc?

bridge/app/lib/src/bridge/routing/reply_to_question_handler.dart
```diff
@@ -0,0 +1,50 @@
...
+      replyRequest = ReplyToQuestionRequest.fromJson(
+        jsonDecode(bodyRaw) as Map<String, dynamic>,
+      );
+    } on FormatException {
+      return buildErrorResponse(request, 400, "invalid JSON body");
+    } on Object {
+      return buildErrorResponse(request, 400, "invalid JSON body");
+    }
+
+    await _plugin.replyToQuestion(questionId, answers: replyRequest.answers);
```
AlexDochioiu: is the question id global? or is opencode relying on some active session thing tracked internally?

shared/sesori_shared/lib/src/models/sesori/update_session_archive_request.dart
```diff
@@ -0,0 +1,24 @@
+import "package:freezed_annotation/freezed_annotation.dart";
+
+part "update_session_archive_request.freezed.dart";
+
+part "update_session_archive_request.g.dart";
+
+@Freezed(fromJson: true, toJson: true)
+sealed class UpdateSessionArchiveRequest with _$UpdateSessionArchiveRequest {
+  const factory UpdateSessionArchiveRequest({
+    @JsonKey(required: true) required UpdateSessionArchiveTime time,
```
AlexDochioiu: Don't ever use `JsonKey(required: ...)`. Just use the dart `required`. This line should be:
`required UpdateSessionArchiveTime time,`

Note: even for nullable entries still keep the dart `requried` keyword, such as `required int? archived` below.

Update this for the entire repo/pr

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -66,9 +66,9 @@ class SessionService {
       "/session/$sessionId",
       // ignore: no_slop_linter/avoid_dynamic_type, json parsing
       fromJson: (json) => Session.fromJson(json as Map<String, dynamic>),
-      body: {
-        "time": {"archived": DateTime.now().millisecondsSinceEpoch},
-      },
+      body: UpdateSessionArchiveRequest(
+        time: UpdateSessionArchiveTime(archived: DateTime.now().millisecondsSinceEpoch),
+      ).toJson(),
```
AlexDochioiu: mobile should only send `UpdateSessionArchiveRequest` containing `archived: true/false`. It's not mobile's responsibility to determine the time at which it was archived. Each plugin should do this individually.

P.s. sending false would un-archive.

shared/sesori_shared/lib/src/models/sesori/reply_to_question_request.dart
```diff
@@ -0,0 +1,14 @@
+import "package:freezed_annotation/freezed_annotation.dart";
+
+part "reply_to_question_request.freezed.dart";
+
+part "reply_to_question_request.g.dart";
+
+@Freezed(fromJson: true, toJson: true)
+sealed class ReplyToQuestionRequest with _$ReplyToQuestionRequest {
+  const factory ReplyToQuestionRequest({
+    required List<List<String>> answers,
```
AlexDochioiu: This type is weird, and I cannot figure out what is meant to be. Why not simple List<String> which is one answer per question? or is some other data passed too? If you pass other data, change this to some custom type such as `List<ReplyAnswer> answers` and `ReplyAnswer` can have multiple named fields to understand what is sent here based on the type.

shared/sesori_shared/lib/src/models/sesori/send_prompt_request.dart
```diff
@@ -0,0 +1,36 @@
...
+
+@Freezed(fromJson: true, toJson: true)
+sealed class PromptPart with _$PromptPart {
+  const factory PromptPart({
+    required String type,
+    String? text,
+  }) = _PromptPart;
+
+  factory PromptPart.fromJson(Map<String, dynamic> json) => _$PromptPartFromJson(json);
+}
```
AlexDochioiu, line 24: turn this into freezed with union type. For now only text is supported `PromptPart.text` and has `required String text` inside. eventually we'll add other types.

bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
```diff
@@ -0,0 +1,69 @@
...
+      return buildErrorResponse(request, 400, "invalid JSON body");
+    } on Object {
+      return buildErrorResponse(request, 400, "invalid JSON body");
+    }
+
+    final archived = archiveRequest.time.archived;
+
+    final updated = await _plugin.updateSessionArchiveStatus(
+      sessionId,
+      archivedAt: archived,
```
AlexDochioiu: send only archived true/false. internally it can manage its own archived time.

bridge/sesori_plugin_interface/lib/src/models/plugin_agent.dart
```diff
@@ -0,0 +1,32 @@
...
+  }) = _PluginAgentModel;
+}
+
+@freezed
+sealed class PluginAgent with _$PluginAgent {
+  const factory PluginAgent({
+    required String name,
+    required String? description,
+    required PluginAgentModel? model,
+    required String? variant,
```
AlexDochioiu: turn this into enum. those are the current options:
```
none, minimal, low, medium, high, xhigh
```

not relevant here I guess but not all models support all variants

bridge/sesori_plugin_interface/lib/src/models/plugin_prompt_part.dart
```diff
@@ -0,0 +1,13 @@
...
+
+part "plugin_prompt_part.g.dart";
+
+@freezed
+sealed class PluginPromptPart with _$PluginPromptPart {
+  const factory PluginPromptPart({
+    required String type,
+    String? text,
+  }) = _PluginPromptPart;
+}
```
AlexDochioiu, line 10: use untion type and add `text` as the only initial type. make `required String text` since it makes no sense to have it missing.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -18,9 +22,39 @@ abstract class BridgePlugin {
   /// Get sessions for a worktree directory.
   Future<List<PluginSession>> getSessions(String worktree, {int? start, int? limit});
 
+  Future<PluginSession> createSession(String worktree);
```
AlexDochioiu: projectId instead of `worktree`

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -18,9 +22,39 @@ abstract class BridgePlugin {
...
+
+  Future<List<PluginAgent>> getAgents();
+
+  Future<List<PluginPendingQuestion>> getPendingQuestions();
+
+  Future<void> replyToQuestion(String questionId, {required List<List<String>> answers});
+
+  Future<void> rejectQuestion(String questionId);
+
+  Future<PluginProject> getCurrentProject(String worktree);
```
AlexDochioiu: projectId not worktree

bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
```diff
@@ -309,6 +387,60 @@ class OpenCodePlugin implements BridgePlugin {
...
+        next: next,
+      ),
+    };
+  }
+
+  PluginAgent _mapAgent(AgentInfo a) {
+    return PluginAgent(
+      name: a.name,
+      description: a.description,
+      model: switch (a.model) {
```
AlexDochioiu: Better put those mappers as some extension functions or similar. Maybe even in the same file each as the local model owned by the plugin

bridge/app/lib/src/bridge/routing/get_agents_handler.dart
```diff
@@ -31,7 +31,15 @@ class GetAgentsHandler extends RequestHandler {
...
-            variant: a.variant,
+            variant: switch (a.variant) {
+              PluginAgentVariant.none => "none",
+              PluginAgentVariant.minimal => "minimal",
+              PluginAgentVariant.low => "low",
+              PluginAgentVariant.medium => "medium",
+              PluginAgentVariant.high => "high",
+              PluginAgentVariant.xhigh => "xhigh",
+              null => null,
+            },
```
AlexDochioiu: add inside the enum a `final String safeName` or similar and this becomes a.variant?.variantName.

Do not use use the `.name` added automatically to enums because this would break due to obfuscation.

bridge/app/lib/src/bridge/routing/reply_to_question_handler.dart
```diff
@@ -29,16 +34,23 @@ class ReplyToQuestionHandler extends RequestHandler {
...
     } on FormatException {
       return buildErrorResponse(request, 400, "invalid JSON body");
     } on Object {
       return buildErrorResponse(request, 400, "invalid JSON body");
     }
 
-    await _plugin.replyToQuestion(questionId, answers: replyRequest.answers);
+    await _plugin.replyToQuestion(
+      questionId,
+      answers: replyRequest.answers.map((answer) => answer.values).toList(),
```
AlexDochioiu, line 52: Make sure to document for the `replyToQuestion` plugin interface method that the `answers` is List<List<String>> because we may send answers to multiple questions at once, and also each question answered may contain multiple answers (multiple choice)

bridge/app/test/bridge/routing/reply_to_question_handler_test.dart
```diff
@@ -30,8 +30,8 @@ void main() {
           body: jsonEncode(
             const ReplyToQuestionRequest(
               answers: [
-                ["yes"],
-                ["tool-a", "tool-b"],
+                ReplyAnswer(values: ["yes"]),
+                ReplyAnswer(values: ["tool-a"]),
```
AlexDochioiu: Keep here `["tool-a", "tool-b"],` to test that multiple answers per question still works well

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -22,9 +22,9 @@ abstract class BridgePlugin {
   /// Get sessions for a worktree directory.
   Future<List<PluginSession>> getSessions(String worktree, {int? start, int? limit});
 
-  Future<PluginSession> createSession(String worktree);
+  Future<PluginSession> createSession({required String projectId, required String sessionId});
```
AlexDochioiu: It doesn't make sense to send a `sessionId` here? Since it's a new session we can't know its id yet.

Unless this is to create a sub-session? In which case rename the field to `required String? parentSessionId`

bridge/sesori_plugin_opencode/lib/src/models/agent_info.dart
```diff
@@ -32,3 +33,39 @@ sealed class AgentModel with _$AgentModel {
...
+PluginAgentVariant? _parsePluginAgentVariant(String? value) {
+  return switch (value) {
+    "none" => PluginAgentVariant.none,
+    "minimal" => PluginAgentVariant.minimal,
+    "low" => PluginAgentVariant.low,
+    "medium" => PluginAgentVariant.medium,
+    "high" => PluginAgentVariant.high,
+    "xhigh" => PluginAgentVariant.xhigh,
+    _ => null,
+  };
```
AlexDochioiu: This should be a static function / factory / whatever in `PluginAgentVariant`

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -66,7 +66,7 @@ class OpenCodeApi {
           "content-type": "application/json",
           "x-opencode-directory": directory,
         },
-        body: "{}",
+        body: jsonEncode({"id": sessionId}),
```
AlexDochioiu: what does this change do? seems like a mistake?

shared/sesori_shared/lib/src/models/sesori/create_session_request.dart
```diff
@@ -0,0 +1,12 @@
+import "package:freezed_annotation/freezed_annotation.dart";
+
+part "create_session_request.freezed.dart";
+
+part "create_session_request.g.dart";
+
+@Freezed(fromJson: true, toJson: true)
+sealed class CreateSessionRequest with _$CreateSessionRequest {
+  const factory CreateSessionRequest({required String id}) = _CreateSessionRequest;
```
AlexDochioiu: maybe this should be `projectId` rather than just `id`. I guess this is what the id is, right?  if we support child sessions, you can also include an optional `required String? parentSessionId`

bridge/app/lib/src/bridge/routing/get_current_project_handler.dart
```diff
@@ -18,9 +18,9 @@ class GetCurrentProjectHandler extends RequestHandler {
     required Map<String, String> queryParams,
     String? fragment,
   }) async {
-    final projectId = findHeader(request.headers, "x-opencode-directory");
+    final projectId = queryParams["projectId"];
```
AlexDochioiu: use a static const _projectIdQueryKey or similar

bridge/app/lib/src/bridge/routing/get_sessions_handler.dart
```diff
@@ -21,20 +21,20 @@ class GetSessionsHandler extends RequestHandler {
     required Map<String, String> queryParams,
     String? fragment,
   }) async {
-    final worktree = findHeader(request.headers, "x-opencode-directory");
-    if (worktree == null || worktree.isEmpty) {
+    final projectId = queryParams["projectId"];
```
AlexDochioiu: use a static const _projectIdQueryKey or similar

bridge/app/lib/src/bridge/routing/get_current_project_handler.dart
AlexDochioiu: I also have some doubts about putting this a query param. What if it contains invalid chars/etc? Let's change to x-project-id header

bridge/app/lib/src/bridge/routing/get_sessions_handler.dart
AlexDochioiu: I also have some doubts about putting this a query param. What if it contains invalid chars/etc? Let's change to x-project-id header

AlexDochioiu: make sure mobile always sends this for this request

AlexDochioiu: make sure mobile always sends this for this request

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -56,17 +56,21 @@ class OpenCodeApi {
...
+      }
       final response = await client.post(
         Uri.parse("$serverURL/session"),
         headers: {
           ..._authHeaders,
           "content-type": "application/json",
           "x-opencode-directory": directory,
         },
-        body: jsonEncode({"id": sessionId}),
+        body: jsonEncode(body),
```
AlexDochioiu, line 85: Are you sure this is how you start a sub-session? Did you look at opencode's repo/docs?

shared/sesori_shared/lib/src/models/sesori/create_session_request.dart
```diff
@@ -7,8 +7,8 @@ part "create_session_request.g.dart";
 @Freezed(fromJson: true, toJson: true)
 sealed class CreateSessionRequest with _$CreateSessionRequest {
   const factory CreateSessionRequest({
-    required String id,
     required String projectId,
+    String? parentSessionId,
```
AlexDochioiu: `required String? parentSessionId`

mobile/module_core/lib/src/capabilities/project/project_service.dart
```diff
@@ -23,11 +23,12 @@ class ProjectService {
...
 
-  /// Returns the project the server resolved for the current request directory.
-  Future<ApiResponse<Project>> getCurrentProject() {
+  /// Returns the project matching the requested project ID.
+  Future<ApiResponse<Project>> getCurrentProject({required String projectId}) {
     return _client.get(
       "/project/current",
       fromJson: (json) => Project.fromJson(json as Map<String, dynamic>),
+      queryParameters: {"projectId": projectId},
     );
```
AlexDochioiu, line 40: opencode API doesn't seem to take a "projectId" at all for getCurrent. It looks like `getCurrentProject` does something different. It's not a `getProject` by ID

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -15,12 +19,53 @@ abstract class BridgePlugin {
...
+  /// [answers] is a `List<List<String>>` because:
+  /// - The outer list contains one entry per question in the prompt
+  ///   (a single prompt can ask multiple questions at once).
+  /// - Each inner list contains the selected answers for that question
+  ///   (supports multi-select — one or more values can be chosen).
+  Future<void> replyToQuestion(String questionId, {required List<List<String>> answers});
+
+  Future<void> rejectQuestion(String questionId);
+
+  Future<PluginProject> getCurrentProject(String projectId);
```
AlexDochioiu: `getCurrentProject` should actually be `getProject` and take a projectId (as it does now).

Not sure when/if we need this at all.
NOTE: get current project (from opencode) returns the project at which the opencode server was started. this is **useless for the app**.

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -65,6 +134,96 @@ class OpenCodeApi {
...
+        headers: _authHeaders,
+        body: "",
+      );
+      _ensureSuccess(response, "POST /question/$questionId/reject");
+    } finally {
+      client.close();
+    }
+  }
+
+  Future<Project> getCurrentProject(String directory) async {
```
AlexDochioiu: rename to `getProject`

mobile/module_core/lib/src/cubits/session_list/session_list_cubit.dart
```diff
@@ -20,7 +20,7 @@ class SessionListCubit extends Cubit<SessionListState> {
   final ProjectService _projectService;
   final SseEventRepository _sseEventRepository;
   final String _projectId;
-  final String _worktree;
+  String? _resolvedWorktree;
```
AlexDochioiu: forget about the stupid worktree, you don't need it here.

mobile/module_core/lib/src/cubits/session_list/session_list_cubit.dart
```diff
@@ -72,7 +60,7 @@ class SessionListCubit extends Cubit<SessionListState> {
   void _onSessionActivityUpdated(Map<String, Set<String>> activityById) {
     if (isClosed) return;
     if (state is! SessionListLoaded) return;
-    final activeIds = activityById[_worktree] ?? {};
+    final activeIds = _resolvedWorktree != null ? (activityById[_resolvedWorktree] ?? <String>{}) : <String>{};
```
AlexDochioiu: `activityById` should be activity by project id, not by worktree. Rename everywhere to be activityByProjectId

mobile/module_core/lib/src/cubits/session_list/session_list_cubit.dart
```diff
@@ -330,19 +320,24 @@ class SessionListCubit extends Cubit<SessionListState> {
...
-        if (!_allSessions.any(_belongsHere)) {
-          final projectResponse = await _projectService.getCurrentProject(projectId: _projectId);
-          if (isClosed) return false;
-
-          if (projectResponse case SuccessResponse(data: final current) when current.id != _projectId) {
-            logw("Stale project: expected $_projectId, server resolved ${current.id}");
-            emit(SessionListState.staleProject(resolvedProjectId: current.id));
-            return true;
-          }
+        // Resolve worktree from project service for activity tracking.
```
AlexDochioiu: you DO NOT NEED TO KNOW ABOUT WORKTREES. THE APP DOES NOT EVER NEED TO KNOW ABOUT WORKTREES

PR #15, opened 2026-03-23
feat(mobile): auto-refresh project list with route visibility gating and timestamp ticker

AlexDochioiu:
## Summary

- **Route-level visibility abstraction**: New `RouteSource` interface in `module_core` (pure Dart) exposes the currently active `AppRoute` as a `ValueStream`. `GoRouterRouteSource` in `app/` provides the Flutter implementation by listening to GoRouter navigation changes and matching locations to `AppRoute` via regex.
- **Throttled project data refresh**: `ProjectListCubit` now auto-refreshes project data when SSE activity events arrive, but only when the project list is the visible page. Uses a 30-second trailing-edge throttle to coalesce rapid events during active sessions (idle→busy→idle per message). Events while the page is not visible are deferred — one immediate refresh fires when the user navigates back.
- **Periodic timestamp ticker**: `_ProjectListBody` converted to `StatefulWidget` with a 60-second `Timer.periodic` that forces rebuilds, keeping relative timestamps ("5m ago") fresh without relying on state changes.

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Session activity while on project list | Stale timestamps, no refresh | Auto-refresh after 30s throttle |
| Session activity while on session detail | Wasted API calls (if we had auto-refresh) | No refresh — deferred until user returns |
| User navigates back to project list | Stale until pull-to-refresh | Immediate refresh if events were pending |
| "Updated 5m ago" text over time | Stays "5m ago" forever | Recomputes every 60s ("6m ago", "7m ago"...) |
| 10 rapid SSE events in 5 seconds | Would have been 10 API calls | Coalesced into 1 call after 30s |

## New files

| File | Purpose |
|------|---------|
| `module_core/.../platform/route_source.dart` | Abstract interface — mirrors `LifecycleSource` pattern |
| `app/.../platform/go_router_route_source.dart` | Flutter implementation — `@Singleton(as: RouteSource)` |

## Modified files

| File | Change |
|------|--------|
| `module_core/.../project_list_cubit.dart` | Added `RouteSource` dependency, 30s throttle, route visibility gating |
| `app/.../project_list_screen.dart` | Pass `RouteSource` to cubit, 60s rebuild timer for timestamps |
| `app/.../di/injection.config.dart` | Regenerated — registers `GoRouterRouteSource` |
| `app/test/.../test_helpers.dart` | Added `MockRouteSource` |
| `app/test/.../project_list_cubit_test.dart` | 6 new fakeAsync throttle tests + updated constructor in existing tests |

## Verification

- `dart analyze` — clean in both `module_core` and `app`
- `dart test` — 40 module_core tests pass
- `flutter test` — all app tests pass (18 cubit tests including 6 new)

### DIFF
--- a/mobile/app/lib/core/platform/go_router_route_source.dart
+++ b/mobile/app/lib/core/platform/go_router_route_source.dart
@@ -1,0 +1,62 @@
+import "dart:async";
+
+import "package:get_it/get_it.dart";
+import "package:injectable/injectable.dart";
+import "package:rxdart/rxdart.dart";
+import "package:sesori_dart_core/sesori_dart_core.dart";
+
+import "../routing/app_router.dart";
+
+@Singleton(as: RouteSource)
+class GoRouterRouteSource implements RouteSource, Disposable {
+  final BehaviorSubject<AppRoute?> _currentRouteStream;
+
+  GoRouterRouteSource()
+    : _currentRouteStream = BehaviorSubject.seeded(
+        _matchRoute(appRouter.routerDelegate.currentConfiguration.uri.path),
+      ) {
+    appRouter.routerDelegate.addListener(_onRouteChanged);
+  }
+
+  @override
+  ValueStream<AppRoute?> get currentRouteStream => _currentRouteStream.stream;
+
+  @override
+  FutureOr<void> onDispose() {
+    appRouter.routerDelegate.removeListener(_onRouteChanged);
+    _currentRouteStream.close();
+  }
+
+  void _onRouteChanged() {
+    final matchedRoute = _matchRoute(appRouter.routerDelegate.currentConfiguration.uri.path);
+    if (_currentRouteStream.valueOrNull == matchedRoute) {
+      return;
+    }
+    _currentRouteStream.add(matchedRoute);
+  }
+
+  static AppRoute? _matchRoute(String path) {
+    final orderedRoutes = AppRoute.values.toList()..sort((a, b) => b.path.length.compareTo(a.path.length));
+
+    for (final route in orderedRoutes) {
+      if (_routeRegex(route).hasMatch(path)) {
+        return route;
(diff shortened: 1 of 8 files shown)


mobile/app/lib/core/platform/go_router_route_source.dart
```diff
@@ -0,0 +1,62 @@
...
+  void _onRouteChanged() {
+    final matchedRoute = _matchRoute(appRouter.routerDelegate.currentConfiguration.uri.path);
+    if (_currentRouteStream.valueOrNull == matchedRoute) {
+      return;
+    }
+    _currentRouteStream.add(matchedRoute);
+  }
+
+  static AppRoute? _matchRoute(String path) {
+    final orderedRoutes = AppRoute.values.toList()..sort((a, b) => b.path.length.compareTo(a.path.length));
```
AlexDochioiu: We don't need to compute this every time routes change. This can even be a private static final _orderedRoutes that we just access.

actually, do we need this sorting? Since go router probably checks them in the enum order, maybe so should we?

mobile/module_core/lib/src/cubits/project_list/project_list_cubit.dart
```diff
@@ -38,12 +56,58 @@ class ProjectListCubit extends Cubit<ProjectListState> {
...
+
+    // Always update activity badges immediately — this is cheap (no API call).
     emit(
       ProjectListState.loaded(
         projects: (state as ProjectListLoaded).projects,
         activityById: activityById,
       ),
     );
+
+    // Schedule a throttled project data refresh so that timestamps update too.
```
AlexDochioiu: I was under the impression timestamp update is managed by the UI timer trigger a re-build every minute. This seems like duplicated handling now.

PR #17, opened 2026-03-23
feat(bridge): add token refresh for push notifications

AlexDochioiu:
## Summary

- Fix push notification 401 errors by adding token lifecycle management to the bridge's push notification path
- Add `TokenRefreshManager` with singleflight deduplication, proactive JWT-based TTL checking (soft 90s / hard 30s thresholds), and reactive 401 retry
- Fix the "double-swallow" error handling pattern where `PushNotificationClient` caught all errors AND `PushNotificationService.catchError` was dead code

## Problem

Push notifications from the bridge fail with `[push] notification failed: 401` because the access token expires mid-session and is never refreshed. Token refresh only triggers on relay reconnect — completely unrelated timing to push notification sends.

### New: `TokenRefreshManager` (singleflight + proactive/reactive refresh)
- **Proactive**: Parses JWT `exp` claim to check TTL before sending. Background refresh at <90s, blocking refresh at <30s.
- **Reactive**: On 401, force-refreshes the token and retries once.
- **Singleflight**: Concurrent refresh requests are coalesced — only 1 HTTP call to `/auth/refresh` regardless of how many push notifications hit 401 simultaneously.
- **Persistence**: Refreshed tokens are persisted to disk (preserving `bridgeToken`), so the next relay reconnect doesn't load stale tokens.

### Modified: `PushNotificationClient`
- Uses `TokenRefreshManager.getFreshAccessToken()` for proactively-fresh tokens
- On 401: calls `getFreshAccessToken(forceRefresh: true)`, retries once
- **Now throws on failure** instead of silently swallowing all errors

### Modified: `PushNotificationService`
- `catchError` is now alive (was dead code before since client swallowed everything)
- Differentiates auth failures (`Log.e`) from transient errors (`Log.w`)
- `unawaited()` preserved — push delivery does not block the SSE event loop

## Files Changed

| File | Change |
|------|--------|
| `bridge/app/lib/src/auth/jwt_expiry.dart` | New — JWT `exp` claim parser |
| `bridge/app/lib/src/auth/token_refresh_manager.dart` | New — singleflight token refresh |
| `bridge/app/lib/src/auth/token_refresher.dart` | New — interface for testability |
| `bridge/app/lib/src/push/push_notification_client.dart` | Modified — 401 retry, throws on failure |
| `bridge/app/lib/src/push/push_notification_service.dart` | Modified — differentiated error logging |
| `bridge/app/bin/bridge.dart` | Modified — DI wiring |

## Test Coverage

- **27 new tests** across 3 test files (JWT parser: 8, TokenRefreshManager: 12, PushNotificationClient: 7)
- All 236 workspace tests pass (`make test`)
- Zero analysis issues (`make analyze`)

## Scope

- Bridge-only — no changes to `shared/sesori_shared/` or `mobile/`
- No changes to `AccessTokenService` interfaces or `Orchestrator._refreshAccessToken()`
- No changes to `PushRateLimiter`


shared/sesori_shared/lib/src/auth/jwt_expiry.dart
```diff
@@ -0,0 +1,34 @@
+import "dart:convert";
+
+/// Parses a JWT token and extracts the expiry time from the `exp` claim.
+///
+/// Returns a [DateTime] in UTC if the token is valid and contains an `exp` claim,
+/// or `null` if the token is invalid or missing the `exp` claim.
+///
+/// The function never throws — all errors are caught and return `null`.
+DateTime? parseJwtExpiry(String token) {
```
AlexDochioiu, line 9: at least this can easily be put in the shared repo and re-used by both mobile and bridge

bridge/app/lib/src/auth/token_refresher.dart
```diff
@@ -0,0 +1,3 @@
+abstract interface class TokenRefresher {
+  Future<String> getFreshAccessToken({bool forceRefresh = false});
```
AlexDochioiu: `getAccessToken` is a better name. the word `fresh` implies we fetch it always or similar

bridge/app/bin/bridge.dart
```diff
@@ -152,9 +153,16 @@ Future<void> main(List<String> args) async {
     password: serverPasswordPtr,
   );
   final accessTokenService = AccessTokenService(authTokens.accessToken);
+  final tokenRefreshManager = TokenRefreshManager(
```
AlexDochioiu: Should we keep `AccessTokenService` and `TokenRefreshManager` as two separate classes? I feel like it makes sense to merge them as a single class. Won't it make stuff more readable?

shared/sesori_shared/lib/src/auth/jwt_expiry.dart
AlexDochioiu: Done — moved `parseJwtExpiry` to `shared/sesori_shared/lib/src/auth/jwt_expiry.dart` and exported from the barrel file. Both bridge and mobile now use the shared version. Removed the duplicate static method from mobile's `TokenStorageService`.

bridge/app/lib/src/auth/token_refresher.dart
AlexDochioiu: Done — renamed to `getAccessToken` across the `TokenRefresher` interface, `TokenManager` implementation, `PushNotificationClient`, and all tests.

bridge/app/bin/bridge.dart
AlexDochioiu: Done — merged into a single `TokenManager` class that implements `AccessTokenProvider`, `AccessTokenUpdater`, and `TokenRefresher`. It owns the `BehaviorSubject` for token state + all refresh logic. Deleted `AccessTokenService`. DI in `bridge.dart` is now one object passed to push client, relay client, and orchestrator.

PR #23, opened 2026-03-24
feat(bridge): replace per-message push notifications with state-based session completion

AlexDochioiu:
## Summary

- Replace noisy per-message push notifications with a single "session completed" notification that fires when the AI finishes all work (main agent + all subtasks idle)
- Add `PushSessionStateTracker` to track session statuses, parent-child relationships, pending questions/permissions, session titles, and latest assistant text
- Refactor `PushNotificationService` from stateless to stateful with 500ms debounce, group-idle detection, and question/permission suppression
- Notification content now shows truncated session title + ~10 words from AI reply instead of generic "New session message"

### Shared (`sesori_shared`)
- Add `NotificationEventType.sessionCompleted` enum value

### Bridge
- **New**: `PushSessionStateTracker` — pure state machine tracking session lifecycle events for notification decisions
- **Refactored**: `PushNotificationService` — now stateful with:
  - 500ms debounce after busy→idle transition before sending completion notification
  - Parent-child group awareness (only fires when root + all children idle)
  - Pending question/permission suppression (defers to existing `aiInteraction` notification)
  - `dispose()` / `reset()` lifecycle methods for timer management
  - Content formatting: truncated session title (50 chars) + message snippet (10 words) with fallbacks
- **Orchestrator**: `reset()` on reconnect, `dispose()` on shutdown
- **Removed**: `SesoriMessageUpdated` notification mapping (no more per-message notifications)

### Tests
- 58 tests covering all acceptance criteria and edge cases
- `fakeAsync` for deterministic timer testing
- Hand-written fakes (no mockito)

## What This Solves
1. **User's own messages no longer trigger notifications** — `SesoriMessageUpdated` mapping removed entirely
2. **AI responses don't spam** — one notification when done, not per-message
3. **Subtasks respected** — waits for main agent + all child sessions to finish
4. **No duplicate with question/permission** — if AI is asking a question, only the question notification fires
5. **Friendly content** — session title + message preview instead of "New session message" / "assistant updated message m-123"

### DIFF
--- a/bridge/app/lib/src/push/completion_notifier.dart
+++ b/bridge/app/lib/src/push/completion_notifier.dart
@@ -7,33 +7,36 @@ import "push_session_state_tracker.dart";
 /// Manages debounced "session completed" notifications.
 ///
 /// Tracks when session groups transition from busy to idle and fires
-/// a callback after a debounce period, provided no pending interactions
+/// a stream event after a debounce period, provided no pending interactions
 /// (questions/permissions) block it.
 class CompletionNotifier {
   final PushSessionStateTracker _tracker;
-  final void Function(String rootSessionId) _onCompletion;
   final Duration _debounceDuration;
+  final StreamController<String> _completionController = StreamController<String>.broadcast();
   final Map<String, Timer> _debounceTimers = {};
   final Set<String> _completionSentForRoots = {};
   final Map<String, String> _permissionRequestToSession = {};
 
+  Stream<String> get completions => _completionController.stream;
+
   CompletionNotifier({
     required PushSessionStateTracker tracker,
-    required void Function(String rootSessionId) onCompletion,
     Duration debounceDuration = const Duration(milliseconds: 500),
   }) : _tracker = tracker,
-       _onCompletion = onCompletion,
        _debounceDuration = debounceDuration;
 
   /// Process an SSE event for completion tracking.
   /// Call AFTER tracker.handleEvent() has already updated state.
   void handleEvent(SesoriSseEvent event) {
     switch (event) {
+      // A new question blocks completion for this session group.
       case SesoriQuestionAsked(:final sessionID):
         _cancelDebounceForSessionGroup(sessionID);
+      // A new permission request blocks completion for this session group.
       case SesoriPermissionAsked(:final requestID, :final sessionID):
         _permissionRequestToSession[requestID] = sessionID;
         _cancelDebounceForSessionGroup(sessionID);
+      // Session status transitions determine when completion can fire.
       case SesoriSessionStatus(:final sessionID, :final status):
         final rootSessionId = _tracker.resolveRootSessionId(sessionID);
         switch (status) {
@@ -44,39 +47,53 @@ class CompletionNotifier {
           case SessionStatusIdle():
             _maybeScheduleCompletion(sessionID);
         }
+      // Deletion clears any pending completion work for this session group.
       case SesoriSessionDeleted(:final info):
         _cancelDebounceForSessionGroup(info.id);
         _completionSentForRoots.remove(info.id);
         _permissionRequestToSession.removeWhere((_, sessionId) => sessionId == info.id);
+      // User already handled the question, so cancel any pending completion ping.
       case SesoriQuestionReplied(:final sessionID):
-        _maybeScheduleCompletion(sessionID);
+        _cancelDebounceForSessionGroup(sessionID);
+      // Rejected questions are also already seen by the user.
       case SesoriQuestionRejected(:final sessionID):
-        _maybeScheduleCompletion(sessionID);
+        _cancelDebounceForSessionGroup(sessionID);
+      // Permission replies are user actions, so cancel completion debounce.
       case SesoriPermissionReplied(:final requestID):
         final sessionId = _permissionRequestToSession.remove(requestID);
         if (sessionId != null) {
-          _maybeScheduleCompletion(sessionId);
+          _cancelDebounceForSessionGroup(sessionId);
         }
+      // Ignore unsupported events.
       default:
         break;
     }
   }
 
-  /// Cancel all timers.
+  /// Cancels all timers and closes the completion stream.
   void dispose() {
+    _cancelAllDebounceTimers();
+    if (!_completionController.isClosed) {
+      _completionController.close();
+    }
+  }
+
+  /// Clears debounce state but keeps the completion stream open.
+  void reset() {
+    _cancelAllDebounceTimers();
+    _completionSentForRoots.clear();
+    _permissionRequestToSession.clear();
+  }
+
+  /// Cancels every active debounce timer.
+  void _cancelAllDebounceTimers() {
     for (final timer in _debounceTimers.values) {
       timer.cancel();
     }
     _debounceTimers.clear();
   }
 
-  /// Cancel all timers and clear internal state.
-  void reset() {
-    dispose();
-    _completionSentForRoots.clear();
-    _permissionRequestToSession.clear();
-  }
-
+  /// Schedules (or reschedules) completion emission for [sessionId]'s root.
   void _maybeScheduleCompletion(String sessionId) {
     final rootSessionId = _tracker.resolveRootSessionId(sessionId);
     if (!_shouldTriggerCompletion(rootSessionId)) {
@@ -90,11 +107,15 @@ class CompletionNotifier {
       if (!_shouldTriggerCompletion(rootSessionId)) {
         return;
       }
-      _onCompletion(rootSessionId);
+      if (_completionController.isClosed) {
+        return;
+      }
+      _completionController.add(rootSessionId);
       _completionSentForRoots.add(rootSessionId);
     });
   }
 
+  /// Returns true when completion notification criteria are satisfied.
   bool _shouldTriggerCompletion(String rootSessionId) {
     return _tracker.wasPreviouslyBusy(rootSessionId) &&
         _tracker.isSessionGroupFullyIdle(rootSessionId) &&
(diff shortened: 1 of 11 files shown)


bridge/app/lib/src/push/completion_notifier.dart
```diff
@@ -0,0 +1,117 @@
...
+        _permissionRequestToSession.removeWhere((_, sessionId) => sessionId == info.id);
+      case SesoriQuestionReplied(:final sessionID):
+        _maybeScheduleCompletion(sessionID);
+      case SesoriQuestionRejected(:final sessionID):
+        _maybeScheduleCompletion(sessionID);
+      case SesoriPermissionReplied(:final requestID):
+        final sessionId = _permissionRequestToSession.remove(requestID);
+        if (sessionId != null) {
+          _maybeScheduleCompletion(sessionId);
+        }
```
AlexDochioiu, line 66: Those all are user interactions. If we have any pending notification for this session (like agent replied, etc) they should be cancelled and not sent. The user is aware that it finished since it just replied to some question. Unless am I misunderstanding something? Add some very brief docs to the private functions to understand what they do. Also some brief docs to each SSE event that we handle to explain what we do with it.

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -5,41 +5,107 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
...
   final PushRateLimiter _rateLimiter;
+  final PushSessionStateTracker _tracker;
+  late final CompletionNotifier _completionNotifier;
 
   PushNotificationService({
     required PushNotificationClient client,
     required PushRateLimiter rateLimiter,
+    required PushSessionStateTracker tracker,
+    Duration debounceDuration = const Duration(milliseconds: 500),
+    CompletionNotifier? completionNotifier,
```
AlexDochioiu: make this required and non-nullable

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -5,41 +5,107 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
...
   final PushNotificationClient _client;
   final PushRateLimiter _rateLimiter;
+  final PushSessionStateTracker _tracker;
+  late final CompletionNotifier _completionNotifier;
 
   PushNotificationService({
     required PushNotificationClient client,
     required PushRateLimiter rateLimiter,
+    required PushSessionStateTracker tracker,
+    Duration debounceDuration = const Duration(milliseconds: 500),
```
AlexDochioiu: I think this won't be needed anymore when completionNotifier becomes required non-nullable

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -5,41 +5,107 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
...
 import "push_notification_client.dart";
 import "push_rate_limiter.dart";
 import "push_send_exception.dart";
+import "push_session_state_tracker.dart";
 
 class PushNotificationService {
   final PushNotificationClient _client;
   final PushRateLimiter _rateLimiter;
+  final PushSessionStateTracker _tracker;
+  late final CompletionNotifier _completionNotifier;
```
AlexDochioiu: get rid of late

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -5,41 +5,107 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
...
+    _completionNotifier =
+        completionNotifier ??
+        CompletionNotifier(
+          tracker: _tracker,
+          debounceDuration: debounceDuration,
+          onCompletion: _sendCompletionNotification,
+        );
+  }
 
   void maybeSendForEvent(SesoriSseEvent event) {
```
AlexDochioiu: rename to processSseEvent or handleSseEvent

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -5,41 +5,107 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
...
   }) : _client = client,
-       _rateLimiter = rateLimiter;
+       _rateLimiter = rateLimiter,
+       _tracker = tracker {
+    _completionNotifier =
+        completionNotifier ??
+        CompletionNotifier(
+          tracker: _tracker,
+          debounceDuration: debounceDuration,
+          onCompletion: _sendCompletionNotification,
```
AlexDochioiu: Instead of callback, I prefer that `CompletionNotifier` exposes some PublishSubject stream (or similar) that `PushNotificationService` listens to.

bridge/app/lib/src/push/push_session_state_tracker.dart
```diff
@@ -0,0 +1,203 @@
...
+  final Map<String, SessionStatus> _sessionStatuses = {};
+  final Map<String, String?> _sessionParentIds = {};
+  final Map<String, Set<String>> _sessionChildren = {};
+  final Map<String, String> _sessionTitles = {};
+  final Map<String, String> _messageRoles = {};
+  final Map<String, String> _latestAssistantText = {};
+  final Set<String> _pendingQuestions = {};
+  final Set<String> _pendingPermissions = {};
+  final Map<String, String> _permissionRequestToSession = {};
+  final Set<String> _previouslyBusy = {};
```
AlexDochioiu: This is awful to look at. Can't we track it something like:
Map<String, SessionState> where the key is the session id and the value is some new state object that tracks all things related to given session. Would be much easier to read. You can consider the entire `SessionState` class to be private in the same file or @visibleForTesting with mutable fields (that get mutated by `PushSessionStateTracker`)

This would make reset/ cleanup/ state deleted/etc much easier to follow

shared/sesori_shared/lib/src/models/sesori/notification_data.dart
```diff
@@ -76,6 +76,8 @@ enum NotificationEventType {
   messageUpdated,
   @JsonValue("installation_update_available")
   installationUpdateAvailable,
+  @JsonValue("session_completed")
+  sessionCompleted,
```
AlexDochioiu: name may be a bit confusing? Is this when the turn completed from the AI? `agentTurnCompleted` may be a better name. Also `messageUpdated` seems not needed anymore? if so, remove it.

bridge/app/lib/src/push/push_session_state_tracker.dart
```diff
@@ -0,0 +1,203 @@
...
+    if (children == null || children.isEmpty) {
+      return true;
+    }
+    for (final childId in children) {
+      if (_sessionStatuses.containsKey(childId)) {
+        return false;
+      }
+    }
+    return true;
+  }
```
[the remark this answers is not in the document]
AlexDochioiu: Done — `isSessionGroupFullyIdle` now uses iterative BFS with a queue and visited set, traversing the full descendant tree instead of just direct children.

bridge/app/lib/src/push/push_session_state_tracker.dart
```diff
@@ -0,0 +1,203 @@
...
+    final children = _sessionChildren[sessionId];
+    if (children != null) {
+      for (final childId in children) {
+        if (_pendingQuestions.contains(childId) || _pendingPermissions.contains(childId)) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
AlexDochioiu: Done — `hasPendingInteraction` also uses BFS now, checking the full descendant tree for pending questions/permissions. Added grandchild-level tests for both methods.

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -8,38 +8,173 @@ import "../auth/token_refresh_exception.dart";
...
+        !_tracker.hasPendingInteraction(rootSessionId) &&
+        !_completionSentForRoots.contains(rootSessionId);
+  }
+
+  void _cancelDebounceForSessionGroup(String sessionId) {
+    final rootSessionId = _tracker.resolveRootSessionId(sessionId);
+    _cancelDebounceForRoot(rootSessionId);
+    if (rootSessionId != sessionId) {
+      _cancelDebounceForRoot(sessionId);
+    }
```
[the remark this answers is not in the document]
AlexDochioiu: Done — added the reparenting comment explaining why we cancel for both root and sessionId.

bridge/app/lib/src/push/completion_notifier.dart
AlexDochioiu: Done — all three (QuestionReplied, QuestionRejected, PermissionReplied) now call `_cancelDebounceForSessionGroup` instead of rescheduling. If the user just interacted, they're already aware. Added brief doc comments to every SSE event case and every private method.

bridge/app/lib/src/push/push_notification_service.dart
AlexDochioiu: Done — `completionNotifier` is now `required CompletionNotifier completionNotifier` (non-nullable).

bridge/app/lib/src/push/push_notification_service.dart
AlexDochioiu: Done — `debounceDuration` removed from the service constructor. It now lives only on `CompletionNotifier`, which is injected externally.

bridge/app/lib/src/push/push_notification_service.dart
AlexDochioiu: Done — `_completionNotifier` is now a plain `final` field, no `late`.

bridge/app/lib/src/push/push_notification_service.dart
AlexDochioiu: Done — renamed to `handleSseEvent`. Updated all callers (orchestrator, all test files).

bridge/app/lib/src/push/push_notification_service.dart
AlexDochioiu: Done — replaced the callback with a `StreamController<String>.broadcast()`. `CompletionNotifier` now exposes `Stream<String> get completions` and the service subscribes via `.listen()`. `dispose()` closes the controller; `reset()` does not (preserves the stream across reconnects).

bridge/app/lib/src/push/push_session_state_tracker.dart
AlexDochioiu: Done — replaced the 10 individual maps/sets with a single `Map<String, _SessionState> _sessions` where `_SessionState` is a private mutable class holding parentId, title, status, previouslyBusy, childIds, latestAssistantText, hasPendingQuestion, hasPendingPermission. Only `_messageRoles` and `_permissionRequestToSession` remain separate (keyed by messageID/requestID, not sessionID). `reset()` is now just `_sessions.clear()` + clearing the two non-session maps. Much cleaner.

shared/sesori_shared/lib/src/models/sesori/notification_data.dart
AlexDochioiu: Done — renamed to `agentTurnCompleted` (`@JsonValue("agent_turn_completed")`). Also removed `messageUpdated` entirely since per-message notifications are gone. Codegen regenerated.

PR #24, opened 2026-03-24
feat(notifications): cancel push notification on question answer/reject

AlexDochioiu:
## Summary

- Cancel the corresponding push notification from the device tray when a user answers or rejects an AI question from the phone
- Use deterministic notification IDs (FNV-1a hash of `sessionId + category`) so the same ID can be recomputed at cancel time without storing state
- Thread `sessionId` through the notification pipeline so `show()` can assign deterministic IDs for `aiInteraction` notifications

### New files
- **`notification_id_utils.dart`** — Pure FNV-1a 32-bit hash function producing Android PendingIntent-safe IDs (`[0, 2^27-1]`)
- **4 test files** — 25 new tests covering ID generation, cancel delegation, deterministic show(), service passthrough, and screen integration

### Modified files
- **`local_notification_manager.dart`** — Constructor injection for testability, `cancel()` method, optional `sessionId` param on `show()` that triggers deterministic IDs for `aiInteraction` category only
- **`notification_service.dart`** — `@visibleForTesting` seam on `onForegroundMessage` (matching existing `onAuthStateChanged` pattern), passes `sessionId` through to `show()`
- **`session_detail_screen.dart`** — Fire-and-forget `cancel()` call after both `replyToQuestion()` and `rejectQuestion()` callbacks

## Design decisions

- **FNV-1a, not `String.hashCode`** — Dart's `hashCode` isn't stable across VM runs; FNV-1a is deterministic
- **Per-session granularity** — `NotificationData` doesn't carry `requestId`, so all questions in the same session share one notification ID. In-app `QuestionModal` still tracks each question independently
- **`aiInteraction` only** — Deterministic IDs scoped to `NotificationCategory.aiInteraction`; all other categories keep timestamp-based IDs
- **Foreground-only** — `cancel()` only works for locally-shown notifications. Background FCM notifications auto-dismiss on tap

## Verification

- `flutter test` in `mobile/app/`: **354 tests pass, 0 failures**
- `dart analyze` on changed files: **0 issues**
- No changes to `sesori_shared` or `module_core`


mobile/app/lib/core/platform/local_notification_manager.dart
```diff
@@ -20,7 +22,10 @@ extension on NotificationImportance {
 
 @lazySingleton
 class LocalNotificationManager {
-  final FlutterLocalNotificationsPlugin _plugin = FlutterLocalNotificationsPlugin();
+  final FlutterLocalNotificationsPlugin _plugin;
+
+  LocalNotificationManager([FlutterLocalNotificationsPlugin? plugin])
+    : _plugin = plugin ?? FlutterLocalNotificationsPlugin();
```
AlexDochioiu: should be `{required this.plugin}` and you can make `FlutterLocalNotificationsPlugin` injectable via module if this is an external class.

mobile/app/lib/core/platform/local_notification_manager.dart
```diff
@@ -49,11 +54,19 @@ class LocalNotificationManager {
     required String title,
     required String body,
     required NotificationCategory category,
+    String? sessionId,
```
AlexDochioiu: required String? sessionId

mobile/app/lib/core/platform/notification_id_utils.dart
```diff
@@ -0,0 +1,23 @@
...
+
+/// Computes a deterministic notification ID from a session ID and notification category.
+///
+/// Uses FNV-1a 32-bit hash to ensure:
+/// - Same inputs always produce the same output (deterministic)
+/// - Different inputs produce different outputs (collision-resistant)
+/// - Output is within Android PendingIntent safe range [0, 2^27 - 1]
+///
+/// This is suitable for use as a notification ID in Android/iOS notification systems.
+int computeNotificationId(String sessionId, NotificationCategory category) {
```
AlexDochioiu: use named args with required. Also, please update AGENTS.md to always use named arguments with required keyword, INCLUDING for nullable ones. Never use positional args.

mobile/app/lib/core/platform/notification_id_utils.dart
```diff
@@ -0,0 +1,23 @@
...
+  const fnvOffsetBasis = 0x811c9dc5;
+  const fnvPrime = 0x01000193;
+  var hash = fnvOffsetBasis;
+
+  for (var i = 0; i < input.length; i++) {
+    hash ^= input.codeUnitAt(i);
+    hash = (hash * fnvPrime) & 0xFFFFFFFF;
+  }
+
+  return hash % 134217728; // safe positive range [0, 2^27 - 1]
```
AlexDochioiu, line 22: Any issue with simply using `input.hashCode` ? Collision concerns?

mobile/app/lib/features/session_detail/session_detail_screen.dart
```diff
@@ -113,6 +115,9 @@ class _SessionDetailBodyState extends State<_SessionDetailBody> {
           sessionId: question.sessionID,
           answers: answers,
         );
+        getIt<LocalNotificationManager>().cancel(
+          computeNotificationId(question.sessionID, NotificationCategory.aiInteraction),
+        );
```
AlexDochioiu: This should be done by `SessionDetailCubit`. Inject `LocalNotificationManager` in construction to `SessionDetailCubit` and do it there. Make sure you add tests to prove it's being done for `replyToQuestion`, `rejectQuestion`, any other place needed

mobile/app/lib/core/platform/local_notification_manager.dart
AlexDochioiu: Done. Changed to `{required FlutterLocalNotificationsPlugin plugin}` and registered `FlutterLocalNotificationsPlugin` as a `@lazySingleton` in `RegisterModule`. Injectable now provides the plugin instance, and the constructor takes it as a required named arg. Tests updated to `LocalNotificationManager(plugin: mockPlugin)`.

mobile/app/lib/core/platform/local_notification_manager.dart
AlexDochioiu: Done. Changed to `required String? sessionId`. All callers now pass it explicitly.

mobile/app/lib/core/platform/notification_id_utils.dart
AlexDochioiu: Done. Switched to `int computeNotificationId({required String sessionId, required NotificationCategory category})`. All callers updated. Also added the named required args convention to the root `AGENTS.md` under a new "Dart Coding Conventions" section — applies to all new code going forward, including nullable params.

mobile/app/lib/core/platform/notification_id_utils.dart
AlexDochioiu: Good question. The main issue with `input.hashCode` is **stability** — Dart's `String.hashCode` is not guaranteed to produce the same value across different VM invocations or Dart versions. It's explicitly documented as session-scoped. This matters here because we compute the ID at `show()` time and again at `cancel()` time, potentially across different app lifecycle states (e.g., app killed and relaunched between showing and cancelling). FNV-1a gives us a deterministic hash that's stable regardless of when/where it's computed. Collision risk is comparable — both are 32-bit hashes. The real differentiator is cross-session stability.

mobile/app/lib/features/session_detail/session_detail_screen.dart
AlexDochioiu: Done. Moved cancel logic entirely to `SessionDetailCubit`.

Created a `NotificationCanceller` abstract interface in `module_core/lib/src/platform/notification_canceller.dart` (following the existing platform interface pattern — `UrlLauncher`, `LifecycleSource`, etc.). `LocalNotificationManager` implements it with a `cancelForSession()` method.

DI registration: `NotificationCanceller` is registered in `RegisterModule` as a factory method that resolves to the `LocalNotificationManager` singleton. The cubit receives it as a required constructor param (`notificationCanceller: getIt<NotificationCanceller>()`).

The cubit now calls `_notificationCanceller.cancelForSession(...)` in both `replyToQuestion` (using the `sessionId` param) and `rejectQuestion` (using the `_sessionId` field). Cancel calls removed from `SessionDetailScreen` entirely.

Cubit tests updated to mock `NotificationCanceller` and verify cancel is called in both methods. Screen notification test file deleted since cancel logic is no longer there.

PR #29, opened 2026-03-26
fix(bridge): preserve SSE orphan queues on phone disconnect and relay reconnect

AlexDochioiu:
## Summary

Fixes SSE event buffering so that events are actually replayed when a phone reconnects within the 5-minute replay window. Two bugs were preventing this from working:

**Bug 1 — Single-phone disconnect destroyed the buffer:**
`SSEManager.unsubscribe()` checked `_subscribers.isEmpty` after removing the phone and, when it was the last (only) subscriber, called `queue.dispose()` instead of orphaning it. Since the typical use case is a single phone, the orphan queue mechanism was effectively dead code.

**Bug 2 — Relay reconnect destroyed all orphan state:**
`orchestrator.dart` called `_sseManager.stop()` whenever the bridge's relay WebSocket dropped, which disposed every subscriber queue and every orphan queue. Even if Bug 1 were fixed, a brief relay hiccup would still wipe all buffered events.

## Changes

- **`sse_manager.dart`**: `unsubscribe()` now always pauses the queue and adds it to the orphan list, regardless of remaining subscriber count
- **`sse_manager.dart`**: Added `orphanAll()` method that moves all active subscribers to orphan state (for relay reconnect)
- **`sse_manager.dart`**: Added `static const defaultReplayWindow = Duration(minutes: 5)` as single source of truth
- **`orchestrator.dart`**: Replaced `_sseManager.stop()` with `_sseManager.orphanAll()` on relay disconnect, and clear stale `activePhones`
- **`bridge.dart`**: Reference `SSEManager.defaultReplayWindow` instead of private constant
- **`sse_manager_test.dart`**: Added tests for single-subscriber orphaning, `orphanAll()`, and single-phone reconnect replay; updated all tests to reference `SSEManager.defaultReplayWindow`

### DIFF
--- a/bridge/app/lib/src/bridge/orchestrator.dart
+++ b/bridge/app/lib/src/bridge/orchestrator.dart
@@ -160,7 +160,8 @@ class OrchestratorSession {
         }
 
         Log.w("Connection lost. Reconnecting...");
-        _sseManager.stop();
+        _sseManager.orphanAll();
+        activePhones.clear();
 
         var backoff = const Duration(seconds: 1);
         while (!_cancelled) {
--- a/bridge/app/lib/src/bridge/orchestrator.dart
+++ b/bridge/app/lib/src/bridge/orchestrator.dart
@@ -159,7 +159,7 @@ class OrchestratorSession {
           break;
         }
 
-        Log.w("Connection lost. Reconnecting...");
+        Log.w("Relay connection lost. Reconnecting...");
         _sseManager.orphanAll();
         activePhones.clear();
 
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -14,6 +14,7 @@ import 'package:sesori_bridge/src/bridge/orchestrator.dart';
 import 'package:sesori_bridge/src/bridge/persistence/bridge_diagnostics.dart';
 import 'package:sesori_bridge/src/bridge/persistence/hidden_projects_store.dart';
 import 'package:sesori_bridge/src/bridge/relay_client.dart';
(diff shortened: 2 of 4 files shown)


bridge/app/lib/src/bridge/orchestrator.dart
```diff
@@ -160,7 +160,8 @@ class OrchestratorSession {
         }
 
         Log.w("Connection lost. Reconnecting...");
-        _sseManager.stop();
+        _sseManager.orphanAll();
```
AlexDochioiu, line 163: Is this correct? what if we have multiple phones connected at the same time and one disconnects? Would this orphane/pause queues for all phones?

AlexDochioiu: This is correct — there are two separate code paths and they don't interfere:

**Single phone disconnects** (relay still up):
The relay sends phone_disconnected, and the orchestrator calls _sseManager.removeSubscriber(connID) on line 267. This only orphans **that one phone's queue**. Other phones' queues keep draining normally.

**Relay connection itself drops** (this line):
When the bridge's WebSocket to the relay drops, _runRelayLoop() throws/ends, and we reach this code. At that point **all phones are unreachable** because the relay is the only pipe between bridge and phones. So orphaning all subscribers is correct — there is no phone we could still deliver to. When the relay reconnects and phones reconnect (with new connIDs), they pick up their orphan queues via subscribePath().

So:
- 1 phone disconnects while relay is up: removeSubscriber(connID) — only that phone orphaned
- Relay drops: orphanAll() — all phones orphaned (all are unreachable anyway)

PR #31, opened 2026-03-26
feat: reduce bridge-to-mobile data transfer — strip unused fields, filter parts, stop heartbeats

AlexDochioiu:
## Summary

- Strip unused fields from shared `Message` (5 fields) and `MessagePart` (8 fields) models, removing 4 orphaned types (`MessageTime`, `MessageTokens`, `TokenCache`, `PartTime`)
- Strip matching fields from plugin interface models (`PluginMessage`, `PluginMessagePart`), removing 2 orphaned types (`PluginMessageTime`, `PluginPartTime`)
- Filter out invisible message parts (`file`, `snapshot`) and truncate `ToolState.output` to 500 chars at the OpenCode plugin mapping layer
- Filter heartbeat SSE events, strip `session.diff` payloads (mobile only uses as refresh trigger), and filter/truncate SSE message parts in the bridge event mapper
- Remove heartbeat watchdog from mobile `ConnectionService` and fix all test fixtures referencing removed fields

## Motivation

Bridge was sending significantly more data to mobile than needed — full file content in `session.diff` SSE events (potentially MBs) that mobile discarded, file/snapshot message parts mobile never rendered, tool output mobile truncated to 500 chars client-side, and heartbeat events that only reset a watchdog timer. This reduces bandwidth to make the app viable on slower connections.

## What changed

| Layer | Change | Impact |
|-------|--------|--------|
| `sesori_shared` | Stripped `Message` to 6 fields, `MessagePart` to 10 fields, removed `diff` from `sessionDiff` SSE event | Both HTTP and SSE paths automatically slimmed via `fromJson`/`toJson` boundary |
| `sesori_plugin_interface` | Matching field removal + `BridgeSseSessionDiff.diff` removed | Data reduction boundary moved to plugin mapping layer |
| `sesori_plugin_opencode` | Filter file/snapshot parts, truncate tool output to 500 chars | HTTP response path optimized |
| `bridge/app` handler | Adapted `GetSessionMessagesHandler` for slim models | 1:1 mapping, removed dead field lines |
| `bridge/app` event mapper | Heartbeats → `null`, session.diff stripped, file/snapshot parts filtered, tool output truncated | SSE path optimized |
| `mobile` | Heartbeat watchdog removed, test fixtures updated | No false 15-second disconnects |

## Test results

- `shared/sesori_shared`: 164/164 ✅
- `bridge/sesori_plugin_opencode`: 105/105 ✅ (includes new filter/truncation tests)
- `bridge/app`: 368/368 ✅ (includes 7 new event mapper tests)
- `mobile/module_core`: 52/52 ✅
- `mobile/app`: 382/382 ✅

**Net: -1,458 lines deleted, +436 added across 24 files**


bridge/app/lib/src/bridge/sse/bridge_event_mapper.dart
```diff
@@ -55,7 +51,14 @@ class BridgeEventMapper {
...
       case BridgeSseMessagePartUpdated(:final part):
-        return _tryParseSseEvent({"type": "message.part.updated", "part": part});
+        final partType = part["type"] as String?;
+        if (partType == "file" || partType == "snapshot") return null;
+
+        final modifiedPart = _truncateToolOutput(part);
+        return _tryParseSseEvent({
+          "type": "message.part.updated",
+          "part": modifiedPart,
+        });
```
AlexDochioiu: I don't like that this uses untyped data. We should fully rely on freezed typed data, for all of `part`, `modifiedPart`, `_tryParseSseEvent` argument.

bridge/app/lib/src/bridge/sse/bridge_event_mapper.dart
```diff
@@ -198,4 +201,19 @@ class BridgeEventMapper {
       return null;
     }
   }
+
+  /// Returns a copy of [part] with `state.output` truncated to 500 characters
+  /// if it exceeds that length, or the original map if no truncation is needed.
+  Map<String, dynamic> _truncateToolOutput(Map<String, dynamic> part) {
```
AlexDochioiu: both return and arg must be proper typed freezed classes

bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
```diff
@@ -507,21 +507,11 @@ class OpenCodePlugin implements BridgePlugin {
...
-          MessageTime(:final created, :final completed) => PluginMessageTime(
-            created: created,
-            completed: completed,
-          ),
-          null => null,
-        },
-        finish: info.finish,
       ),
-      parts: parts.map(_mapMessagePart).toList(),
+      parts: parts.where((p) => p.type != "file" && p.type != "snapshot").map(_mapMessagePart).toList(),
```
AlexDochioiu: `type` must be some enum, not magic strings

bridge/app/lib/src/bridge/routing/get_session_messages_handler.dart
```diff
@@ -7,6 +7,20 @@ import "request_handler.dart";
...
+MessagePartType _mapPartType(PluginMessagePartType type) => switch (type) {
+  PluginMessagePartType.text => MessagePartType.text,
+  PluginMessagePartType.reasoning => MessagePartType.reasoning,
+  PluginMessagePartType.tool => MessagePartType.tool,
+  PluginMessagePartType.subtask => MessagePartType.subtask,
+  PluginMessagePartType.stepStart => MessagePartType.stepStart,
+  PluginMessagePartType.stepFinish => MessagePartType.stepFinish,
+  PluginMessagePartType.file => MessagePartType.file,
+  PluginMessagePartType.snapshot => MessagePartType.snapshot,
+};
```
AlexDochioiu: This code seems duplicated. Set up some shared extension function and use it everywhere we need.

bridge/app/lib/src/bridge/sse/bridge_event_mapper.dart
```diff
@@ -55,7 +65,30 @@ class BridgeEventMapper {
           messageID: messageID,
         );
       case BridgeSseMessagePartUpdated(:final part):
-        return _tryParseSseEvent({"type": "message.part.updated", "part": part});
+        if (!part.type.isVisible) return null;
+        final truncated = _truncateToolOutput(part);
+        return SesoriSseEvent.messagePartUpdated(
+          part: MessagePart(
```
AlexDochioiu: Make `PluginMessagePart` -> `MessagePart` into an extension fct. If we do this mapping anywhere else, make sure to use the same ext function.

bridge/app/lib/src/bridge/sse/bridge_event_mapper.dart
```diff
@@ -55,7 +65,30 @@ class BridgeEventMapper {
...
+            id: truncated.id,
+            sessionID: truncated.sessionID,
+            messageID: truncated.messageID,
+            type: _mapPartType(truncated.type),
+            text: truncated.text,
+            tool: truncated.tool,
+            state: switch (truncated.state) {
```
AlexDochioiu: Make the mapping from `PluginToolState` into an extension function. If we do this mapping elsewhere, make sure the ext fct is used.

bridge/sesori_plugin_interface/lib/src/models/plugin_message.dart
```diff
@@ -4,6 +4,34 @@ part "plugin_message.freezed.dart";
...
+  @JsonValue("step-finish")
+  stepFinish,
+  @JsonValue("file")
+  file,
+  @JsonValue("snapshot")
+  snapshot
+  ;
+
+  /// Whether this part type is visible to mobile (rendered in the UI).
+  bool get isVisible => this != file && this != snapshot;
```
AlexDochioiu: what were we showing previously for those? Is this changing behaviour?

bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
```diff
@@ -507,58 +507,51 @@ class OpenCodePlugin implements BridgePlugin {
...
+  static PluginMessagePartType _toPluginPartType({required String type}) => switch (type) {
+    "text" => PluginMessagePartType.text,
+    "reasoning" => PluginMessagePartType.reasoning,
+    "tool" => PluginMessagePartType.tool,
+    "subtask" => PluginMessagePartType.subtask,
+    "step-start" => PluginMessagePartType.stepStart,
+    "step-finish" => PluginMessagePartType.stepFinish,
+    "file" => PluginMessagePartType.file,
+    "snapshot" => PluginMessagePartType.snapshot,
+    _ => PluginMessagePartType.file,
```
AlexDochioiu: Add some `PluginMessagePartType.unknown` and log a clear warning in console with the `type` that we missed.

bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
AlexDochioiu: No behavior change — mobile's `AssistantMessageCard._isVisible()` already filtered out file/snapshot parts client-side with string matching. This just moves the filtering upstream to the plugin layer, so the data never reaches mobile in the first place.

PR #32, opened 2026-03-26
fix(mobile): session list not updating in real time

AlexDochioiu:
## Summary

- **Fix session list not recovering after connection blips** — `SessionListCubit` now subscribes to `ConnectionService.status` and silently refreshes sessions when the connection is restored, matching the pattern already used by `SessionDetailCubit`
- **Optimistically insert created sessions** — `createSession()` now immediately adds the new session to the local list instead of relying entirely on an SSE event that may never arrive
- **Add safety guard and debug logging** — `isClosed` guard on `_handleEvent`, plus `[SessionList]` prefixed `logd()` calls throughout event handlers for runtime tracing

## Root Cause

`SessionListCubit` subscribed to `connectionService.events` (a broadcast stream) for real-time session CRUD updates, but never subscribed to `connectionService.status`. After any connection drop — even a brief one — missed SSE events were permanently lost because broadcast streams don't replay. Activity updates (running/background tasks) worked because they flow through `SseEventRepository` which uses `BehaviorSubject` with the latest cached value.

## Test Coverage

Added 5 new tests:
- SSE `session.created` for same project adds to list
- SSE `session.updated` for same project updates existing session
- SSE `session.deleted` for same project removes from list
- SSE `session.created` for child session (`parentID != null`) is ignored
- Connection reconnect triggers silent refresh

All 387 tests pass. `dart analyze` clean.


mobile/module_core/lib/src/cubits/session_list/session_list_cubit.dart
```diff
@@ -69,46 +75,85 @@ class SessionListCubit extends Cubit<SessionListState> {
...
+    if (_allSessions.length == before) {
+      logd("[SessionList] session.deleted not_found id=${session.id}");
+      return;
+    }
 
+    logd("[SessionList] session.deleted removed id=${session.id}");
     _emitFiltered();
   }
 
+  void _onConnectionStatusChanged(ConnectionStatus status) {
```
AlexDochioiu, line 151: Do we need similar handling for list of projects?

AlexDochioiu:
Re: [ProjectListCubit question](https://github.com/sesori-ai/sesori_apps_monorepo/pull/32#discussion_r2993436792) — Good catch. ProjectListCubit had the same gap. It was partially mitigated by the navigate-back refresh and the throttled auto-refresh, but if the user stayed on the projects page during a connection blip, stale data would persist.

Added the same `connectionService.status` subscription + `_onConnectionStatusChanged` handler in ee14748, plus a reconnect test. All 388 tests pass.

PR #35, opened 2026-03-26
feat(bridge): return full session history and add bandwidth tracking

AlexDochioiu:
## Summary

- **Remove last-exchange filtering** from `getSessionMessages` — the endpoint now returns the complete message history instead of only the final user/assistant exchange. `OpenCodeService.getLastExchange()` is replaced with a simple `getMessages()` passthrough.
- **Add `bytesSent` broadcast stream** to `OrchestratorSession` — emits byte counts for every API response and SSE event sent to mobile. The orchestrator announces data, it doesn't track it.
- **Add `BandwidthTracker`** — a standalone subscriber that, when the debug server is active (`--debug-port`), listens to the bytes stream and logs rolling bandwidth totals every 60 seconds (last 1m / 10m / 1h).

## Changes

| File | What changed |
|------|-------------|
| `opencode_service.dart` | Replaced `getLastExchange()` with `getMessages()` — no more filtering to last user message |
| `opencode_plugin_impl.dart` | Updated call site |
| `bridge_plugin.dart` | Updated interface doc comment |
| `opencode_service_test.dart` | Rewrote tests to verify full history returned |
| `orchestrator.dart` | Added `StreamController<int>.broadcast()`, exposed `bytesSent` stream, wired SSE manager via `onBytesSent` callback |
| `sse_manager.dart` | Replaced `BandwidthTracker?` injection with `void Function(int)?` callback |
| `bandwidth_tracker.dart` | New — records bytes, prunes >1h entries, logs `[bandwidth] last 1m / 10m / 1h` every 60s |
| `bridge.dart` | Creates `BandwidthTracker` only when `debugPort != null`, subscribes to `session.bytesSent` |

## Testing

- `dart analyze` clean across all three bridge packages
- `dart test` passes (369 app tests + 100 plugin tests)

### DIFF
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -197,11 +197,7 @@ Future<void> main(List<String> args) async {
   final session = orchestrator.create();
 
   // Wire up bandwidth tracking when debug server is active
-  BandwidthTracker? bandwidthTracker;
-  if (debugPort != null) {
-    bandwidthTracker = BandwidthTracker();
-    session.bytesSent.listen((bytes) => bandwidthTracker!.record(bytes: bytes));
-  }
+  final bandwidthTracker = debugPort != null ? BandwidthTracker(bytesSent: session.bytesSent) : null;
 
   // Create and start debug server if requested
   DebugServer? debugServer;
@@ -229,18 +225,15 @@ Future<void> main(List<String> args) async {
   }
 
   // Run bridge; stop the server process on exit regardless of outcome
-  bandwidthTracker?.start();
   try {
     await session.run();
   } catch (e) {
     Log.e('$e');
-    bandwidthTracker?.stop();
-    await _shutdown(cmd, sigintSub, sigtermSub, debugServer, db);
+    await _shutdown(cmd, sigintSub, sigtermSub, debugServer, db, bandwidthTracker);
     exit(1);
   }
 
-  bandwidthTracker?.stop();
-  await _shutdown(cmd, sigintSub, sigtermSub, debugServer, db);
+  await _shutdown(cmd, sigintSub, sigtermSub, debugServer, db, bandwidthTracker);
 }
 
 /// Performs graceful shutdown: stops the server and cancels signal listeners.
@@ -253,11 +246,14 @@ Future<void> _shutdown(
   StreamSubscription<ProcessSignal>? sigtermSub,
   DebugServer? debugServer,
   AppDatabase db,
+  BandwidthTracker? bandwidthTracker,
 ) async {
   final safetyTimer = Timer(const Duration(seconds: 10), () {
     Log.e('Failed to finish gracefully');
     exit(0);
   });
+
+  bandwidthTracker?.dispose();
 
   await Future.wait([
     stopServer(serverProcess),
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -8,6 +8,7 @@ import 'package:sesori_bridge/src/auth/profile.dart';
 import 'package:sesori_bridge/src/auth/token.dart';
 import 'package:sesori_bridge/src/auth/token_manager.dart';
 import 'package:sesori_bridge/src/auth/validate.dart';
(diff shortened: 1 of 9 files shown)


bridge/app/bin/bridge.dart
```diff
@@ -195,6 +196,13 @@ Future<void> main(List<String> args) async {
   );
   final session = orchestrator.create();
 
+  // Wire up bandwidth tracking when debug server is active
+  BandwidthTracker? bandwidthTracker;
+  if (debugPort != null) {
+    bandwidthTracker = BandwidthTracker();
```
AlexDochioiu: `session.bytesSent` should be sent to `BandwidthTracker` in constructor and `BandwidthTracker` should start listening in construction automatically. On `dispose()` it stops.

bridge/app/bin/bridge.dart
```diff
@@ -195,6 +196,13 @@ Future<void> main(List<String> args) async {
...
+    session.bytesSent.listen((bytes) => bandwidthTracker!.record(bytes: bytes));
```
AlexDochioiu: No public `.record` method should exist

bridge/app/bin/bridge.dart
```diff
@@ -221,14 +229,17 @@ Future<void> main(List<String> args) async {
   }
 
   // Run bridge; stop the server process on exit regardless of outcome
+  bandwidthTracker?.start();
```
AlexDochioiu: it should auto-start if created

bridge/app/bin/bridge.dart
```diff
@@ -221,14 +229,17 @@ Future<void> main(List<String> args) async {
...
   try {
     await session.run();
   } catch (e) {
     Log.e('$e');
+    bandwidthTracker?.stop();
     await _shutdown(cmd, sigintSub, sigtermSub, debugServer, db);
```
AlexDochioiu: `_shutdown` should dispose this too (if not null)

bridge/app/lib/src/bridge/sse/sse_manager.dart
```diff
@@ -16,6 +16,8 @@ class SSEManager {
   /// How long a disconnected subscriber's orphan queue stays valid.
   final Duration replayWindow;
 
+  final void Function(int bytes)? _onBytesSent;
```
AlexDochioiu: `final void Function(int bytes) _onBytesSent`

bridge/app/lib/src/bridge/sse/sse_manager.dart
```diff
@@ -25,7 +27,7 @@ class SSEManager {
 
   List<int>? _roomKey;
 
-  SSEManager({required this.replayWindow});
+  SSEManager({required this.replayWindow, required void Function(int bytes)? onBytesSent}) : _onBytesSent = onBytesSent;
```
AlexDochioiu: `required void Function(int bytes) onBytesSent`

bridge/app/lib/src/bridge/sse/sse_manager.dart
```diff
@@ -157,6 +159,7 @@ class SSEManager {
       final relayMessage = RelayMessage.sseEvent(data: eventData);
       final payloadBytes = utf8.encode(jsonEncode(relayMessage.toJson()));
       Log.v("[sse] sending ${payloadBytes.length} bytes to connID=$connID");
+      _onBytesSent?.call(payloadBytes.length);
```
AlexDochioiu:
```
_onBytesSent.call(payloadBytes.length);
```

bridge/app/lib/src/bridge/bandwidth_tracker.dart
```diff
@@ -0,0 +1,72 @@
...
+    _timer = null;
+  }
+
+  void _logStats() {
+    final now = DateTime.now();
+    final last1m = _sumSince(now.subtract(const Duration(minutes: 1)));
+    final last10m = _sumSince(now.subtract(const Duration(minutes: 10)));
+    final last1h = _sumSince(now.subtract(const Duration(hours: 1)));
+
+    Log.i(
```
AlexDochioiu: `Log.d`

PR #40, opened 2026-03-26
fix(mobile): make inline subtask widget clickable in chat

AlexDochioiu:
## Summary

- **Propagate child session title updates via SSE** — `SessionDetailCubit._handleGlobalEvent()` now handles `SesoriSessionUpdated` for child sessions, keeping the `children` list up-to-date (previously titles stayed `null` forever after SSE creation)
- **Improve subtask-to-child session matching** — `SubtaskPartWidget._findChildSession()` now uses multi-strategy matching: exact → case-insensitive → contains (previously only exact match, which failed for null/mismatched titles)
- **Add `readOnly: "true"` to navigation** — Clicking an inline subtask now navigates to the child session detail in read-only mode, matching the existing bottom bar/sheet behavior

## Problem

The inline subtask widget in chat messages was not clickable. Root cause was two-layered:
1. **Stale data**: Child session titles were never updated after initial SSE creation (the cubit dropped `SesoriSessionUpdated` events for children)
2. **Brittle matching**: The widget used exact string comparison only (`child.title == desc`), failing when titles were `null` or had different casing

## Changes

| File | Change |
|------|--------|
| `mobile/module_core/.../session_detail_cubit.dart` | Add `SesoriSessionUpdated` handler in `_handleGlobalEvent` + new `_onChildSessionUpdated()` method |
| `mobile/app/.../subtask_part_widget.dart` | Multi-strategy matching in `_findChildSession()` + `readOnly: "true"` in navigation queryParams |

## Verification

- `dart analyze` exits 0 in all three mobile modules
- `dart test` passes in `module_core/` (107 tests)
- `flutter test` passes in `app/` (339 tests)

### DIFF
--- a/mobile/app/lib/features/session_detail/widgets/subtask_part_widget.dart
+++ b/mobile/app/lib/features/session_detail/widgets/subtask_part_widget.dart
@@ -38,13 +38,17 @@ class SubtaskPartWidget extends StatelessWidget {
           borderRadius: BorderRadius.circular(8),
           onTap: childSession != null
               ? () {
+                  final queryParams = <String, String>{"readOnly": "true"};
+                  if (childSession.title != null) {
+                    queryParams["title"] = childSession.title!;
+                  }
                   context.pushRoute(
                     AppRoute.sessionDetail,
                     pathParams: {
                       "projectId": childSession.projectID,
                       "sessionId": childSession.id,
                     },
-                    queryParams: childSession.title != null ? {"title": childSession.title!} : null,
+                    queryParams: queryParams,
                   );
                 }
               : null,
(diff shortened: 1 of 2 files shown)


mobile/app/lib/features/session_detail/widgets/subtask_part_widget.dart
```diff
@@ -44,7 +44,10 @@ class SubtaskPartWidget extends StatelessWidget {
                       "projectId": childSession.projectID,
                       "sessionId": childSession.id,
                     },
-                    queryParams: childSession.title != null ? {"title": childSession.title!} : null,
+                    queryParams: {
+                      "readOnly": "true",
+                      if (childSession.title != null) "title": childSession.title!,
```
AlexDochioiu:
```suggestion
                      "title": ?childSession.title,
```

PR #45, opened 2026-03-27
feat: add rename functionality for sessions and projects

AlexDochioiu:
## Summary

- Add full-stack rename support for both sessions and projects — from OpenCode API through the bridge plugin/handlers to the mobile UI
- Users long-press a session or project in the list, tap "Rename" in the bottom sheet, edit the name in a dialog pre-filled with the current value, and save
- TDD tests at every layer (handlers, plugin, services, cubits)

### Shared (`sesori_shared`)
- New Freezed request models: `RenameSessionRequest` (title field) and `RenameProjectRequest` (name field)

### Bridge
- **Plugin interface**: Added `renameSession()` and `renameProject()` to `BridgePlugin` abstract class
- **OpenCode plugin**: `renameSession()` calls `PATCH /session/:id` with title body; `renameProject()` resolves worktree→UUID via `GET /project/current`, then calls `PATCH /project/:id` with name body
- **API client**: Added `updateProject()` method to `OpenCodeApi`
- **Handlers**: New `RenameSessionHandler` (`PATCH /session/:id/title`) and `RenameProjectHandler` (`PATCH /project/:id/name`) — sub-paths avoid conflict with existing archive handler
- **Tests**: 18 new handler tests + 2 plugin integration tests

### Mobile
- **Services**: Added `renameSession()` to `SessionService` and `renameProject()` to `ProjectService`
- **Cubits**: Added rename methods to `SessionListCubit` and `ProjectListCubit` (server-confirmed: call API → refresh list → return bool)
- **UI**: New `RenameSessionDialog` and `RenameProjectDialog` following `AddProjectDialog` pattern (pre-filled text field, loading state, disabled-when-empty save, snackbar feedback)
- **Menu items**: "Rename" option added as first item in both session and project long-press bottom sheets
- **Localization**: 10 new keys in `app_en.arb`
- **Tests**: 9 new service/cubit tests

## Test Results

| Suite | Result |
|-------|--------|
| Bridge app (385 tests) | ✅ All pass |
| Bridge plugin (102 tests) | ✅ All pass |
| Mobile module_core (113 tests) | ✅ All pass |
| Mobile app (342 tests) | ✅ All pass |

### DIFF
--- a/bridge/app/lib/src/bridge/routing/rename_project_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/rename_project_handler.dart
@@ -3,15 +3,14 @@ import "dart:convert";
 import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 import "package:sesori_shared/sesori_shared.dart";
 
+import "plugin_project_mapper.dart";
 import "request_handler.dart";
 
-const _idParam = "id";
-
-/// Handles `PATCH /project/:id/name` — renames a project.
+/// Handles `PATCH /project/name` — renames a project.
 class RenameProjectHandler extends RequestHandler {
   final BridgePlugin _plugin;
 
-  RenameProjectHandler(this._plugin) : super(HttpMethod.patch, "/project/:$_idParam/name");
+  RenameProjectHandler(this._plugin) : super(HttpMethod.patch, "/project/name");
 
   @override
   Future<RelayResponse> handle(
@@ -20,11 +19,6 @@ class RenameProjectHandler extends RequestHandler {
     required Map<String, String> queryParams,
     String? fragment,
   }) async {
-    final projectId = pathParams[_idParam];
-    if (projectId == null || projectId.isEmpty) {
-      return buildErrorResponse(request, 400, "missing project id");
-    }
-
     final RenameProjectRequest renameRequest;
     try {
       final decoded = jsonDecode(request.body ?? "{}");
@@ -41,21 +35,11 @@ class RenameProjectHandler extends RequestHandler {
     }
 
     final updated = await _plugin.renameProject(
-      projectId,
+      projectId: renameRequest.projectId,
       name: renameRequest.name,
     );
 
-    final project = Project(
-      id: updated.id,
-      name: updated.name,
-      time: switch (updated.time) {
-        PluginProjectTime(:final created, :final updated) => ProjectTime(
-          created: created,
-          updated: updated,
-        ),
-        null => null,
-      },
-    );
+    final project = updated.toSharedProject();
 
     return buildOkJsonResponse(request, jsonEncode(project.toJson()));
   }
--- a/bridge/app/lib/src/bridge/routing/rename_project_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/rename_project_handler.dart
@@ -1,0 +1,62 @@
+import "dart:convert";
+
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "request_handler.dart";
+
+const _idParam = "id";
+
+/// Handles `PATCH /project/:id/name` — renames a project.
+class RenameProjectHandler extends RequestHandler {
+  final BridgePlugin _plugin;
+
+  RenameProjectHandler(this._plugin) : super(HttpMethod.patch, "/project/:$_idParam/name");
+
+  @override
+  Future<RelayResponse> handle(
+    RelayRequest request, {
+    required Map<String, String> pathParams,
+    required Map<String, String> queryParams,
+    String? fragment,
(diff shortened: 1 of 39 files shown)


bridge/app/lib/src/bridge/routing/rename_project_handler.dart
```diff
@@ -0,0 +1,62 @@
...
+      return buildErrorResponse(request, 400, "invalid JSON body");
+    }
+
+    final updated = await _plugin.renameProject(
+      projectId,
+      name: renameRequest.name,
+    );
+
+    final project = Project(
+      id: updated.id,
```
AlexDochioiu: Don't we have some shared extension function or similar to do this? I think this mapping may repeat in multiple places. If it doesn't exist, make an extension function for it to share.

bridge/app/lib/src/bridge/routing/rename_session_handler.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    final updated = await _plugin.renameSession(
+      sessionId,
+      title: renameRequest.title,
+    );
+
+    final session = Session(
+      id: updated.id,
+      projectID: updated.projectID,
+      directory: updated.directory,
```
AlexDochioiu: Don't we have some shared extension function or similar to do this? I think this mapping may repeat in multiple places. If it doesn't exist, make an extension function for it to share.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -30,6 +30,12 @@ abstract class BridgePlugin {
 
   Future<PluginSession> updateSessionArchiveStatus(String sessionId, {required bool archived});
 
+  /// Rename a session's title.
+  Future<PluginSession> renameSession(String sessionId, {required String title});
```
AlexDochioiu: both should be `required` named params

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -30,6 +30,12 @@ abstract class BridgePlugin {
...
+
+  /// Rename a project.
+  Future<PluginProject> renameProject(String projectId, {required String name});
```
AlexDochioiu: both args should be `required` named params

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -159,6 +159,27 @@ class OpenCodeApi {
...
+    required String projectId,
+    required Map<String, dynamic> body,
+  }) async {
+    final client = http.Client();
+    try {
+      final response = await client.patch(
+        Uri.parse("$serverURL/project/$projectId"),
+        headers: {
+          ..._authHeaders,
+          "content-type": "application/json",
```
AlexDochioiu, line 175: include `        _directoryOpenCodeHeader: directory,`

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -159,6 +159,27 @@ class OpenCodeApi {
     }
   }
 
+  Future<Project> updateProject({
+    required String projectId,
```
AlexDochioiu, line 165: add a comment that this is the actual project id (as tracked by opencode) and not the work path which we use as project id in sesori.

Also, add `required String directory`

mobile/module_core/lib/src/capabilities/project/project_service.dart
```diff
@@ -97,4 +97,19 @@ class ProjectService {
...
     );
   }
+
+  /// Renames the project with the given [projectId] to [name].
+  Future<ApiResponse<Project>> renameProject({
+    required String projectId,
+    required String name,
+  }) {
+    return _client.patch(
+      "/project/$projectId/name",
```
AlexDochioiu: `projectId` is often generally a path which will break if used as url path param. You must move the projectId inside the `RenameProjectRequest` body.

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -93,6 +93,17 @@ class SessionService {
...
   }
 
+  Future<ApiResponse<Session>> renameSession({required String sessionId, required String title}) {
+    return _client.patch(
+      "/session/$sessionId/title",
+      fromJson: (json) => switch (json) {
+        final Map<String, dynamic> map => Session.fromJson(map),
+        _ => throw FormatException("expected map, got ${json.runtimeType}"),
+      },
+      body: RenameSessionRequest(title: title).toJson(),
```
AlexDochioiu: to avoid any surprises with unexpected chars or formats, let's move session Id as part of the `RenameSessionRequest` body

PR #57, opened 2026-03-29
feat: lazy session creation — create on first message

AlexDochioiu:
## Summary

- **Navigate immediately**: Tapping "new session" FAB now opens the new session screen instantly — no waiting for backend session creation
- **Create on first message**: Session is only created when user sends their first message, via an enriched `POST /session` endpoint that combines session creation with the initial prompt
- **Route replacement**: After session creation succeeds, the new session screen is replaced with the real session detail screen (fresh `SessionDetailCubit` with the assigned session ID)

### Shared (`sesori_shared`)
- `CreateSessionRequest` enriched with `parts`, `agent`, `model` fields; `parentSessionId` removed

### Bridge
- `BridgePlugin.createSession()` interface updated with prompt fields
- OpenCode plugin chains: create session → register tracker → send prompt
- `CreateSessionHandler` parses enriched request and passes to plugin
- All bridge test doubles and handler tests updated

### Mobile
- `SessionService.createSessionWithMessage()` replaces `createSession()`
- `SessionListCubit.createSession()` removed; `projectId` exposed via getter
- New `AppRoute.newSession` route at `/projects/:projectId/sessions/new`
- New `NewSessionCubit` + `NewSessionState` (idle → sending → created/error)
- New `NewSessionScreen` with `PromptInput` reuse and inline error display
- FAB navigates to new session route instead of blocking on session creation
- Tests updated across both `module_core` and `app`

## Verification
- `dart analyze` passes on `sesori_shared`, `module_core`, and `app`
- `dart test` passes on `sesori_shared` (164 tests), bridge `app` (422 tests), and `module_core` (139 tests)

### DIFF
--- a/bridge/app/lib/src/bridge/routing/create_session_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/create_session_handler.dart
@@ -4,6 +4,7 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 import "package:sesori_shared/sesori_shared.dart";
 
 import "../worktree_service.dart";
+import "prompt_part_mapper.dart";
 import "request_handler.dart";
 
 /// Handles `POST /session` — creates a session for a given project.
@@ -48,13 +49,7 @@ class CreateSessionHandler extends RequestHandler {
       parentSessionId: parentSessionId,
     );
 
-    final parts = createRequest.parts
-        .map(
-          (p) => switch (p) {
-            PromptPartText(:final text) => PluginPromptPart.text(text: text),
-          },
-        )
-        .toList();
+    final parts = createRequest.parts.map((p) => p.toPlugin()).toList();
 
     final model = switch (createRequest.model) {
       PromptModel(:final providerID, :final modelID) => (providerID: providerID, modelID: modelID),
--- a/bridge/app/lib/src/bridge/routing/create_session_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/create_session_handler.dart
@@ -41,12 +41,25 @@ class CreateSessionHandler extends RequestHandler {
     }
 
     final projectId = createRequest.projectId;
-    final parentSessionId = createRequest.parentSessionId;
+    final String? parentSessionId = null;
 
     final worktreeResult = await _worktreeService.prepareWorktreeForSession(
       projectId: projectId,
       parentSessionId: parentSessionId,
     );
+
+    final parts = createRequest.parts
+        .map(
+          (p) => switch (p) {
+            PromptPartText(:final text) => PluginPromptPart.text(text: text),
+          },
+        )
+        .toList();
+
+    final model = switch (createRequest.model) {
+      PromptModel(:final providerID, :final modelID) => (providerID: providerID, modelID: modelID),
+      null => null,
+    };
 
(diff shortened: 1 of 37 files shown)


bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
```diff
@@ -220,19 +220,50 @@ class OpenCodePlugin implements BridgePlugin {
...
-        parentSessionId: parentSessionId,
+        parentSessionId: null,
       ),
     );
     _service.tracker.registerSession(
       sessionId: session.id,
       directory: session.directory,
     );
+
+    final body = <String, dynamic>{
```
AlexDochioiu: Create a freezed class for this body. Make sure `_service.repository.api.sendPrompt` expects the actual freezed object and it does `toJson` internally. Update all places to use the same model to avoid re-constructing this map in multiple places which will likely lead to issues in the future (where we may not construct the json the same everywhere)

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -60,14 +60,25 @@ class SessionService {
     );
   }
 
-  Future<ApiResponse<Session>> createSession({required String projectId}) {
+  Future<ApiResponse<Session>> createSessionWithMessage({
+    required String projectId,
+    required String text,
+    required String? agent,
+    required String? providerID,
+    required String? modelID,
```
AlexDochioiu: since it's both or none, change to `required PromptModel model` directly

mobile/app/test/capabilities/session/session_service_test.dart
```diff
@@ -265,7 +262,7 @@ void main() {
           equals({
             "projectId": "/tmp/project",
             "parts": [
-              {"type": "text", "text": "first prompt"},
+              {"text": "first prompt"},
```
AlexDochioiu: is this correct? Shouldn't the `type` exist always?

bridge/app/lib/src/bridge/routing/create_session_handler.dart
```diff
@@ -52,6 +52,21 @@ class CreateSessionHandler extends RequestHandler {
         .map(
           (p) => switch (p) {
```
AlexDochioiu: this mapping logic is duplicated. extract it in an extension fct and reuse it

shared/sesori_shared/lib/src/models/sesori/send_prompt_request.dart
```diff
@@ -15,11 +15,44 @@ sealed class SendPromptRequest with _$SendPromptRequest {
...
 }
 
+/// Prompt part types for the mobile ↔ bridge protocol.
+///
+/// These are OUR domain types — the OpenCode plugin converts them to
+/// whatever wire format OpenCode expects.
+///
+/// Note: a second variant is required to work around a Freezed codegen bug
+/// where single-variant unions omit the `type` discriminator in `toJson()`.
+/// See: https://github.com/rrousselGit/freezed/issues/1342
```
AlexDochioiu: Simplify those docs to just `Prompt part types for the mobile ↔ bridge protocol`

PR #63, opened 2026-03-30
refactor(mobile): type-safe navigation with sealed class AppRoute

AlexDochioiu:
## Summary

- Replaces the `AppRoute` enum with a sealed class hierarchy where each route subclass defines its required parameters as constructor arguments
- Navigation extensions (`goRoute`, `pushRoute`) now take an `AppRoute` instance directly — no more `Map<String, String>` params that could silently miss required keys
- Route comparisons in cubits use idiomatic `is` checks instead of enum equality

## Before

```dart
context.pushRoute(
  AppRoute.sessionDetail,
  pathParams: {"projectId": session.projectID, "sessionId": session.id},
  queryParams: {"title": session.title ?? ""},
);
```

## After

```dart
context.pushRoute(
  AppRoute.sessionDetail(
    projectId: session.projectID,
    sessionId: session.id,
    sessionTitle: session.title ?? "",
  ),
);
```

Compile-time enforcement — forgetting `sessionId` is now a type error, not a silent runtime bug.

## Files changed (16)

- **`module_core/.../app_routes.dart`** — sealed class + 6 subclasses with factory constructors
- **`app/.../app_router.dart`** — simplified extensions, exhaustive switch on sealed subtypes
- **`app/.../go_router_route_source.dart`** — route matching unchanged (uses `path` templates)
- **`module_core/.../auth_redirect_service.dart`** — returns `const AppRoute.login()` / `const AppRoute.projects()`
- **`module_core/.../project_list_cubit.dart`** — `is AppRouteProjects` instead of `== AppRoute.projects`
- **6 screen files** — all navigation calls use typed constructors
- **`app/.../notification_service.dart`** — builds paths from sealed class instances
- **4 test files** — updated for new API

### DIFF
--- a/mobile/app/lib/core/platform/notification_service.dart
+++ b/mobile/app/lib/core/platform/notification_service.dart
@@ -129,20 +129,21 @@ class NotificationService {
 
   void _pushSessionRoute({required String sessionId, required String? projectId}) {
     if (projectId == null) {
-      goForTesting(AppRoute.projects.path);
+      goForTesting(const AppRoute.projects().path);
       return;
     }
 
-    final sessionDetailPath = AppRoute.sessionDetail.buildPath(
-      pathParams: {"projectId": projectId, "sessionId": sessionId},
-    );
+    final sessionDetailPath = AppRoute.sessionDetail(
+      projectId: projectId,
+      sessionId: sessionId,
+    ).buildPath();
 
     if (currentPathProviderForTesting() == sessionDetailPath) {
       return;
     }
 
-    final sessionPath = AppRoute.sessions.buildPath(pathParams: {"projectId": projectId});
-    goForTesting(AppRoute.projects.path);
+    final sessionPath = AppRoute.sessions(projectId: projectId).buildPath();
+    goForTesting(const AppRoute.projects().path);
     unawaited(pushForTesting(sessionPath));
     unawaited(pushForTesting(sessionDetailPath));
   }
--- a/mobile/app/lib/core/platform/notification_service.dart
+++ b/mobile/app/lib/core/platform/notification_service.dart
@@ -129,21 +129,23 @@ class NotificationService {
 
   void _pushSessionRoute({required String sessionId, required String? projectId}) {
     if (projectId == null) {
-      goForTesting(const AppRoute.projects().path);
+      goForTesting(AppRouteDef.projects.path);
       return;
     }
 
     final sessionDetailPath = AppRoute.sessionDetail(
       projectId: projectId,
       sessionId: sessionId,
+      sessionTitle: null,
+      readOnly: false,
     ).buildPath();
 
     if (currentPathProviderForTesting() == sessionDetailPath) {
       return;
     }
 
-    final sessionPath = AppRoute.sessions(projectId: projectId).buildPath();
-    goForTesting(const AppRoute.projects().path);
+    final sessionPath = AppRoute.sessions(projectId: projectId, projectName: null).buildPath();
+    goForTesting(AppRouteDef.projects.path);
     unawaited(pushForTesting(sessionPath));
     unawaited(pushForTesting(sessionDetailPath));
   }
--- a/mobile/app/lib/core/routing/app_router.dart
+++ b/mobile/app/lib/core/routing/app_router.dart
@@ -10,31 +10,38 @@ import "../../features/session_list/session_list_screen.dart";
 import "../../features/settings/notification_settings_screen.dart";
 import "../di/injection.dart";
 
-extension AppRouteToGoRoute on AppRoute {
-  /// Returns the [GoRoute] for this route with an exhaustive builder switch.
-  ///
-  /// Adding a new [AppRoute] subclass without handling it here will cause a
-  /// compile-time error.
+extension AppRouteToGoRoute on AppRouteDef {
+  /// Returns the [GoRoute] for this route definition with an exhaustive
+  /// builder switch over decoded [AppRoute] values.
   GoRoute toGoRoute() {
     return GoRoute(
       path: path,
-      builder: (context, state) => switch (this) {
-        AppRouteLogin() => const LoginScreen(),
-        AppRouteProjects() => const ProjectListScreen(),
-        AppRouteNotificationSettings() => const NotificationSettingsScreen(),
-        AppRouteSessions() => SessionListScreen(
-          projectId: state.pathParameters["projectId"] ?? "",
-          projectName: state.uri.queryParameters["name"],
-        ),
-        AppRouteNewSession() => NewSessionScreen(
-          projectId: state.pathParameters["projectId"] ?? "",
-        ),
-        AppRouteSessionDetail() => SessionDetailScreen(
-          projectId: state.pathParameters["projectId"],
-          sessionId: state.pathParameters["sessionId"] ?? "",
-          sessionTitle: state.uri.queryParameters["title"],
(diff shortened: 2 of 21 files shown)


mobile/app/lib/core/platform/notification_service.dart
```diff
@@ -129,20 +129,21 @@ class NotificationService {
...
     if (currentPathProviderForTesting() == sessionDetailPath) {
       return;
     }
 
-    final sessionPath = AppRoute.sessions.buildPath(pathParams: {"projectId": projectId});
-    goForTesting(AppRoute.projects.path);
+    final sessionPath = AppRoute.sessions(projectId: projectId).buildPath();
+    goForTesting(const AppRoute.projects().path);
     unawaited(pushForTesting(sessionPath));
     unawaited(pushForTesting(sessionDetailPath));
```
AlexDochioiu: what are those? This is production code, not test code

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,137 @@
...
+  const factory AppRoute.login() = AppRouteLogin;
+  const factory AppRoute.projects() = AppRouteProjects;
+  const factory AppRoute.notificationSettings() = AppRouteNotificationSettings;
+  const factory AppRoute.sessions({required String projectId, String? projectName}) = AppRouteSessions;
+  const factory AppRoute.newSession({required String projectId}) = AppRouteNewSession;
+  const factory AppRoute.sessionDetail({
+    required String projectId,
+    required String sessionId,
+    String? sessionTitle,
+    bool readOnly,
```
AlexDochioiu:
```suggestion
    required String? sessionTitle,
    required bool readOnly,
```

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,137 @@
...
+
+  @override
+  String buildPath() => path;
+}
+
+class AppRouteSessions extends AppRoute {
+  final String projectId;
+  final String? projectName;
+
+  const AppRouteSessions({required this.projectId, this.projectName});
```
AlexDochioiu:
```suggestion
  const AppRouteSessions({required this.projectId, required this.projectName});
```

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,137 @@
...
+  final String projectId;
+  final String sessionId;
+  final String? sessionTitle;
+  final bool readOnly;
+
+  const AppRouteSessionDetail({
+    required this.projectId,
+    required this.sessionId,
+    this.sessionTitle,
+    this.readOnly = false,
```
AlexDochioiu:
```suggestion
    required this.sessionTitle,
    required this.readOnly,
```

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,137 @@
...
+  });
+
+  @override
+  String get path => "/projects/:projectId/sessions/:sessionId";
+
+  @override
+  String buildPath() {
+    final base = "/projects/${Uri.encodeComponent(projectId)}/sessions/${Uri.encodeComponent(sessionId)}";
+    final queryParams = <String, String>{};
+    if (sessionTitle != null) queryParams["title"] = sessionTitle!;
```
AlexDochioiu: avoid bang operator. You can just inline as

final queryParams = {
  "title": ?sessionTitle,
  "readOnly": readOnly.toString(),
};

p.s. let's always pass `readOnly` too, not just when `true`

mobile/app/lib/core/routing/app_router.dart
```diff
@@ -13,23 +13,23 @@ import "../di/injection.dart";
 extension AppRouteToGoRoute on AppRoute {
   /// Returns the [GoRoute] for this route with an exhaustive builder switch.
   ///
-  /// Adding a new [AppRoute] value without handling it here will cause a
+  /// Adding a new [AppRoute] subclass without handling it here will cause a
   /// compile-time error.
   GoRoute toGoRoute() {
```
AlexDochioiu, line 16: This should not be an extension fct anymore. We should define this directly as an abstract function to `AppRoute` that every route uses (without a switch statement).

Actually, based on where this is used I don't like that it takes an instance of `AppRoute` to build, since we create an "empty" route only to call the `toGoRoute`. Think of another option where we can apply it directly on the AppRoute type maybe without creating an instance only to call `toGoRoute` on it

--

L.e. I think we actually want a mix of sealed class + enum to avoid those. `AppRoute` stays as sealed class but there's also some `AppRouteEnum` which contains the `path`.

`AppRoute` instead of defining the `path` per child class, it defines some `matchingEnum` which will be the pair from the enum to the child class. After this, we can keep this extension and switch statement applying on the `AppRouteEnum`.

We should also stop building the AppRoute here in the extension fct. Instead, `AppRoute` and all the child classes should have some `fromState` factory method. The AppRoute one just redirects to the correct individual child class. Each child class manages reading the params from state. This will delegate to the child classes both the encoding to URI and decoding to URI, so that we don't split this logic that's basically two sides of the same coin. Less likely to get them falling out of sync this way.

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,137 @@
...
+  /// matching. Parameter values are irrelevant; only [path] templates matter.
+  static const values = <AppRoute>[
+    AppRouteLogin(),
+    AppRouteProjects(),
+    AppRouteNotificationSettings(),
+    AppRouteSessions(projectId: ""),
+    AppRouteNewSession(projectId: ""),
+    AppRouteSessionDetail(projectId: "", sessionId: ""),
+  ];
+}
```
AlexDochioiu: I am concerned this will fall out of sync. Any ideas on making this somehow compile-time guaranteed to include all AppRoute children?

mobile/app/lib/core/routing/deep_link_service.dart
```diff
@@ -52,7 +52,7 @@ class DeepLinkService {
       final route = await _authRedirectService.handleOAuthCallback(uri);
       if (route != null) {
         try {
-          appRouter.goRoute(route);
```
AlexDochioiu, line 55: I want to keep on using `goRoute`/etc and ensure compile-time safety.

mobile/module_core/lib/src/routing/app_routes.dart
```diff
@@ -1,35 +1,207 @@
...
+
+  @override
+  AppRouteDef get def => AppRouteDef.sessionDetail;
+
+  @override
+  String buildPath() {
+    final base = "/projects/${Uri.encodeComponent(projectId)}/sessions/${Uri.encodeComponent(sessionId)}";
+    final queryParams = <String, String>{
+      "readOnly": readOnly.toString(),
+      "title": ?sessionTitle,
```
AlexDochioiu: move all the keys in file to `static const _titleQueryParam = "title"` and so forth

mobile/app/lib/core/routing/deep_link_service.dart
```diff
@@ -52,7 +52,7 @@ class DeepLinkService {
       final route = await _authRedirectService.handleOAuthCallback(uri);
```
AlexDochioiu, line 52: handleOAuthCallback should return `AppRoute` directly rather than `AppRouteDef`

PR #70, opened 2026-03-31
feat(bridge): add distribution and release pipeline

stanvladut:
## Summary
- add bridge installers, self-update components, and CLI wiring
- add GitHub release workflow with multi-platform build and npm publish jobs
- add versioning, update cache, checksum verification, and supporting tests

## Testing
- dart test
- dart analyze (infos only)
- shellcheck install.sh
- bash -n install.sh


stanvladut:

bridge/app/lib/src/updater/models/cached_release.dart
```diff
@@ -0,0 +1,42 @@
+/// Plain Dart model for cached release information with TTL support.
+class CachedRelease {
```
AlexDochioiu: Use freezed with auto-generated toJson and fromJson

bridge/app/lib/src/updater/models/release_info.dart
```diff
@@ -0,0 +1,42 @@
+/// Plain Dart model representing a newer release available for download.
+class ReleaseInfo {
```
AlexDochioiu: Use freezed with auto-generated toJson and fromJson

bridge/app/bin/bridge.dart
```diff
@@ -272,14 +311,41 @@ Future<void> main(List<String> args) async {
 
   // Run bridge; stop the server process on exit regardless of outcome
   try {
-    await session.run();
+    final runFuture = session.run();
+    updateOrchestrator.startPeriodicCheck(interval: const Duration(hours: 4));
+    final updateSub = updateOrchestrator.updateAvailable.listen((version) {
```
AlexDochioiu: Add this class in the dart shared module + include tests:

```dart
import "dart:async";

import "package:rxdart/rxdart.dart";

/// A [Stream] that converts a single-subscription Stream into a broadcast Stream that replays
/// the latest value to any new listener while subscribed (if [replayLastValueToNewListeners] is true.
/// Once all subscribers have canceled or the stream completes, the underlying subscription is canceled.
///
/// Adding listeners after the stream has completed/disposed and the [delayBeforeCancel] time has passed
/// will result in a new broadcast stream being created with [_factory] and shared to new listeners.
class RefCountReusableStream<T> extends Stream<T> {
  final bool replayLastValueToNewListeners;
  final Duration delayBeforeCancel;
  final void Function()? onCancel;
  final Stream<T> Function() _factory;
  StreamSubscription<T>? _subscription;
  StreamController<T>? _controller;

  RefCountReusableStream._(
    this._factory, {
    required this.replayLastValueToNewListeners,
    required this.delayBeforeCancel,
    this.onCancel, // called after delayBeforeCancel once all listeners have canceled (can be called more than once)
  });

  /// Creates a [RefCountReusableStream] that will create a new broadcast stream using [_factory] and
  /// share it to new listeners. New listeners WILL NOT receive the last value emitted by the stream.
  RefCountReusableStream.publish(
    Stream<T> Function() factory, {
    Duration delayBeforeCancel = Duration.zero,
    void Function()? onCancel,
  }) : this._(
         factory,
         replayLastValueToNewListeners: false,
         delayBeforeCancel: delayBeforeCancel,
         onCancel: onCancel,
       );

  /// Creates a [RefCountReusableStream] that will create a new broadcast stream using [_factory] and
  /// share it to new listeners. New listeners WILL receive the last value emitted by the stream.
  RefCountReusableStream.behaviour(
    Stream<T> Function() factory, {
    Duration delayBeforeCancel = Duration.zero,
    void Function()? onCancel,
  }) : this._(
         factory,
         replayLastValueToNewListeners: true,
         delayBeforeCancel: delayBeforeCancel,
         onCancel: onCancel,
       );

  Timer? _disposeTask;

  void _requestDispose() {
    _disposeTask ??= Timer(delayBeforeCancel, () {
      final onCancel = this.onCancel;
      unawaited(_subscription?.cancel());
      if (onCancel != null) onCancel();
      _subscription = null;
      _controller = null;
      _disposeTask = null;
    });
  }

  @override
  StreamSubscription<T> listen(
    void Function(T)? onData, {
    Function? onError,
    void Function()? onDone,
    bool? cancelOnError,
  }) {
    _disposeTask?.cancel();
    _disposeTask = null;

    final controller = _controller;
    if (controller == null) {
      final newController = replayLastValueToNewListeners
          ? BehaviorSubject<T>(onCancel: _requestDispose, sync: true)
          : PublishSubject<T>(onCancel: _requestDispose, sync: true);
      _controller = newController;
      _subscription = _factory().listen(newController.add, onError: newController.addError, onDone: newController.close);
      return newController.stream.listen(onData, onError: onError, onDone: onDone, cancelOnError: cancelOnError);
    }

    return controller.stream.listen(onData, onError: onError, onDone: onDone, cancelOnError: cancelOnError);
  }
}
```

After this, use it for the `updateAvailable` stream. When there's a listener to it, it automatically checks every 4 hours for updates using a PeriodicStream (seeded so that it emits when created too, rather than emitting after delay its first emision). 

Add this stream extension if we don't have it in shared code too
(shared/sesori_shared/lib/src/extensions/sugar_dart.dart)
```dart
Stream<T> seeded(T initialValue) => ConcatStream([Stream.value(initialValue), this]);
```

--

Add also tests for this stream but I believe implementation is correct cause I am using it widely in another project.

By using this we get rid of explicit checker start via `updateOrchestrator.startPeriodicCheck` while also supporting multiple subscribers and re-subscribe (if needed)

bridge/app/bin/bridge.dart
```diff
@@ -302,6 +369,7 @@ Future<void> _shutdown(
 
   bandwidthTracker?.dispose();
   httpClient.close();
+  updateOrchestrator.dispose();
```
AlexDochioiu: I think we'll need just to stop listening to the updated stream to stop the checker?

bridge/app/lib/src/updater/platform_info.dart
```diff
@@ -0,0 +1,162 @@
...
+/// Throws [ArgumentError] for unsupported platform combinations.
+String getAssetName({required String os, required String arch}) {
+  final key = '$os $arch';
+  const assetMap = {
+    'macos arm64': 'sesori-bridge-macos-arm64.tar.gz',
+    'macos x64': 'sesori-bridge-macos-x64.tar.gz',
+    'linux x64': 'sesori-bridge-linux-x64.tar.gz',
+    'linux arm64': 'sesori-bridge-linux-arm64.tar.gz',
+    'windows x64': 'sesori-bridge-windows-x64.zip',
+  };
```
AlexDochioiu: Can't we upload the executable without archiving it (via github action)? Isn't it a single file? Would simplify a lot the PR and process if we get rid of the un-archiving/etc.

bridge/app/lib/src/updater/platform_info.dart
```diff
@@ -0,0 +1,162 @@
...
+    // This is a simplified detection; in production, you might use dart:ffi more directly
+    if (Platform.operatingSystem == 'macos' || Platform.operatingSystem == 'linux') {
+      // For macOS and Linux, we can use the version string or environment
+      // For now, default to x64 as a fallback (most common)
+      return 'x64';
+    }
+    return 'x64';
+  } on Object {
+    return 'x64'; // Default fallback
+  }
```
AlexDochioiu: what's the value here? This always return x64

bridge/app/lib/src/updater/platform_info.dart
```diff
@@ -0,0 +1,162 @@
...
+    default:
+      return Platform.operatingSystem;
+  }
+}
+
+/// Detects the current CPU architecture from [Platform.version].
+///
+/// Returns `'arm64'` or `'x64'`. Defaults to `'x64'` when the architecture
+/// cannot be determined from the version string.
+String currentArch() {
```
AlexDochioiu: Seems similar to `_detectArchitecture`. remove pointless duplication

bridge/app/lib/src/updater/release_repository.dart
```diff
@@ -0,0 +1,117 @@
...
+  final String _currentVersion;
+  final String _os;
+  final String _arch;
+
+  ReleaseRepository({
+    required GitHubReleasesApi api,
+    required UpdateCache cache,
+    required String currentVersion,
+    String? os,
+    String? arch,
```
AlexDochioiu: make those required, and create and use enums instead of String

bridge/app/lib/src/updater/self_updater.dart
```diff
@@ -0,0 +1,184 @@
...
+    required ChecksumVerifier checksumVerifier,
+  }) : _downloadApi = downloadApi,
+       _extractor = extractor,
+       _checksumVerifier = checksumVerifier,
+       _updateLock = UpdateLock(
+         currentPid: pid,
+         cleanup: _cleanup,
+         isPermissionDenied: _isPermissionDenied,
+       ),
+       _fileReplacer = FileReplacer();
```
AlexDochioiu: Inject this in construction for better unit testing.

bridge/app/lib/src/updater/self_updater.dart
```diff
@@ -0,0 +1,184 @@
...
+  final FileReplacer _fileReplacer;
+
+  SelfUpdater({
+    required UpdateDownloadApi downloadApi,
+    required ArchiveExtractor extractor,
+    required ChecksumVerifier checksumVerifier,
+  }) : _downloadApi = downloadApi,
+       _extractor = extractor,
+       _checksumVerifier = checksumVerifier,
+       _updateLock = UpdateLock(
```
AlexDochioiu: This should either be injected in constructor, or a factory should be created and injected so that you can test `SelfUpdater` in isolation.

bridge/app/lib/src/updater/services/update_installer_service.dart
```diff
@@ -0,0 +1,184 @@
...
+    final String archivePath = p.join(
+      installRoot,
+      Platform.isWindows ? '.sesori-bridge-update.zip' : '.sesori-bridge-update.tar.gz',
+    );
+    final String stagingPath = p.join(installRoot, '.sesori-bridge-staging');
+    final File lockFile = File(lockPath);
+    bool lockAcquired = false;
+
+    try {
+      final LockAcquireResult lockResult = await _updateLock.acquireLock(lockFile: lockFile);
```
AlexDochioiu: I don't see us ever releasing this lock? I'd rather see something like

```dart
final UpdateResult result = _updateLock.locked<UpdateResult>(
  lockFile: lockFile,
  onLockAcquired: () async => {}, // returns Future<UpdateResult>
  onLockRejected: (reason) async => {}, // returns Future<UpdateResult>
);
```

and this should auto-release the lock after `onLockAcquired` finishes.

bridge/app/lib/src/updater/services/update_installer_service.dart
```diff
@@ -0,0 +1,184 @@
...
+        case FileSystemEntityType.directory:
+          Directory(path).deleteSync(recursive: recursive);
+        case FileSystemEntityType.link:
+          Link(path).deleteSync();
+        case FileSystemEntityType.unixDomainSock:
+        case FileSystemEntityType.pipe:
+        case FileSystemEntityType.notFound:
+      }
+    } on Object {
+      // Best-effort cleanup.
```
AlexDochioiu: always log some warning/debug/etc message when catching unexpected errors.

bridge/app/lib/src/updater/update_cache.dart
```diff
@@ -0,0 +1,99 @@
...
+
+    final CachedRelease cached;
+    try {
+      cached = CachedRelease.fromJson(decodedJson);
+    } on Object catch (error) {
+      if (error is FormatException || error is TypeError) {
+        return null;
+      }
+      rethrow;
+    }
```
AlexDochioiu: check how we use `jsonDecodeMap` in other places cause it's a lot more concise.

bridge/app/lib/src/updater/update_lock.dart
```diff
@@ -0,0 +1,105 @@
...
+      }
+    } on FileSystemException catch (error) {
+      if (_isPermissionDenied(error: error)) {
+        return LockAcquireResult.permissionDenied;
+      }
+      rethrow;
+    }
+  }
+
+  Future<bool> removeStaleLockIfNeeded({required File lockFile}) async {
```
AlexDochioiu: Maybe we don't need this managed externally. We can just do "if last lock was acquired more than 12 hours ago, release it and acquire new lock". Since we do the callback-based approach with self-release of the lock after callback, this should be a simpler implementation that is easier to follow/etc. Also the 12 hours should be way more than enough in case there is a bug or the old cleanup failed for some reason.

We can even do a simpler in-memory lock that doesn't require cleanup at all. A single instance of the lock that rejects re-entry when a lock is acquired. This will work pretty much always and it will avoid separate platform considerations.

bridge/app/tool/bump_version.dart
```diff
@@ -0,0 +1,188 @@
...
+  final formatted = const JsonEncoder.withIndent('  ').convert(json);
+  await writeFile(path: path, content: '$formatted\n');
+}
+
+Future<void> main(final List<String> args) async {
+  if (args.isEmpty) {
+    stderr.writeln('Usage: dart run tool/bump_version.dart <version>');
+    stderr.writeln('Example: dart run tool/bump_version.dart 0.3.0');
+    exit(1);
+  }
```
AlexDochioiu, line 107: does it make sense to include this in the `makefile`? Not 100% sure if this should be normally there.

bridge/app/npm/sesori-bridge/lib/bootstrap.js
```diff
@@ -0,0 +1,273 @@
...
+      shellPath: process.env.SHELL || "",
+    });
+    if (launcherResult && launcherResult.message) {
+      console.error(launcherResult.message);
+    }
+  } catch (error) {
+    fail(
+      "sesori-bridge: Failed to persist the managed command path.\n" +
+      String(error && error.message ? error.message : error)
+    );
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 9260d87. PATH persistence failures now log a warning and the managed runtime still launches, so constrained environments do not lose a valid bootstrap.

install.ps1
```diff
@@ -0,0 +1,184 @@
...
+        }
+        if ($release.draft -or $release.prerelease) {
+            continue
+        }
+
+        $asset = $release.assets | Where-Object { $_.name -eq $ArchiveName } | Select-Object -First 1
+        $checksums = $release.assets | Where-Object { $_.name -eq 'checksums.txt' } | Select-Object -First 1
+        if ($asset -and $checksums) {
+            $eligible += [pscustomobject]@{
+                Version = [version]($release.tag_name.Substring(8))
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 1dbd016. The PowerShell installer now validates `bridge-v*` tags with `TryParse` and skips malformed/non-stable entries instead of aborting the whole release resolution path.

bridge/app/lib/src/updater/repositories/release_repository.dart
```diff
@@ -0,0 +1,142 @@
...
+    required DistributionTarget target,
+  }) : _api = api,
+       _cache = cache,
+       _currentVersion = currentVersion,
+       _target = target;
+
+  Future<ReleaseInfo?> checkForNewerRelease() async {
+    final cached = await _cache.read(ttl: const Duration(hours: 24));
+    if (cached != null) {
+      return _evaluateCached(cached: cached);
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed across 656e1cb and e654f30. Cached releases now carry `assetName`, the repository validates cache entries against the current target before trusting them, and target mismatches fall through to a fresh HTTP lookup.

install.sh
```diff
@@ -0,0 +1,308 @@
...
+def compare_versions(a: str, b: str) -> int:
+    def split_pre(v: str):
+        if "-" in v:
+            core, pre = v.split("-", 1)
+        else:
+            core, pre = v, None
+        return core, pre
+
+    def parse_core(v: str):
+        return [int(part) for part in v.split(".")]
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 1dbd016. The Unix installer now filters release tags to stable `X.Y.Z` versions before sorting, so malformed `bridge-v*` tags no longer abort installation.

bridge/app/lib/src/updater/models/github_release_dto.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'github_release_dto.freezed.dart';
+part 'github_release_dto.g.dart';
+
+@Freezed(fromJson: true, toJson: false)
+sealed class GitHubReleaseDto with _$GitHubReleaseDto {
+  const factory GitHubReleaseDto({
+    @JsonKey(name: 'tag_name') required String tagName,
+    @JsonKey(name: 'published_at') required String publishedAt,
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. `published_at` is now nullable in the GitHub release DTO, and the updater skips draft/null cases without breaking valid stable release checks.

bridge/app/lib/src/updater/api/update_cache_api.dart
```diff
@@ -0,0 +1,67 @@
...
+    final dir = Directory(cacheDirectory);
+    if (!dir.existsSync()) {
+      await dir.create(recursive: true);
+    }
+
+    final tmpFile = File(p.join(cacheDirectory, 'update_cache.json.tmp'));
+    await tmpFile.writeAsString(jsonEncode(release.toJson()));
+
+    final cacheFile = File(_cacheFilePath);
+    await tmpFile.rename(cacheFile.path);
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. The cache writer now deletes the previous cache file before renaming the temp file into place, which keeps Windows cache rewrites working.

bridge/app/npm/sesori-bridge/lib/bootstrap.js
```diff
@@ -0,0 +1,130 @@
...
+        beforeInstallSwap: function() {
+          recordInstallAttempt();
+          sleepForTest();
+          if (process.env.SESORI_BRIDGE_TEST_WRITE_FAIL === "1") {
+            throw new Error("Simulated managed install write failure.");
+          }
+        },
+      });
+    } catch (error) {
+      fail(
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 9260d87. The bootstrap failure paths now throw instead of calling `process.exit` inside the lock callback, so the lock wrapper gets to clean up `.sesori-bootstrap.lock` normally.

bridge/app/lib/src/updater/api/file_replacement_api.dart
```diff
@@ -0,0 +1,221 @@
...
+      '    } catch [System.UnauthorizedAccessException] {',
+      r'      if ((Get-Date) -ge $deadline) { throw }',
+      '      Start-Sleep -Milliseconds 200',
+      '    }',
+      '  }',
+      '}',
+      '\$args = @($escapedArgs)',
+      r'Wait-ForUnlockedFile -Path $binaryPath -TimeoutSeconds 30',
+      r'if (Test-Path $binaryPath) { Move-Item -Force $binaryPath $oldBinaryPath }',
+      r'Move-Item -Force $newBinaryPath $binaryPath',
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. The Windows swap script now wraps the binary/lib moves in a rollback-aware try/catch and restores the previous binary/lib directory if the staged move fails.

install.ps1
```diff
@@ -0,0 +1,196 @@
...
+$ManagedManifest = Join-Path $InstallRoot '.managed-runtime.json'
+
+# ── Architecture detection ────────────────────────────────────────────────────
+$arch = $null
+try {
+    $procArch = [System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture
+    switch ($procArch.ToString()) {
+        'X64'   { $arch = 'x64' }
+        'Amd64' { $arch = 'x64' }
+        default { $arch = $null }
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 1dbd016. Windows install support is now determined from OS architecture / WOW64-aware signals rather than only the current PowerShell process architecture.

install.ps1
```diff
@@ -0,0 +1,196 @@
...
+    Expand-Archive -Path $TempZip -DestinationPath $InstallRoot -Force
+
+    # ── Verify binary ─────────────────────────────────────────────────────────
+    $BinaryPath = Join-Path $BinDir $BinaryName
+    if (-not (Test-Path $BinaryPath)) {
+        Write-Error "Expected binary not found at '$BinaryPath' after extraction. Check archive structure."
+        exit 1
+    }
+
+    @{ version = $Release.TagName.Substring(8) } | ConvertTo-Json -Compress | Set-Content -Path $ManagedManifest -Encoding UTF8
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 1dbd016 and 9260d87. The PowerShell installer now writes `.managed-runtime.json` without a BOM, and the npm bootstrap reader also strips a UTF-8 BOM on read as a defensive fallback.

bridge/app/lib/src/updater/foundation/update_lock.dart
```diff
@@ -0,0 +1,200 @@
...
+      final Duration age = DateTime.now().difference(lastModified);
+      if (age < _invalidLockGracePeriod) {
+        return LockAcquireResult.alreadyLocked;
+      }
+      return _deleteStaleLock(lockFile: lockFile);
+    }
+
+    final bool isAlive = await isProcessAlive(pidToCheck: existingPid);
+    if (isAlive) {
+      return LockAcquireResult.alreadyLocked;
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. The update lock now persists owner identity beyond PID alone by recording a process marker and validating it before treating an existing lock as live.

bridge/app/lib/src/updater/api/file_replacement_api.dart
```diff
@@ -0,0 +1,221 @@
...
+    try {
+      if (targetBinary.existsSync()) {
+        targetBinary.renameSync(backupBinary.path);
+        movedTargetBinary = true;
+      }
+      newBinary.renameSync(targetBinary.path);
+      movedNewBinary = true;
+
+      if (newLibDir.existsSync()) {
+        if (targetLibDir.existsSync()) {
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. Unix replacement now rejects staged updates unless both the binary and lib runtime directory are present, instead of silently treating missing libs as optional.

bridge/app/lib/src/updater/api/file_replacement_api.dart
```diff
@@ -0,0 +1,221 @@
...
+    required String stagingPath,
+  }) async {
+    const String binaryName = 'sesori-bridge.exe';
+    final File newBinary = File(p.join(stagingPath, 'bin', binaryName));
+
+    if (!newBinary.existsSync()) {
+      return const FileReplacementResult.failure();
+    }
+
+    return FileReplacementResult.pending(
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 656e1cb. `replaceWindows` now enforces the same staged runtime contract as Unix by requiring both `bin/sesori-bridge.exe` and `lib/` before reporting a pending success.

shared/sesori_shared/lib/src/streams/ref_count_reusable_stream.dart
```diff
@@ -0,0 +1,101 @@
...
+    _disposeTask?.cancel();
+    _disposeTask = null;
+
+    final controller = _controller;
+    if (controller == null) {
+      final newController = replayLastValueToNewListeners
+          ? BehaviorSubject<T>(onCancel: _requestDispose, sync: true)
+          : PublishSubject<T>(onCancel: _requestDispose, sync: true);
+      _controller = newController;
+      _subscription = _factory().listen(
```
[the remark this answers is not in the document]
AlexDochioiu: I did not address this one in this pass. The current work was scoped to bridge distribution/bootstrap/updater correctness plus the install/release docs; this shared-stream issue is outside those touched areas and should be fixed as a separate shared runtime change.

PR #73, opened 2026-04-01
feat: add slash command support to mobile sessions

daniil-shumko:
## Summary

- Add full slash command pipeline: OpenCode API → bridge plugin → relay → mobile UI
- Mobile users can browse and execute slash commands (skills, MCP tools, built-in commands) from a searchable bottom sheet in both new session and session detail screens
- Commands can also be launched from the session list via long-press context menu

### Bridge
- **`GetCommandsHandler`** — new `GET /command` route that fetches commands from the OpenCode plugin and returns them as `CommandListResponse`
- **`SendCommandHandler`** — new `POST /session/:id/command` route that forwards command execution to OpenCode
- **Plugin interface** — added `getCommands()` and `sendCommand()` to `BridgePlugin` contract
- **Template stripping** — bridge omits the `template` field (full prompt text) from the mobile response, reducing payload from ~424KB to ~23KB (95% reduction)

### Mobile
- **`CommandPickerSheet`** — searchable bottom sheet that displays available slash commands grouped by source (command/skill/MCP)
- **`PromptInput`** — slash button opens the command picker; staged command shows as a chip with argument hints
- **`NewSessionCubit`** — fetches commands on load, supports creating sessions via command execution
- **`SessionDetailCubit`** — fetches commands alongside session data, supports sending commands in existing sessions
- **`QueuedSessionSubmission`** — sealed class replacing raw strings in the send queue, distinguishing prompts from commands
- **`SessionLaunchCommandStore`** — passes staged commands from session list → new session screen

### Shared
- **`CommandInfo`**, **`CommandListResponse`**, **`SendCommandRequest`** — new Freezed models in `sesori_shared`
- **`SesoriSseEvent`** — added `mcpToolsChanged` variant for real-time command list updates

### Bug Fix
- **Root cause**: OpenCode returns `template` as `{}` (empty map) for MCP tool commands instead of a string, causing `Command.fromJson` to throw a type cast error that crashed the entire command list
- **Fix**: custom `@JsonKey(readValue:)` coerces non-string template values to null; all template/hints fields made nullable with safe defaults

## Testing
- Bridge: 508 tests pass, 0 analysis errors
- Plugin: 100 tests pass
- Mobile module_core: 157 tests pass, 0 analysis errors
- Mobile app: 0 analysis errors
- Shared: 171 tests pass, 0 analysis errors
- Verified end-to-end via bridge debug server: 80 commands returned successfully


bridge/app/lib/src/bridge/routing/get_commands_handler.dart
```diff
@@ -0,0 +1,26 @@
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "../plugin_to_shared_mapping.dart";
+import "request_handler.dart";
+
+const _projectIdHeader = "x-project-id";
+
+/// Handles `GET /command` — returns slash commands available to the project.
```
AlexDochioiu: turn it into a post command please and use `ProjectIdRequest` as a request body instead of this header. There's multiple other request handlers following this pattern already to look for reference.

bridge/app/lib/src/bridge/routing/send_command_handler.dart
```diff
@@ -0,0 +1,41 @@
...
+
+/// Handles `POST /session/:id/command` — executes a slash command in a session.
+class SendCommandHandler extends BodyRequestHandler<SendCommandRequest, SuccessEmptyResponse> {
+  final BridgePlugin _plugin;
+
+  SendCommandHandler(this._plugin)
+    : super(
+        HttpMethod.post,
+        "/session/:id/command",
+        fromJson: SendCommandRequest.fromJson,
```
AlexDochioiu: Change this to `/session/command` and move the session id inside `SendCommandRequest` please

bridge/sesori_plugin_opencode/lib/src/models/command.dart
```diff
@@ -0,0 +1,35 @@
...
+
+@Freezed(fromJson: true, toJson: true)
+sealed class Command with _$Command {
+  const factory Command({
+    required String name,
+    @JsonKey(readValue: _readTemplate) String? template,
+    @Default(<String>[]) List<String> hints,
+    String? description,
+    String? agent,
+    String? model,
```
AlexDochioiu, line 29: I think usually there's both model and provider, never only one. Can you double check if this is correct as is? Or was the `provider` forgotten?

P.s. provider = openai / claude / etc ; model = gpt-5.3codex / opus4.6 / etc

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -67,6 +69,22 @@ class OpenCodeApi {
...
 
+  Future<List<Command>> listCommands({required String? directory}) async {
+    final queryParameters = <String, String>{
+      if (directory != null && directory.isNotEmpty) "directory": directory,
+    };
+    final response = await http.get(
+      Uri.parse("$serverURL/command").replace(
+        queryParameters: queryParameters.isEmpty ? null : queryParameters,
+      ),
+      headers: _authHeaders,
```
AlexDochioiu: Let's just change to:
```
headers: {
  ..._authHeaders,
  _directoryOpenCodeHeader: ?directory,
}
```
The `directory` query param actually does the same thing like the header. Let's stick with the header everywhere.

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -264,6 +282,28 @@ class OpenCodeApi {
     }
   }
 
+  Future<void> sendCommand({
+    required String sessionId,
+    required SendCommandBody body,
+    required String? directory,
+  }) async {
+    final client = http.Client();
```
AlexDochioiu: Maybe for a separate task, but we shouldn't create a new client per request.

bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
```diff
@@ -242,13 +270,15 @@ class OpenCodePlugin implements BridgePlugin {
...
     final body = SendPromptBody(parts: parts, agent: agent, model: model);
 
-    await _call(
-      () => _service.repository.api.sendPrompt(
-        sessionId: session.id,
-        directory: session.directory,
-        body: body,
-      ),
-    );
+    if (parts.isNotEmpty) {
```
AlexDochioiu: I'd say parts.isEmpty should be an error scenario. We shouldn't do this ever? Is this being done because of the command?

mobile/app/lib/features/new_session/new_session_screen.dart
```diff
@@ -26,13 +26,15 @@ class NewSessionScreen extends StatelessWidget {
         sessionService: getIt<SessionService>(),
         projectId: projectId,
       ),
-      child: const _NewSessionBody(),
+      child: _NewSessionBody(projectId: projectId),
```
AlexDochioiu, line 29: nice fix 👌🏻

mobile/app/lib/features/new_session/new_session_screen.dart
```diff
@@ -154,14 +151,26 @@ class _NewSessionBodyState extends State<_NewSessionBody> {
...
                   dedicatedWorktree: _dedicatedWorktree,
                 );
               },
+              onSendCommand: (command, arguments) {
+                context.read<NewSessionCubit>().createSessionWithCommand(
+                  command: command,
+                  arguments: arguments,
+                  dedicatedWorktree: _dedicatedWorktree,
+                );
+              },
```
AlexDochioiu: can we collapse this into a single createSession? To me it feels like it's the same thing. We basically create a new session and we may send a) only text OR b) command + text . Isn't `arguments` basically the same as `text` more or less? So we should have a `createSession` that takes mandatory text and optional command. On the bridge side there should be a single request interceptor for both as well.

mobile/app/lib/features/session_detail/widgets/prompt_input.dart
```diff
@@ -23,8 +31,14 @@ class PromptInput extends StatefulWidget {
   const PromptInput({
     super.key,
     required this.isBusy,
-    required this.onSend,
+    required this.onSendPrompt,
+    required this.onSendCommand,
```
AlexDochioiu: I don't want a split. It's on send and the command is optional.

mobile/app/lib/features/session_detail/widgets/queued_message_bubble.dart
```diff
@@ -1,14 +1,15 @@
 import "package:flutter/material.dart";
+import "package:sesori_dart_core/sesori_dart_core.dart";
 
 import "../../../core/extensions/build_context_x.dart";
 
 class QueuedMessageBubble extends StatelessWidget {
```
AlexDochioiu, line 6: Is this still used? If not, please delete it entirely.

mobile/app/lib/features/session_detail/session_detail_screen.dart
```diff
@@ -301,16 +306,25 @@ class _SessionDetailBodyState extends State<_SessionDetailBody> {
                   ),
                 PromptInput(
                   isBusy: sessionStatus is! SessionStatusIdle,
-                  onSend: (text) => context.read<SessionDetailCubit>().sendMessage(text),
+                  onSendPrompt: (text) => context.read<SessionDetailCubit>().sendMessage(text),
+                  onSendCommand: (command, arguments) => context.read<SessionDetailCubit>().sendCommand(
+                    command: command.name,
+                    arguments: arguments,
+                  ),
```
AlexDochioiu: merge into a single `onSend`

mobile/app/lib/features/session_detail/session_detail_screen.dart
```diff
@@ -595,7 +612,7 @@ class _ErrorView extends StatelessWidget {
 // -----------------------------------------------------------------------------
 
 class _QueuedMessagesSection extends StatelessWidget {
```
AlexDochioiu, line 731: this still is used? if not delete it and track the chain of unused stuff to delete everything not used anymore.

mobile/module_core/lib/src/capabilities/server_connection/models/sse_event.dart
```diff
@@ -20,6 +20,7 @@ class SseEvent {
     SesoriSessionStatus(:final sessionID) => sessionID,
     // ignore: deprecated_member_use, retained for backward-compatible SSE payloads
     SesoriSessionIdle(:final sessionID) => sessionID,
+    SesoriCommandExecuted(:final sessionID) => sessionID,
```
AlexDochioiu, line 23: Is there / should there be UI associated with this?

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -32,6 +32,15 @@ class SessionService {
     );
   }
 
+  Future<ApiResponse<CommandListResponse>> listCommands({required String? projectId}) {
+    final headers = projectId == null || projectId.isEmpty ? null : {"x-project-id": projectId};
+    return _client.get(
+      "/command",
```
AlexDochioiu: Change to `post` request with body as suggested in handler.

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -61,6 +70,25 @@ class SessionService {
...
+      fromJson: Session.fromJson,
+      body: CreateSessionRequest(
+        projectId: projectId,
+        parts: const [],
+        agent: agent,
+        model: model,
+        dedicatedWorktree: dedicatedWorktree,
+      ),
+    );
+  }
```
AlexDochioiu: This **should not be possible**. No session must be created empty. Sessions get created with a first message + optional command/etc

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -208,6 +236,18 @@ class SessionService {
...
 
+  Future<ApiResponse<void>> sendCommand({
+    required String sessionId,
+    required String command,
+    required String arguments,
+  }) {
+    return _client.post(
+      "/session/$sessionId/command",
+      fromJson: SuccessEmptyResponse.fromJson,
+      body: SendCommandRequest(command: command, arguments: arguments),
```
AlexDochioiu: update as requested in handler. Although tbh maybe we should even just use the existing API to send a message + include there optional command. Not sure we want a new API for this.

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
```diff
@@ -186,9 +232,66 @@ class NewSessionCubit extends Cubit<NewSessionState> {
...
+            selectedAgent: current?.agent,
+            selectedProviderID: current?.providerID,
+            selectedModelID: current?.modelID,
+            stagedCommand: current?.stagedCommand,
+          ),
+        );
+    }
+  }
+
+  Future<void> createSessionWithCommand({
```
AlexDochioiu: don't think we need a new function. Just use the existing one and add some optional command there. Also don't manually create a session first before sending the command. This should not be mobile resposibility.

mobile/module_core/lib/src/cubits/session_detail/prompt_send_service.dart
```diff
@@ -61,12 +63,41 @@ class PromptSendService {
...
       _onQueueChanged();
     }
   }
 
-  String? cancelQueuedMessage(int index) {
+  Future<void> sendCommand({
+    required String command,
+    required String arguments,
+    required bool isConnected,
+  }) async {
```
AlexDochioiu: No dedicated sendCommand

mobile/module_core/lib/src/cubits/session_detail/prompt_send_service.dart
```diff
@@ -89,24 +120,31 @@ class PromptSendService {
...
+          agent: current.agent,
+          providerID: current.providerID,
+          modelID: current.modelID,
+        ),
+        QueuedCommandSubmission(:final command, :final arguments) => _service.sendCommand(
+          sessionId: _sessionId,
+          command: command,
+          arguments: arguments,
+        ),
+      };
```
AlexDochioiu: No dedicated API for send command. And no dedicated `QueuedCommandSubmission` model. A shared one with optional command is enough.

mobile/module_core/lib/src/cubits/session_detail/queued_session_submission.dart
```diff
@@ -0,0 +1,34 @@
+sealed class QueuedSessionSubmission {
+  const QueuedSessionSubmission();
+
+  String get displayText;
+  bool get isCommand;
+}
```
AlexDochioiu: Just a simple class with `String text` and `String? command` is enough

mobile/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -244,24 +261,29 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
       providersResponse,
-    ) = await wait6(
-      _service.getMessages(_sessionId),
-      _service.getPendingQuestions(_sessionId),
-      _service.getChildren(_sessionId),
-      _service.getSessionStatuses(),
-      _service.listAgents(),
-      _service.listProviders(),
-    );
+    ) = baseResponses;
```
AlexDochioiu: just do seven of them with await (_args_).wait and directly place them in individual variables.

mobile/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -302,9 +331,21 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
 
+  CommandInfo? _resolveStagedCommand({
+    required List<CommandInfo> availableCommands,
+    required CommandInfo? stagedCommand,
+  }) {
+    if (stagedCommand == null) return null;
+    for (final command in availableCommands) {
+      if (command.name == stagedCommand.name) return command;
+    }
+    return null;
```
AlexDochioiu: use firstWhereOrNull ?

mobile/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -655,6 +699,18 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
     );
   }
 
+  Future<void> sendCommand({
```
AlexDochioiu: no dedicated send command anywhere

mobile/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -777,10 +833,35 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
   }
 
+  Future<void> _consumeLaunchCommand() async {
+    final action = _launchCommandStore.take(_sessionId);
+    if (action == null) return;
+    await sendCommand(
+      command: action.command.name,
+      arguments: action.arguments,
+    );
+  }
```
AlexDochioiu: don't think we need this too

mobile/module_core/lib/src/cubits/session_detail/session_launch_command_store.dart
```diff
@@ -0,0 +1,36 @@
...
+  final CommandInfo command;
+  final String arguments;
+
+  const SessionLaunchCommand({
+    required this.command,
+    required this.arguments,
+  });
+}
+
+class SessionLaunchCommandStore {
```
AlexDochioiu: Don't think we need this at all anymore if we merge the command in the message API. Not sure if `SessionLaunchCommand` is needed either.

shared/sesori_shared/lib/src/models/sesori/command_info.dart
```diff
@@ -0,0 +1,29 @@
...
+/// Represents an available slash command from `GET /command`.
+@Freezed(fromJson: true, toJson: true)
+sealed class CommandInfo with _$CommandInfo {
+  const factory CommandInfo({
+    required String name,
+    String? template,
+    List<String>? hints,
+    String? description,
+    String? agent,
+    String? model,
```
AlexDochioiu: use `AgentModel` class. Also I think this is missing the provider currently.

shared/sesori_shared/lib/src/models/sesori/send_command_request.dart
```diff
@@ -0,0 +1,15 @@
+import "package:freezed_annotation/freezed_annotation.dart";
+
+part "send_command_request.freezed.dart";
+
+part "send_command_request.g.dart";
+
+@Freezed(fromJson: true, toJson: true)
+sealed class SendCommandRequest with _$SendCommandRequest {
```
AlexDochioiu: prob not needed anymore

daniil-shumko:

bridge/sesori_plugin_opencode/lib/src/opencode_repository.dart
```diff
@@ -48,6 +58,57 @@ class OpenCodeRepository {
 
   OpenCodeApi get api => _api;
 
+  Future<List<PluginCommand>> getCommands({required String? projectId}) async {
+    final commands = await _api.listCommands(directory: _normalizeDirectory(projectId));
+    return commands.map<PluginCommand>(_mapCommand).toList();
+  }
+
+  Future<PluginSession> createSession({
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 169f1e9. `OpenCodeRepository.createSession()` now trims the incoming directory before calling the API and uses the normalized value for the returned `PluginSession.projectID` as well. Added coverage in `bridge/sesori_plugin_opencode/test/opencode_repository_test.dart`.

mobile/module_core/lib/src/services/slash_command_service.dart
```diff
@@ -0,0 +1,73 @@
...
+    required String? agent,
+    required String? providerID,
+    required String? modelID,
+    required String? command,
+  }) {
+    return _repository.sendMessage(
+      sessionId: sessionId,
+      text: text,
+      agent: command != null ? null : agent,
+      providerID: command != null ? null : providerID,
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 169f1e9. `SlashCommandService.sendMessage()` now normalizes blank/whitespace `providerID` and `modelID` to `null`, and I also normalized blank commands consistently so they behave like no command at all. Added regression coverage in `mobile/module_core/test/services/slash_command_service_test.dart`.

mobile/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -661,33 +683,99 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
-      modelID: current is SessionDetailLoaded ? current.selectedModelID : null,
-      isConnected: current is SessionDetailLoaded && _isConnected,
+    final trimmed = text.trim();
+    if (trimmed.isEmpty && command == null) return;
+
+    final submission = QueuedSessionSubmission(text: trimmed, command: command);
+    if (current is! SessionDetailLoaded || !_isConnected) {
+      _promptQueue.enqueue(submission);
+      _emitQueueUpdate(current is SessionDetailLoaded ? current : null);
+      return;
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 169f1e9. `SessionDetailCubit.sendMessage()` now enqueues when work is already queued or a drain is in flight, so connected sends can no longer bypass earlier pending items. I also added a regression test in `mobile/app/test/features/session_detail/session_detail_cubit_test.dart` covering the in-flight drain case.

bridge/sesori_plugin_opencode/lib/src/opencode_service.dart
```diff
@@ -49,6 +60,62 @@ class OpenCodeService {
...
+
+    if (parts.isNotEmpty) {
+      await repository.sendPrompt(
+        sessionId: session.id,
+        directory: session.directory,
+        parts: parts,
+        agent: agent,
+        model: model,
+      );
+    }
```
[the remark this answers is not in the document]
AlexDochioiu: Addressed in 169f1e9, but with cleanup rather than silent success. If the first prompt send fails after session creation, `OpenCodeService.createSession()` now deletes the just-created backend session and rethrows, so we do not lose consistency or return a session whose initial prompt never landed. Added regression coverage in `bridge/sesori_plugin_opencode/test/opencode_service_test.dart`.

bridge/app/lib/src/bridge/repositories/command_repository.dart
```diff
@@ -0,0 +1,20 @@
...
+import "package:sesori_shared/sesori_shared.dart" show CommandListResponse;
+
+import "mappers/plugin_command_mapper.dart";
+
+class CommandRepository {
+  final BridgePlugin _plugin;
+
+  CommandRepository({required BridgePlugin plugin}) : _plugin = plugin;
+
+  Future<CommandListResponse> getCommands({required String? projectId}) async {
```
AlexDochioiu: Commands are session scoped, and SessionRepository already handles "sendCommand". I think we should not create this `CommandRepository` at all. It either handles everything to do with commands or nothing. Let's not split commands stuff into multiple repos.

bridge/app/lib/src/bridge/repositories/session_repository.dart
```diff
@@ -75,6 +75,35 @@ class SessionRepository {
     return enrichPluginSession(pluginSession: updated);
   }
 
+  Future<void> sendCommand({
+    required String sessionId,
+    required String command,
+    required String arguments,
+  }) {
+    return _plugin.sendCommand(
```
AlexDochioiu, line 97: I think this plugin function should also include agent and model, just like sendPrompt? and the plugin should use this information.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -66,6 +70,12 @@ abstract class BridgePlugin {
     required ({String providerID, String modelID})? model,
   });
 
+  Future<void> sendCommand({
+    required String sessionId,
+    required String command,
+    required String arguments,
```
AlexDochioiu, line 76: agent and model should prob be included.

bridge/app/lib/src/bridge/repositories/session_repository.dart
AlexDochioiu: Fixed in 4eb3194. `SessionRepository.sendCommand(...)` now takes `agent` and `model`, and both `SessionPromptService` and `SessionCreationService` pass those through on command-backed sends. I also updated the bridge-side tests to assert that command sends preserve the selected agent/model.

bridge/app/lib/src/bridge/repositories/command_repository.dart
AlexDochioiu: Fixed in 4eb3194. I removed `CommandRepository` and folded the command-list lookup into `SessionRepository`, so command-related plugin access now stays in one repository boundary instead of being split across two repos. `GetCommandsHandler` now depends on `SessionRepository`, and the bridge routing tests were updated accordingly.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
AlexDochioiu: Fixed in 4eb3194. `BridgePlugin.sendCommand(...)` now includes `agent` and `model`, and the OpenCode plugin threads those values all the way into `SendCommandBody`. I also followed through on the mobile side so command requests no longer strip the selected agent/model before they ever reach the bridge, and updated the plugin/mobile tests for the new end-to-end behavior.

mobile/module_core/lib/src/api/slash_command_api.dart
```diff
@@ -0,0 +1,63 @@
+import "package:injectable/injectable.dart";
+import "package:sesori_auth/sesori_auth.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "client/relay_http_client.dart";
+
+@lazySingleton
+class SlashCommandApi {
```
AlexDochioiu: This should be under SessionApi. The commands, etc are all session scoped. So is the sending of a message (with or without a command)

AlexDochioiu: If there is no SessionApi already existing, basically just rename this to SessionApi (and the file name, etc)

mobile/module_core/lib/src/capabilities/session/session_service.dart
AlexDochioiu: Based on what this does, I'd say this is a `SessionApi` not a `SessionService` as it only calls the API endpoints and does nothing else. Implementation is correct, it's just the name wrong. And this should absorb the SlashCommandApi

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
```diff
@@ -3,35 +3,42 @@ import "package:sesori_auth/sesori_auth.dart";
 import "package:sesori_shared/sesori_shared.dart";
 
 import "../../capabilities/session/session_service.dart";
+import "../../services/slash_command_service.dart";
 import "new_session_state.dart";
 
 class NewSessionCubit extends Cubit<NewSessionState> {
   final SessionService _sessionService;
```
AlexDochioiu, line 9: Since we rename this to `SessionApi`, it's technically an arch violation to inject it directly in NewSessionCubit. This PR is already quite big. We accept this arch violation for now to not grow this PR further.

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
```diff
@@ -128,46 +144,72 @@ class NewSessionCubit extends Cubit<NewSessionState> {
...
   // ignore: no_slop_linter/prefer_required_named_parameters, public cubit API consumed by UI layer
   void selectModel(String providerID, String modelID) {
     _emitAgentModelUpdate(selectedProviderID: providerID, selectedModelID: modelID);
   }
 
-  Future<void> createSessionWithMessage({
+  Future<void> createSession({
     required String text,
     required bool dedicatedWorktree,
+    String? command,
```
AlexDochioiu: `required String? command`. We should always use required even for nullable args. make sure this is mentioned in AGENTS.md as a strong requirement. It makes omissions by mistake much less common.

mobile/module_core/lib/src/repositories/slash_command_repository.dart
```diff
@@ -0,0 +1,52 @@
+import "package:injectable/injectable.dart";
+import "package:sesori_auth/sesori_auth.dart";
+import "package:sesori_shared/sesori_shared.dart";
+
+import "../api/slash_command_api.dart";
+
+@lazySingleton
+class SlashCommandRepository {
```
AlexDochioiu: should be `SessionRepository`

mobile/module_core/lib/src/api/slash_command_api.dart
AlexDochioiu: Fixed in 4961e30. I kept the API boundary but renamed it into the session layer as requested: `SlashCommandApi` -> `SessionApi`, file renamed to `session_api.dart`, and all usages/exports/DI wiring now point at the session naming.

mobile/module_core/lib/src/repositories/slash_command_repository.dart
AlexDochioiu: Fixed in 4961e30. The repository stays in place but is now renamed into the session layer as well: `SlashCommandRepository` -> `SessionRepository`, including imports, exports, and DI/codegen fallout. `SlashCommandService` still goes through that repository rather than calling the API directly.

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
AlexDochioiu: Fixed in 4961e30. I changed the new public mobile command-bearing APIs to use explicit nullable named parameters (`required String? command`) and updated the direct call sites/tests accordingly. I also added the rule to `mobile/module_core/AGENTS.md` so the convention is documented locally.

mobile/module_core/lib/src/services/slash_command_service.dart
```diff
@@ -2,13 +2,13 @@ import "package:injectable/injectable.dart";
 import "package:sesori_auth/sesori_auth.dart";
 import "package:sesori_shared/sesori_shared.dart";
 
-import "../repositories/slash_command_repository.dart";
+import "../repositories/session_repository.dart";
 
 @lazySingleton
 class SlashCommandService {
```
AlexDochioiu: SessionService?

mobile/module_core/lib/src/repositories/slash_command_repository.dart
AlexDochioiu: Fixed in b7dda23. I kept the repository but renamed it into the session layer (`SessionRepository`) and wired the command-related methods through it instead of letting the renamed `SessionApi` be used directly for those service-level flows.

mobile/module_core/lib/src/api/slash_command_api.dart
AlexDochioiu: Fixed in b7dda23. I merged the slash-command transport wrapper into the existing session transport layer under the honest name `SessionApi`, so the command endpoints now live alongside the other session-scoped HTTP calls instead of in a separate `SlashCommandApi`.

mobile/module_core/lib/src/capabilities/session/session_service.dart
AlexDochioiu: Fixed in b7dda23. The old transport wrapper that was named `SessionService` is now `SessionApi`, and the command-aware orchestration layer is now the real `SessionService`. I kept the repository in between so the layer split stays `SessionApi -> SessionRepository -> SessionService` for these flows.

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
AlexDochioiu: Left as-is intentionally. Even after the rename in b7dda23, `NewSessionCubit` still injects `SessionApi` directly for now, matching your note that this specific architecture violation can stay in this PR to avoid widening scope further.

mobile/module_core/lib/src/cubits/new_session/new_session_cubit.dart
AlexDochioiu: Fixed in b7dda23. The command-bearing public mobile APIs now use explicit nullable named parameters (`required String? command`), and I updated the direct call sites/tests plus the local `mobile/module_core/AGENTS.md` guidance to match that rule.

PR #83, opened 2026-04-02
feat(bridge): propagate session archive to OpenCode

AlexDochioiu:
## Summary

- Adds `archiveSession` to the `BridgePlugin` interface and implements it in `OpenCodePlugin` via `PATCH /session/{id}` with `{"time": {"archived": <now>}}`
- The archive handler fires the plugin notification as fire-and-forget (`unawaited` + `try/catch` + `catchError`) — the local DB is authoritative, so the request never blocks on or fails for a plugin error
- Unarchive is intentionally NOT propagated to OpenCode; the existing merge logic in `GetSessionsHandler` already overrides the plugin's stale `archived` timestamp with the DB's `null` value

## Test coverage

- **Handler**: archive fires plugin call, unarchive does not, plugin error doesn't fail the request, handler doesn't await plugin (never-completing future test)
- **Plugin**: verifies `PATCH /session/{id}` is sent and `time.archived` is set as an `int` on the server
- **Merge (pre-existing)**: plugin returns `archived: 300` + DB has `archivedAt: null` → result is `archived: null`

### DIFF
--- a/bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
@@ -179,15 +179,11 @@ class UpdateSessionArchiveStatusHandler extends BodyRequestHandler<UpdateSession
 
     // Fire-and-forget: notify the backend so it can reflect the archive state.
     // The local DB is authoritative — we don't block on or fail for this.
-    try {
-      unawaited(
-        _plugin.archiveSession(sessionId: sessionDto.sessionId).catchError((Object e) {
-          Log.w("[archive] failed to notify plugin for session ${sessionDto.sessionId}: $e");
-        }),
-      );
-    } on Object catch (e) {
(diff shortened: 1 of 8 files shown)


bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
```diff
@@ -176,6 +177,18 @@ class UpdateSessionArchiveStatusHandler extends BodyRequestHandler<UpdateSession
...
+    // The local DB is authoritative — we don't block on or fail for this.
+    try {
+      unawaited(
+        _plugin.archiveSession(sessionId: sessionDto.sessionId).catchError((Object e) {
+          Log.w("[archive] failed to notify plugin for session ${sessionDto.sessionId}: $e");
+        }),
+      );
+    } on Object catch (e) {
+      Log.w("[archive] failed to notify plugin for session ${sessionDto.sessionId}: $e");
+    }
```
AlexDochioiu: since it's unawaited, the try catch wrapper in pointless. The error would go to `catchError`

PR #87, opened 2026-04-04
fix: make session.error sessionID nullable to match OpenCode schema

AlexDochioiu:
## Summary

- **Fix**: `SseEventParser` crashes with `type 'Null' is not a subtype of type 'String' in type cast` when receiving `session.error` SSE events from OpenCode
- **Root cause**: OpenCode defines `sessionID` as optional on `session.error` events ([source](https://github.com/sst/opencode/blob/main/packages/opencode/src/session/index.ts#L232)), but our models declared it as `required String sessionID`
- **Change**: Made `sessionID` nullable (`String?`) across all three layers — bridge plugin model (`SseEventData.sessionError`), plugin interface (`BridgeSseSessionError`), and shared model (`SesoriSseEvent.sessionError`) — plus regenerated Freezed/JSON code

### DIFF
--- a/bridge/sesori_plugin_interface/lib/src/bridge_sse_event.dart
+++ b/bridge/sesori_plugin_interface/lib/src/bridge_sse_event.dart
@@ -42,7 +42,7 @@ class BridgeSseSessionDiff extends BridgeSseEvent {
 }
 
 class BridgeSseSessionError extends BridgeSseEvent {
-  final String sessionID;
+  final String? sessionID;
   const BridgeSseSessionError({required this.sessionID});
 }
 
--- a/bridge/sesori_plugin_opencode/lib/src/models/sse_event_data.dart
+++ b/bridge/sesori_plugin_opencode/lib/src/models/sse_event_data.dart
@@ -76,7 +76,7 @@ sealed class SseEventData with _$SseEventData {
   @FreezedUnionValue("session.error")
   @Implements<SseSessionEventData>()
   const factory SseEventData.sessionError({
-    required String sessionID,
+    required String? sessionID,
   }) = SseSessionError;
 
   @FreezedUnionValue("session.compacted")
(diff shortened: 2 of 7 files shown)


bridge/sesori_plugin_interface/lib/src/bridge_sse_event.dart
```diff
@@ -42,8 +42,8 @@ class BridgeSseSessionDiff extends BridgeSseEvent {
 }
 
 class BridgeSseSessionError extends BridgeSseEvent {
-  final String sessionID;
-  const BridgeSseSessionError({required this.sessionID});
+  final String? sessionID;
+  const BridgeSseSessionError({this.sessionID});
```
AlexDochioiu: keep `required` even if nullable

bridge/sesori_plugin_opencode/lib/src/models/sse_event_data.dart
```diff
@@ -76,7 +76,7 @@ sealed class SseEventData with _$SseEventData {
   @FreezedUnionValue("session.error")
   @Implements<SseSessionEventData>()
   const factory SseEventData.sessionError({
-    required String sessionID,
+    String? sessionID,
```
AlexDochioiu: keep `required` even if nullable . update agents.md to mention that we should ALWAYS use required, even if nullable

shared/sesori_shared/lib/src/models/sesori/sesori_sse_event.dart
```diff
@@ -75,7 +75,7 @@ sealed class SesoriSseEvent with _$SesoriSseEvent {
   @FreezedUnionValue("session.error")
   @Implements<SesoriSessionEvent>()
   const factory SesoriSseEvent.sessionError({
-    required String sessionID,
+    String? sessionID,
```
AlexDochioiu: keep `required` even if nullable

PR #91, opened 2026-04-06
fix(bridge): pass model status and releaseDate through provider pipeline

AlexDochioiu:
## Summary

- Adds `status` and `releaseDate` fields to `PluginModel` in the plugin interface, threading them from the OpenCode API through the bridge to the mobile app
- Refactors provider mapping into proper repository layers to satisfy architectural layering rules

## Problem

The mobile model picker's "pick newest per family" algorithm was receiving `null` for all model release dates, making selection effectively random within each family. This caused:
- Spark/variant models shown instead of regular versions (e.g., Codex 5.4 Spark instead of Codex 5.4)
- Older model versions shown when newer exist (e.g., Codex 5.3 shown alongside 5.4)
- Old Haiku variants shown instead of the latest

## Root Cause

`PluginModel` only carried `id`, `name`, `family` — missing `status` and `releaseDate`. The OpenCode API provides both, but they were dropped during the plugin mapping. The bridge handler hardcoded `releaseDate: null` for all models.

## Changes

| File | Change |
|------|--------|
| `sesori_plugin_interface/.../plugin_provider.dart` | Added `status` (default "active") and `releaseDate` to `PluginModel` |
| `sesori_plugin_opencode/.../provider_mapper.dart` | **New** — extracted provider mapping (OpenCode → PluginModel/PluginProvider) to repository layer |
| `sesori_plugin_opencode/.../opencode_repository.dart` | Added `getProviders()` delegating to mapper |
| `sesori_plugin_opencode/.../opencode_service.dart` | Updated to delegate to repository |
| `sesori_plugin_opencode/.../opencode_plugin_impl.dart` | Simplified to delegate through service (removed ~100 lines of inline mapping) |
| `bridge/app/.../provider_repository.dart` | **New** — wraps plugin.getProviders() and maps to shared models |
| `bridge/app/.../plugin_provider_mapper.dart` | **New** — PluginProvider → ProviderInfo/ProviderModel mapping |
| `bridge/app/.../get_providers_handler.dart` | Simplified to use ProviderRepository (removed inline mapping) |

## Verification

- `make analyze` — 0 issues across all 3 bridge modules
- `make test` — 101 plugin tests + 593 app tests pass


bridge/sesori_plugin_interface/lib/src/models/plugin_provider.dart
```diff
@@ -18,6 +18,8 @@ sealed class PluginModel with _$PluginModel {
     required String id,
     required String name,
     String? family,
+    @Default("active") String status,
```
AlexDochioiu: just change to a boolean `isAvailable`. I don't think we care about other statuses.

bridge/sesori_plugin_interface/lib/src/models/plugin_provider.dart
```diff
@@ -18,6 +18,8 @@ sealed class PluginModel with _$PluginModel {
...
+    String? releaseDate,
```
AlexDochioiu: `DateTime?` please. If you don't have one, create a json converter for date time

bridge/app/lib/src/bridge/repositories/mappers/plugin_provider_mapper.dart
```diff
@@ -16,8 +16,8 @@ extension PluginProviderMapper on PluginProvider {
             providerID: id,
             name: m.name,
             family: m.family,
-            status: m.status,
-            releaseDate: m.releaseDate,
+            status: m.isAvailable ? "active" : "deprecated",
```
AlexDochioiu: update shared type to also have boolean isAvailable instead

bridge/app/lib/src/bridge/repositories/mappers/plugin_provider_mapper.dart
```diff
@@ -16,8 +16,8 @@ extension PluginProviderMapper on PluginProvider {
...
+            releaseDate: m.releaseDate?.toIso8601String().split("T").first,
```
AlexDochioiu: keep `DateTime?` type in shared model

bridge/sesori_plugin_interface/lib/src/models/plugin_provider.dart
```diff
@@ -12,17 +12,27 @@ enum PluginProviderAuthType {
...
 
+class _NullableDateTimeConverter implements JsonConverter<DateTime?, String?> {
+  const _NullableDateTimeConverter();
+
+  @override
+  DateTime? fromJson(String? json) => json == null ? null : DateTime.tryParse(json);
+
+  @override
+  String? toJson(DateTime? object) => object?.toIso8601String().split("T").first;
+}
```
AlexDochioiu: Better create one public in shared dart repo that everyone can use.

p.s. if you don't ever use fromJson or toJson on this `PluginModel` you don't need this here. 

P.s. should be called DateConverter if it strips out the time or DateTimeConverter if it leaves it untouched. 

do not make it nullable. Should be `JsonConverter<DateTime, String>`. The freezed gen code already knows to not call it for null values.

bridge/sesori_plugin_opencode/lib/src/provider_mapper.dart
```diff
@@ -15,7 +15,13 @@ PluginProvidersResult mapProviderResponse({
   final providers = source.map((providerInfo) {
     final models = providerInfo.models.values
         .map(
-          (m) => PluginModel(id: m.id, name: m.name, family: m.family, status: m.status, releaseDate: m.releaseDate),
+          (m) => PluginModel(
+            id: m.id,
+            name: m.name,
+            family: m.family,
+            isAvailable: m.status != "deprecated",
```
AlexDochioiu: hmm, this can silently break if underlying API changes values. do a switch statement on know values and for unknown string values log warning about unknown status.

bridge/sesori_plugin_opencode/lib/src/provider_mapper.dart
```diff
@@ -15,7 +15,13 @@ PluginProvidersResult mapProviderResponse({
...
+            releaseDate: m.releaseDate != null ? DateTime.tryParse(m.releaseDate!) : null,
```
AlexDochioiu: do a switch or similar to avoid bang operator

PR #93, opened 2026-04-06
feat: session worktree flagging + shared-worktree protection

AlexDochioiu:
## Summary

- Add `hasWorktree` boolean to the shared `Session` model so the mobile app knows whether a session has a dedicated worktree
- Skip the worktree/branch cleanup bottom sheet for sessions without worktrees — delete/archive directly
- Reject worktree/branch cleanup when another active session shares the same worktree or branch (non-forceable, even with `force: true`)
- Mobile shows an informational-only dialog (no force button) for shared-worktree rejections

### Shared (`sesori_shared`)
- `Session` gains `@Default(false) bool hasWorktree` (backward-compatible JSON)
- `CleanupIssue.sharedWorktree()` variant added to the sealed class

### Bridge
- `SessionDao.getOtherActiveSessionsSharing()` — finds other active sessions sharing worktree/branch (OR logic, excludes self + archived)
- `SessionRepository.hasOtherActiveSessionsSharing()` — Layer 2 wrapper for the DAO query
- `performWorktreeCleanup` — shared-worktree check before the `!force` guard (non-bypassable)
- `SessionRepository.getSessionsForProject()` — merges `hasWorktree` from DB
- `CreateSessionHandler` / `UpdateSessionArchiveStatusHandler` — populate `hasWorktree` on response sessions

### Mobile
- `_showDeleteSheet` / `_showArchiveSheet` — skip bottom sheet when `hasWorktree == false`
- `_showForceDialog` — conditionally hides force button for non-forceable issues
- `_describeCleanupIssue` — handles `CleanupIssueSharedWorktree`
- Extracted `session_force_dialog.dart` from `session_cleanup_dialogs.dart` (file size compliance)

## Testing
- 187 shared tests pass
- 615 bridge tests pass (24 new tests)
- 393 mobile flutter tests pass
- `dart analyze` clean across all workspaces


bridge/app/lib/src/bridge/persistence/daos/session_dao.dart
```diff
@@ -63,6 +63,30 @@ class SessionDao extends DatabaseAccessor<AppDatabase> with _$SessionDaoMixin {
...
+          Expression<bool>? sharingCondition;
+          if (worktreePath != null) {
+            sharingCondition = t.worktreePath.equals(worktreePath);
+          }
+          if (branchName != null) {
+            final branchCondition = t.branchName.equals(branchName);
+            sharingCondition = sharingCondition != null ? (sharingCondition | branchCondition) : branchCondition;
+          }
+
+          return condition & sharingCondition!;
```
AlexDochioiu: No bang operators allowed. Write this code in a way where we can guarantee at compile-time that this is not null and allow smartcasting/etc

bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
```diff
@@ -188,7 +193,7 @@ class UpdateSessionArchiveStatusHandler extends BodyRequestHandler<UpdateSession
     final responseSession = _withArchivedTime(
       session: pluginSession.toSharedSession(),
       archivedAt: archivedAt,
-    );
+    ).copyWith(hasWorktree: sessionDto.worktreePath != null);
```
AlexDochioiu: This usage of `copyWith` seems hacky. I don't like it. find a proper way to set `hasWorktree`

bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
```diff
@@ -227,7 +232,7 @@ class UpdateSessionArchiveStatusHandler extends BodyRequestHandler<UpdateSession
     final responseSession = _withArchivedTime(
       session: pluginSession.toSharedSession(),
       archivedAt: null,
-    );
+    ).copyWith(hasWorktree: sessionDto.worktreePath != null);
```
AlexDochioiu: This usage of `copyWith` seems hacky. I don't like it. find a proper way to set `hasWorktree`

bridge/app/lib/src/bridge/routing/worktree_cleanup.dart
```diff
@@ -30,6 +32,20 @@ Future<CleanupResult> performWorktreeCleanup({
   required bool deleteBranch,
   required bool force,
 }) async {
+  // Shared-worktree check — non-bypassable, even with force=true.
```
AlexDochioiu: we should be able to bypass it, otherwise it's a "stalemate" where no session can possibly clean up the worktree. 
tl;dr "force" flag should allow deletion.

mobile/app/lib/features/session_list/session_list_actions.dart
```diff
@@ -9,6 +9,17 @@ void _showArchiveSheet({
   required SessionListCubit cubit,
   required Session session,
 }) {
+  if (!session.hasWorktree) {
+    _archiveSession(
```
AlexDochioiu, line 15: leave a comment that non-worktree sessions should bypass the "delete worktree and branch" modal

mobile/app/lib/features/session_list/session_list_actions.dart
```diff
@@ -121,6 +132,17 @@ void _showDeleteSheet({
   required SessionListCubit cubit,
   required Session session,
 }) {
+  if (!session.hasWorktree) {
+    _deleteSession(
```
AlexDochioiu, line 140: leave a comment that non-worktree sessions should bypass the "delete worktree and branch" modal

PR #95, opened 2026-04-08
fix(bridge): persist plugin projects & sessions to satisfy FK constraints + v5 migration

AlexDochioiu:
## Summary

Fixes an edge-case bug where projects and sessions fetched from the plugin (OpenCode) were returned to mobile without being persisted to the local SQLite database, causing downstream \`FOREIGN KEY constraint failed\` errors when \`PrSyncService\` inserted a PR row referencing a project that only existed in plugin memory.

## The bug

\`pull_requests_table.projectId\` has a \`REFERENCES projects_table(project_id) ON DELETE CASCADE\` constraint and \`PRAGMA foreign_keys = ON\` is enabled at runtime (\`database.dart:87\`). When the mobile client fetched projects from the plugin and then triggered a PR sync, SQLite rejected the PR insert for any project that wasn't already persisted to the local DB.

## Fix — three layered defenses

1. **Forward-prevention** — \`ProjectRepository.getProjects()\` (new Layer 2 class) fetches plugin projects, upserts each one via \`ProjectsDao.insertProjectIfMissing\` in a single transaction, applies the hidden filter, and sorts. \`GetProjectsHandler\` delegates to it. Every project returned to mobile is guaranteed to have a DB row.

2. **Defensive backstop** — \`PullRequestRepository.upsertFromGhPr\` and \`upsertPullRequest\` now call \`ProjectsDao.insertProjectIfMissing(projectId)\` at the start (Layer 2 → Layer 1, downward dep). Even if SSE triggers PR sync before mobile has called \`GET /projects\`, the parent row exists. This also covers \`PrSyncService\`'s two call sites automatically.

3. **Schema-level enforcement** — v5 Drift migration adds a proper FK constraint on \`session_table.projectId → projects_table.projectId\` with \`ON DELETE CASCADE\`. The migration seeds placeholder projects for any existing orphan sessions (no data loss) and validates via \`PRAGMA foreign_key_check\` post-migration. Uses Drift's \`TableMigration\` API for correct table recreation.

## New architectural pieces

- **\`ProjectRepository\`** (Layer 2, 40 lines, single public method \`getProjects()\`) — consumed by \`GetProjectsHandler\`
- **\`SessionPersistenceService\`** (Layer 3) — owns all cross-concern persistence with transaction wrapping. Public surface: \`ensureProject\`, \`persistSessionsForProject\`, \`createSession\`. Consumed by \`CreateSessionHandler\` and \`GetSessionsHandler\`. Matches the existing \`WorktreeService\` pattern (Layer 3 → direct Layer 1 deps).

## Other fixes rolled into this PR

- **T1 — mapper relocation**: \`PluginProjectMapper\` moved from \`routing/\` (Layer 4) to \`repositories/mappers/\` (Layer 2), matching the layered-architecture reference diagram
- **T4 — \`unhideProject\` correctness**: previously an UPDATE-only statement that silently no-oped for missing rows (would have bricked \`OpenProjectHandler\` post-v5). Rewritten to use Drift's \`DoUpdate\` with \`target\` so it creates the row if missing while preserving \`baseBranch\` and \`worktreeCounter\` on existing rows.
- **T8 — \`CreateSessionHandler\` first-time-project fix**: non-dedicated sessions for brand-new projects skipped \`WorktreeService.prepareWorktreeForSession\` (which also created the project row). Now calls \`SessionPersistenceService.createSession()\` which ensures the project exists AND inserts the session in a single transaction.
- **\`hiddenStore\` alias removal**: the ugly \`final hiddenStore = projectsDao;\` alias in \`request_router.dart\` is gone; the field on \`OpenProjectHandler\` is also renamed

## Test coverage

- \`make test\` → **651 tests passing**
- **New unit tests**: \`ProjectRepository\` (3), \`SessionPersistenceService\` (5), \`PullRequestRepository\` regression (1)
- **New DAO tests**: \`insertProjectIfMissing\` (3 preservation tests), \`insertSessionIfMissing\` (2), \`unhideProject\` (3 preservation tests), runtime FK enforcement (2)
- **New migration tests**: v4→v5 structural validation, data integrity, orphan cleanup, runtime FK enforcement, cascade delete (5 tests total)
- **New integration test**: \`pr_sync_fk_regression_test.dart\` — 3 scenarios reproducing the original bug and verifying all 3 defense paths work without v5 being strictly required

## Out-of-scope (explicitly preserved per Aristotle legacy clause)

- \`SessionRepository\` is **byte-for-byte unchanged from main**. Its \`getChildSessions\` and \`getProjectPath\` methods have a similar latent gap but are not addressed here.
- The dual \`bridge/persistence/\` vs \`bridge/api/database/\` directory structure is untouched (separate refactor).
- Other handlers still calling \`ProjectsDao\` directly (\`HideProjectHandler\`, \`OpenProjectHandler\`, \`GetBaseBranchHandler\`, \`SetBaseBranchHandler\`) are untouched.

## High-risk commit — v5 migration is irreversible

Commit \`fbed13c\` bumps the schema version from 4 to 5. Users who upgrade the bridge and run it once will have a v5 SQLite database. Rolling the schema back would require a separate v5→v4 migration (not provided). All forward-prevention (commits before \`fbed13c\`) fixes the bug on its own; v5 is belt-and-suspenders.

## Commit history (14 atomic commits)

\`\`\`
e36833e fix(bridge/services): add createSession to SessionPersistenceService; remove SessionDao dep from CreateSessionHandler
50fa90a chore(bridge): T13 validation fix-ups (codegen/analyze/test)
9dd5cce test(bridge/persistence): runtime FK enforcement for SessionDao post-v5
fbed13c feat(bridge/persistence): bump schemaVersion to 5 with FK on session_table.projectId
2c5b329 test(bridge/integration): regression test for PR sync FK constraint bug
2d5d8e3 fix(bridge/routing): CreateSessionHandler ensures project exists before insertSession
8510e22 fix(bridge/repositories): PullRequestRepository ensures project exists before upsert
5191917 refactor(bridge/routing): GetProjectsHandler uses ProjectRepository
084e766 feat(bridge/repositories): introduce ProjectRepository
4a8a9fc feat(bridge/services): introduce SessionPersistenceService + wire into GetSessionsHandler
e21ba96 fix(bridge/persistence): unhideProject creates row if missing
96247ce chore(bridge): move PluginProjectMapper to repositories/mappers/
b3b4ae6 feat(bridge/persistence): add SessionDao.insertSessionIfMissing
24af4cc feat(bridge/persistence): add ProjectsDao.insertProjectIfMissing
\`\`\`

Planned via Prometheus, Metis, and Momus; architecturally reviewed via the Aristotle specification. Work plan at \`.sisyphus/plans/plugin-project-session-db-sync.md\`.


bridge/app/lib/src/bridge/persistence/daos/projects_dao.dart
```diff
@@ -74,4 +80,14 @@ class ProjectsDao extends DatabaseAccessor<AppDatabase> with _$ProjectsDaoMixin
...
+
+  /// Inserts a minimal project row if none exists for [projectId].
+  /// Preserves all fields of existing rows — uses InsertMode.insertOrIgnore.
+  /// Use this to satisfy FK constraints without clobbering user-set state.
+  Future<void> insertProjectIfMissing({required String projectId}) async {
+    await into(projectsTable).insert(
+      ProjectsTableCompanion.insert(projectId: projectId),
+      mode: InsertMode.insertOrIgnore,
+    );
+  }
```
AlexDochioiu: is there a function/etc in drift for insert many with insertOrIgnore? May be more useful to send an array of project ids

bridge/app/lib/src/bridge/persistence/daos/session_dao.dart
```diff
@@ -86,6 +86,30 @@ class SessionDao extends DatabaseAccessor<AppDatabase> with _$SessionDaoMixin {
...
+  }) async {
+    await into(sessionTable).insert(
+      SessionTableCompanion(
+        sessionId: Value(sessionId),
+        projectId: Value(projectId),
+        // isDedicated hardcoded false — placeholders are non-dedicated by default.
+        // Callers (plugin-sourced sessions) never have meaningful worktree state.
+        isDedicated: const Value(false),
+        createdAt: Value(createdAt),
+        // worktreePath, branchName, archivedAt, baseBranch, baseCommit intentionally
```
AlexDochioiu: `archivedAt` would be good to get from opencode result so that we don't un-archive stuff as we put it in our db

AlexDochioiu:
## Addressing PR Review Comments

Pushed 3 new commits addressing 4 of the 6 review comments. 659 tests pass.

### ✅ Fixed

| Comment | Fix Commit | Details |
|---|---|---|
| **Codex P1** — `UpdateSessionArchiveStatusHandler._getSessionDto()` FK violation for first-time projects | `c558df3` | Handler now takes a `SessionPersistenceService` dep and calls `ensureProject(projectId:)` before the legacy `insertSession`. Added regression test. |
| **@AlexDochioiu** — Use Drift batch insert for bulk insertOrIgnore | `2e27951` | Added `ProjectsDao.insertProjectsIfMissing(List<String>)` and `SessionDao.insertSessionsIfMissing(List<Record>)` using Drift's `batch((b) => b.insertAll(...))`. `ProjectRepository.getProjects()` now uses the batch method directly (dropped per-item loop + `AppDatabase` dep). |
| **@AlexDochioiu** — Preserve `archivedAt` from plugin sessions | `2e27951` | `insertSessionIfMissing` now takes `required int? archivedAt`. `SessionPersistenceService.persistSessionsForProject()` passes `session.time?.archived` so archived state from the plugin is preserved in placeholder rows. |
| **Gemini** — Use `on Object` instead of `on Exception` | `66aa727` | Changed the best-effort try/catch in `GetSessionsHandler` to catch `Object` so `Error` subclasses (TypeError etc.) are also caught and logged. |

### ❌ Not Applied (with rationale)

| Comment | Reason |
|---|---|
| **Gemini** — Parallelize `insertProjectIfMissing` with `Future.wait` (in `ProjectRepository`) | SQLite transactions are single-writer — `Future.wait` inside a Drift transaction doesn't actually parallelize at the SQLite level (Drift serializes them). Instead, the proper idiom is Drift's `batch` API (applied above via `insertProjectsIfMissing`), which issues a single batched SQL statement. |
| **Gemini** — Parallelize `insertSessionIfMissing` with `Future.wait` (in `SessionPersistenceService`) | Same reason. Replaced with bulk `insertSessionsIfMissing` using Drift batch API. |

### ⚠️ Lint suppression (bot comment)
The `schema_v5.dart:3: // ignore_for_file: type=lint,unused_import` suppression flagged by the lint bot is in a **Drift-generated file** (`test/drift/default/generated/schema_v5.dart`). This suppression is auto-generated by `drift_dev make-migrations` and matches the pattern in all other schema files (v1-v4). It is not hand-written.

bridge/app/lib/src/bridge/persistence/daos/session_dao.dart
```diff
@@ -86,6 +86,59 @@ class SessionDao extends DatabaseAccessor<AppDatabase> with _$SessionDaoMixin {
...
+        // worktreePath, branchName, baseBranch, baseCommit intentionally
+        // omitted — they default to absent (null) via SessionTableCompanion
+      ),
+      mode: InsertMode.insertOrIgnore,
+    );
+  }
+
+  /// Bulk version of [insertSessionIfMissing]. Uses Drift's `batch` API.
+  /// Each entry provides (sessionId, projectId, createdAt, archivedAt).
+  Future<void> insertSessionsIfMissing({
```
AlexDochioiu: is there a point to have non-batched function too? Maybe just one single batched one is ok? Should work fine with single value too.

bridge/app/lib/src/bridge/persistence/daos/projects_dao.dart
```diff
@@ -74,4 +80,28 @@ class ProjectsDao extends DatabaseAccessor<AppDatabase> with _$ProjectsDaoMixin
...
+
+  /// Bulk version of [insertProjectIfMissing]. Uses Drift's `batch` API
+  /// with `insertAll` and [InsertMode.insertOrIgnore] for a single SQL
+  /// statement that preserves existing rows' fields.
+  Future<void> insertProjectsIfMissing({required List<String> projectIds}) async {
+    if (projectIds.isEmpty) return;
```
AlexDochioiu: is there a point to have non-batched function too? Maybe just one single batched one is ok? Should work fine with single value too.

bridge/app/lib/src/bridge/routing/request_router.dart
```diff
@@ -77,10 +77,17 @@ class RequestRouter {
...
+    final projectRepository = ProjectRepository(
+      plugin: plugin,
+      projectsDao: projectsDao,
+    );
     final permissionRepository = PermissionRepository(plugin: plugin);
+    final sessionPersistenceService = SessionPersistenceService(
+      projectsDao: projectsDao,
+      sessionDao: sessionDao,
+      db: projectsDao.attachedDatabase,
+    );
```
AlexDochioiu: those 3 should be created in orchestrator and injected in constructor

PR #101, opened 2026-04-09
feat: enable auto_vacuum on OpenCode database at bridge startup

AlexDochioiu:
## Summary

- On bridge startup, opportunistically enables `auto_vacuum = FULL` on OpenCode's SQLite database and runs `VACUUM`
- One-time conversion — once the DB header is updated, all future `DELETE` operations automatically reclaim disk space
- Non-destructive: if the DB is locked by OpenCode (`SQLITE_BUSY`), silently skips and retries next startup

## Problem

OpenCode stores every tool output in full as JSON in its SQLite `part` table. This causes unbounded database growth (500MB for 5 sessions, up to 25GB observed). OpenCode never runs `VACUUM` and never enables `auto_vacuum`. After session deletion, cascade deletes work correctly but freed pages are never reclaimed. Performance degrades severely — 30s per query at 25GB due to cache pollution, index fragmentation, and bloated WAL.

## Solution

On bridge startup, before plugin creation:
1. Find OpenCode's DB at `$XDG_DATA_HOME/opencode/opencode.db`
2. Check `PRAGMA auto_vacuum` — if already non-zero, skip
3. Set `PRAGMA auto_vacuum = FULL` and run `VACUUM`
4. If `SQLITE_BUSY` (OpenCode has the DB open), skip and retry next startup
5. Log database size, estimated duration, and space reclaimed

## Architecture

```
L1 OpenCodeDbApi → L2 OpenCodeDbRepository → L3 DbMaintenanceService
```

- **OpenCodeDbApi** (L1): Raw sqlite3 operations — open/close DB, read/set PRAGMAs, execute VACUUM
- **OpenCodeDbRepository** (L2): Wraps API, translates SqliteExceptions into nullable returns, handles file existence checks
- **DbMaintenanceService** (L3): Decision logic — should we optimize? Log with estimates, orchestrate the flow

## Files Changed

| File | Lines | Purpose |
|---|---|---|
| `bridge/app/lib/src/bridge/api/opencode_db_api.dart` | 49 | L1 — raw sqlite3 ops |
| `bridge/app/lib/src/bridge/repositories/opencode_db_repository.dart` | 69 | L2 — error translation, file checks |
| `bridge/app/lib/src/bridge/services/db_maintenance_service.dart` | 62 | L3 — decision logic, logging |
| `bridge/app/bin/bridge.dart` | +12 | Startup wiring |
| `bridge/app/test/bridge/api/opencode_db_api_test.dart` | 116 | API layer tests |
| `bridge/app/test/bridge/services/db_maintenance_service_test.dart` | 104 | Service layer tests |

## Verification

- `make analyze` — clean (1 pre-existing info-level issue)
- `make test` — 638 tests pass

### DIFF
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -11,21 +11,25 @@ import 'package:sesori_bridge/src/auth/token_manager.dart';
 import 'package:sesori_bridge/src/auth/validate.dart';
 import 'package:sesori_bridge/src/bridge/api/gh_cli_api.dart';
 import 'package:sesori_bridge/src/bridge/api/git_cli_api.dart';
+import 'package:sesori_bridge/src/bridge/api/opencode_db_api.dart';
 import 'package:sesori_bridge/src/bridge/bandwidth_tracker.dart';
 import 'package:sesori_bridge/src/bridge/debug_server.dart';
 import 'package:sesori_bridge/src/bridge/foundation/process_runner.dart';
 import 'package:sesori_bridge/src/bridge/log_failure_reporter.dart';
 import 'package:sesori_bridge/src/bridge/metadata_service.dart';
+
 import 'package:sesori_bridge/src/bridge/models/bridge_config.dart';
 import 'package:sesori_bridge/src/bridge/orchestrator.dart';
 import 'package:sesori_bridge/src/bridge/persistence/bridge_diagnostics.dart';
 import 'package:sesori_bridge/src/bridge/persistence/database.dart';
 import 'package:sesori_bridge/src/bridge/relay_client.dart';
+import 'package:sesori_bridge/src/bridge/repositories/opencode_db_repository.dart';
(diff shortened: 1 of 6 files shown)


bridge/app/lib/src/bridge/services/db_maintenance_service.dart
```diff
@@ -0,0 +1,63 @@
...
+import "../repositories/opencode_db_repository.dart";
+
+/// Opportunistically enables `auto_vacuum = FULL` on the OpenCode SQLite
+/// database during bridge startup.
+///
+/// This is a one-time conversion. Once the database header is updated, all
+/// future DELETE operations by OpenCode automatically reclaim disk space.
+/// If the database is in use, the operation is silently skipped and retried
+/// on the next bridge startup.
+class DbMaintenanceService {
```
AlexDochioiu: OpencodeDbMain... to be clear it's for opencode

PR #102, opened 2026-04-09
fix(push): use worktree path as projectID in SSE event and session mappings

AlexDochioiu:
## Summary

- Fix push notification deep linking — tapping a notification now navigates to the correct session detail screen instead of always landing on the projects list
- Override `projectID` with `directory` (worktree path) in `SseEventMapper` for all session events (created, updated, deleted)
- Align `SessionToPluginExtension.toPlugin()` to use `directory` as `projectID`, consistent with `Project.toPlugin()` which already uses `id: worktree`

## Root Cause

The mobile app uses worktree paths (e.g., `/Users/alex/my-project`) as project identifiers — set by `Project.toPlugin()` with `id: worktree`. However, the push notification payload carried `Session.projectID`, which is OpenCode's **internal** project ID (e.g., `"p1"`), not the worktree path. This mismatch caused the mobile app's `_pushSessionRoute()` to either:
- Fall back to the projects list (when `projectId` was null due to tracker timing)
- Navigate with the wrong ID (when the internal ID didn't match any known project)

## Changes

| File | Change |
|------|--------|
| `bridge/sesori_plugin_opencode/lib/src/sse_event_mapper.dart` | Override `projectID` with `info.directory` for `SseSessionCreated`, `SseSessionUpdated`, `SseSessionDeleted` |
| `bridge/sesori_plugin_opencode/lib/src/models/session.dart` | Change `toPlugin()` to use `projectID: directory` instead of `projectID: projectID` |
| `bridge/sesori_plugin_opencode/test/sse_event_mapper_test.dart` | New test verifying projectID override for all 3 session event types |
| `bridge/sesori_plugin_opencode/test/opencode_plugin_impl_test.dart` | Update assertion to expect directory-based projectID |

## Why `directory` is correct

- `Session.directory` IS the worktree path — the `ActiveSessionTracker` already uses it (`_sessionWorktrees[sessionId] = info.directory`)
- `Project.toPlugin()` explicitly uses `id: worktree` with comment "worktree is the most reliable UID with opencode"
- `buildSummary()` groups by worktree for `ProjectActivitySummary.id`
- The push tracker's `_applyProjectsSummaryChildLinks` already used worktree paths — but `_upsertSession` was overwriting with the internal ID. After this fix, both sources are consistent.

## Verification

- `make test` — all 625+ tests pass
- `make analyze` — no errors
- New test covers all 3 session event types with concrete values proving the override


bridge/sesori_plugin_opencode/lib/src/sse_event_mapper.dart
```diff
@@ -16,9 +16,9 @@ class SseEventMapper {
       SseServerHeartbeat() => const BridgeSseServerHeartbeat(),
       SseServerInstanceDisposed(:final directory) => BridgeSseServerInstanceDisposed(directory: directory),
       SseGlobalDisposed() => const BridgeSseGlobalDisposed(),
-      SseSessionCreated(:final info) => BridgeSseSessionCreated(info: info.toJson()),
-      SseSessionUpdated(:final info) => BridgeSseSessionUpdated(info: info.toJson()),
-      SseSessionDeleted(:final info) => BridgeSseSessionDeleted(info: info.toJson()),
+      SseSessionCreated(:final info) => BridgeSseSessionCreated(info: info.toJson()..['projectID'] = info.directory),
+      SseSessionUpdated(:final info) => BridgeSseSessionUpdated(info: info.toJson()..['projectID'] = info.directory),
+      SseSessionDeleted(:final info) => BridgeSseSessionDeleted(info: info.toJson()..['projectID'] = info.directory),
```
AlexDochioiu: `info` being a Map goes against our arch patterns and acceptable coding styles. Create a new freezed bridge type for it and keep full types knowledge.

PR #105, opened 2026-04-11
fix: harden OpenCode 1.4 compatibility handling

AlexDochioiu:
## Summary
- Hardened OpenCode SSE parsing for 1.4 compatibility: categorized malformed frames, preserved event type context, and kept `session.diff` invalidation-only behavior compatible with legacy payloads.
- Improved bridge/plugin routing so upstream decode drift and `PluginApiException` responses surface cleanly instead of degrading to generic failures.
- Added regression coverage for parser, plugin, bridge routing, and model-mapping compatibility cases.

## Risks addressed
- Dropped SSE frames are now logged with stable categories plus directory/event type context for diagnostics.
- Message decoding now distinguishes schema drift from unexpected bugs so the bridge can return the right status code.
- Route mappings continue to preserve provider/model/agent compatibility fields.

## Verification
- `GIT_MASTER=1 git status` → clean
- `lsp_diagnostics` on `bridge/app` and `bridge/sesori_plugin_opencode` → 0 errors
- `dart test test/bridge/routing/get_agents_handler_test.dart test/bridge/routing/get_providers_handler_test.dart test/bridge/routing/get_session_messages_handler_test.dart test/bridge/routing/request_router_test.dart`
- `dart test test/sse_event_parser_test.dart test/opencode_plugin_impl_test.dart test/opencode_service_test.dart test/agent_info_test.dart test/message_with_parts_test.dart test/provider_mapper_test.dart`

### DIFF
--- a/bridge/sesori_plugin_opencode/README.md
+++ b/bridge/sesori_plugin_opencode/README.md
@@ -18,7 +18,7 @@ OpenCodeApi             HTTP client — raw requests to the OpenCode REST API
 
 `SseConnection` runs alongside this stack, maintaining a persistent SSE connection to `GET /global/event` and feeding raw event strings to `OpenCodePlugin`. `SseEventParser` translates those strings into typed `SseEventData` objects. `ActiveSessionTracker` watches session status events to maintain a live count of busy sessions per project.
 
-Compatibility note: the plugin keeps bridge-facing SSE behavior intentionally narrow. `session.diff` is normalized into the shared event model for invalidation tracking, parser failures are reported as categorized outcomes instead of exceptions, and dropped SSE frames are logged with stable category tags plus `directory` context when available. Cold-start hydration of pending questions/permissions is best-effort only. Shell routes such as `GET /session/{id}/shell` remain outside the bridge router and are expected to 404.
+Compatibility note: the plugin keeps bridge-facing SSE behavior intentionally narrow. Parser failures are reported as categorized outcomes instead of exceptions, dropped SSE frames are logged with stable category tags plus `directory` context when available, and cold-start hydration of pending questions/permissions is best-effort only. Shell routes such as `GET /session/{id}/shell` remain outside the bridge router and are expected to 404.
 
 ## Key Components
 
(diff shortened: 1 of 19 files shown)


bridge/sesori_plugin_opencode/lib/src/sse_event_parser.dart
```diff
@@ -26,45 +45,170 @@ class SseParseResult {
...
+    }
+
+    return SseEventData.sessionDiff(
+          sessionID: sessionID,
+          diff: _extractLegacyDiffList(properties: properties),
+        )
+        as SseSessionDiff;
+  }
+
+  List<FileDiff> _extractLegacyDiffList({required Map<String, dynamic> properties}) {
```
AlexDochioiu: Ignore the legacy diff. just change the relevant `SseEventData`/etc to parse with freezed the new type. Sesori is not yet live so we don't care much about backwards compatibility.

bridge/sesori_plugin_opencode/README.md
```diff
@@ -18,6 +18,8 @@ OpenCodeApi             HTTP client — raw requests to the OpenCode REST API
 
 `SseConnection` runs alongside this stack, maintaining a persistent SSE connection to `GET /global/event` and feeding raw event strings to `OpenCodePlugin`. `SseEventParser` translates those strings into typed `SseEventData` objects. `ActiveSessionTracker` watches session status events to maintain a live count of busy sessions per project.
 
+Compatibility note: the plugin keeps bridge-facing SSE behavior intentionally narrow. `session.diff` is normalized into the shared event model for invalidation tracking, parser failures are reported as categorized outcomes instead of exceptions, and dropped SSE frames are logged with stable category tags plus `directory` context when available. Cold-start hydration of pending questions/permissions is best-effort only. Shell routes such as `GET /session/{id}/shell` remain outside the bridge router and are expected to 404.
```
AlexDochioiu: I don't want any stuff about `session.diff` in the readme. This is an implementation detail. README is not about specific things like this.

PR #109, opened 2026-04-15
fix: enrich bridge payloads and refresh session list on return

AlexDochioiu:
## Summary
- Bridge now enriches shared session payloads through the repository layer and keeps SSE delivery ordered in the orchestrator/debug server path.
- Mobile now performs a visible sessions-route refresh when returning to the session list, and the session list screen renders PR metadata.

## Verification
- `lsp_diagnostics` on touched bridge and mobile directories: 0 errors
- `dart test test/bridge/repositories/session_repository_test.dart test/bridge/routing/create_session_handler_test.dart test/bridge/routing/rename_session_handler_test.dart test/bridge/routing/update_session_archive_status_handler_test.dart test/bridge/sse/bridge_event_mapper_test.dart test/bridge/orchestrator_emit_bridge_event_test.dart test/bridge/debug_server_test.dart test/updater/release_contract_test.dart test/bridge/routing/get_sessions_handler_test.dart test/bridge/routing/worktree_cleanup_test.dart test/bridge/services/pr_sync_service_test.dart`
- `dart test test/cubits/session_list/session_list_cubit_test.dart`
- `flutter test test/features/session_list/session_lifecycle_ui_test.dart`

### Full mobile workspace verification
- `cd mobile/module_core && dart analyze` — passed
- `cd mobile/module_auth && dart analyze` — passed
- `cd mobile/app && dart analyze` — passed
- `cd mobile/module_core && dart test` — passed
- `cd mobile/module_auth && dart test` — passed
- `cd mobile/app && flutter test` — passed
- Full mobile workspace tests/analyzes passed.

---
## Summary by cubic
Enriches shared session payloads (stored timestamps/flags and PR metadata) across bridge APIs and SSE via repository-backed enrichment with async mapping that preserves SSE order and shields failures so streams don’t break. Centralizes session lifecycle into `SessionRepository` and new services; mobile refreshes the session list on return and shows PR metadata.

- **New Features**
  - Bridge: Async enrichment for `session.created`/`session.updated` via `SessionEventEnrichmentService` in the orchestrator and debug server; SSE order preserved with safe fallback. Archive restore resolves the base branch from current git state.
  - Mobile: Session list refreshes on route return and renders PR info.

- **Refactors**
  - Routed session flows through `SessionRepository`, `SessionCreationService`, and `SessionArchiveService`; handlers stop calling `BridgePlugin` directly. Moved prompt mappers under `repositories/mappers/`. Docs codify the layering rule.

<sup>Written for commit 68e89d9a215959f201f2c35ee9e9ecdd85c8b46c. Summary will update on new commits.</sup>


bridge/app/lib/src/bridge/routing/rename_session_handler.dart
```diff
@@ -26,6 +29,6 @@ class RenameSessionHandler extends BodyRequestHandler<RenameSessionRequest, Sess
       sessionId: body.sessionId,
       title: body.title,
     );
-    return updated.toSharedSession();
+    return _sessionRepository.enrichPluginSession(pluginSession: updated);
```
AlexDochioiu: This does not quite make sense to me. API and repository layers are generally meant to be data providers.

P.s. to me repository layer can transform/merge/etc data sources that it controls. Maybe the problem is that request handlers should not have access to plugin directly. In the end, this plugin is more or less an API abstraction, so I'd say it belongs to API arch layer.

So the flow should be API <- Repository <- ...

Think about this, ask me questions if you have any, and then update AGENTS.md based on the learnings

bridge/app/lib/src/bridge/repositories/session_repository.dart
```diff
@@ -33,36 +33,43 @@ class SessionRepository {
...
+  Future<Session> enrichPluginSession({required PluginSession pluginSession}) {
+    return enrichSession(session: pluginSession.toSharedSession());
+  }
+
+  Future<Session> enrichSessionJson({required Map<String, dynamic> sessionJson}) {
+    return enrichSession(session: Session.fromJson(sessionJson));
+  }
+
+  Future<List<Session>> enrichSessions({required List<Session> sessions}) async {
+    final sessionIds = sessions.map((session) => session.id).toList(growable: false);
```
AlexDochioiu, line 102: too many variations of almost the same thing?

AlexDochioiu:
@cubic review this

bridge/app/lib/src/bridge/routing/rename_session_handler.dart
AlexDochioiu: Addressed. I followed the direction here and moved BridgePlugin access below routing for these session lifecycle flows. The current split is: thin plugin-backed session commands/lookups in SessionRepository, multi-step create/archive orchestration in services, and handlers reduced to HTTP validation/translation only. I also updated both repo AGENTS.md files with the learned rule.

bridge/app/lib/src/bridge/routing/update_session_archive_status_handler.dart
```diff
@@ -99,20 +99,21 @@ class UpdateSessionArchiveStatusHandler extends BodyRequestHandler<UpdateSession
...
     final time = base.time;
     final mergedTime = time != null
         ? time.copyWith(archived: archivedAt)
         : SessionTime(created: 0, updated: 0, archived: archivedAt);
-    return base.copyWith(
+    final response = base.copyWith(
       time: mergedTime,
       hasWorktree: sessionDto.worktreePath != null,
     );
+    return _sessionRepository.enrichSession(session: response);
```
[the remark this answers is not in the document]
AlexDochioiu: Addressed. This thread was fixed earlier by removing the redundant archive-time/worktree merge from the handler and delegating the response build to repository enrichment instead. The current branch keeps that logic out of the handler.

bridge/app/lib/src/bridge/debug_server.dart
```diff
@@ -130,12 +137,11 @@ class DebugServer {
...
-        final mapped = _mapper.map(event);
-        if (mapped != null) {
-          unawaited(_fanOutEvent(jsonEncode(mapped.toJson())));
-        }
-      });
+      _pluginEventsSub ??= _plugin.events
+          .asyncMap<BridgeSseEvent>((event) => _sessionEventEnrichmentService.enrich(event))
+          .map<SesoriSseEvent?>((event) => _mapper.map(event))
+          .asyncMap((mapped) => _fanOutMappedEvent(mapped: mapped))
+          .listen((_) {});
```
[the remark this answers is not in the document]
AlexDochioiu: Addressed. I added explicit debug SSE stream error handling and also made session-event enrichment degrade to the original event with failure reporting instead of throwing through the stream. That keeps one bad enrichment from breaking debug SSE delivery.

AlexDochioiu:
@cubic-dev-ai I checked the issues identified by cubic and addressed the relevant bridge ones:

- Addressed: `plugin_session_mapper.dart` timestamp fallback now uses stored `createdAt` instead of fabricating `0` values.
- Addressed: `session_event_enrichment_service.dart` now catches enrichment failures, records them, and falls back to the original event instead of dropping delivery.
- Addressed: `orchestrator.dart` now records plugin event stream errors through the failure reporter, and the enrichment path no longer throws for handled enrichment failures.
- Addressed: `debug_server.dart` now has explicit SSE stream error handling, and the enrichment path degrades safely.
- Addressed: `session_archive_service.dart` no longer guesses `"main"`; unarchive restore now resolves the repo default branch from git state when the stored base branch is missing.
- Not addressed: the `mobile/module_core/test/cubits/session_list/session_list_cubit_test.dart` warning appears stale. In the current file, `mockRouteSource` is initialized in each test build path before `buildCubit()` is called, so there is no uninitialized use at the reported location.

I also pushed the follow-up commits with these fixes and re-ran the relevant bridge verification.

AlexDochioiu:
Follow-up note on the cubic review: after the first pass I also pushed `8bf9b7f` (`fix(bridge): shield SSE failure reporting fallback`) to close the last robustness gap Oracle found in the enrichment fallback path. The branch is now clean and the bridge verification remains green.

PR #110, opened 2026-04-16
fix(push): prune push session state and add maintenance telemetry

AlexDochioiu:
## Summary
- add push-session maintenance support so idle root subtrees can be pruned safely, bounded message-role and rate-limiter state can be retained/cleaned explicitly, and tracker internals stay split into focused files under the bridge file-size limit
- clear completion preview text at the right point in push notification delivery so completion payloads still send the final text while stale assistant previews do not linger after pruning or rate-limit decisions
- add the 10-minute push maintenance loop and debug telemetry snapshot/logging for retained sessions, assistant text, permission request bookkeeping, aborted/completed roots, and rate-limiter keys

## Verification
- `dart test test/push/push_session_state_tracker_test.dart`
- `dart test test/push/completion_notifier_test.dart`
- `dart test test/push/push_notification_service_test.dart`
- `dart test test/push/push_rate_limiter_test.dart`
- `make analyze && make test`

---
## Summary by cubic
Adds a 10‑minute push maintenance loop with telemetry and pruning to curb memory growth. Replaces the push service with a lean dispatcher, moves pipelines into listeners, tightens startup/abort flow, shares the bridge HTTP client for notifications, and updates docs to prefer composite subscriptions and clarify extraction rules.

- **New Features**
  - Added `MaintenancePushListener` (10‑min) to prune idle roots (>30m) and stale rate‑limiter keys (>30m), and emit telemetry (rss, sessions, idle/prunable roots, roles, assistant text, permissions, completions, aborts, rate‑limiter).
  - Added `CompletionPushListener` to subscribe on start, handle completion bookkeeping, build content with `PushNotificationContentBuilder`, and pass outbound completion data to `PushDispatcher`.
  - Replaced `PushNotificationService` with `PushDispatcher` that owns outbound sends (immediate/completion), rate limiting, payload building, and client disposal; trigger‑owned state now lives in listeners/tracker; composed inline in BridgeRuntime; Orchestrator wiring updated.
  - Introduced `SessionAbortService` to centralize aborts and expose an abortedSessions stream; router/handler use it and Orchestrator subscribes to update push state promptly.

- **Bug Fixes**
  - Hardened startup/abort flow: listeners are passive until explicitly started; suppress completion dispatch while an abort is pending; emit abort events only after a successful abort.
  - Ignore session summary replay when computing root idle age for pruning.
  - Clear latest assistant text when sending a completion (pre rate‑limit) to avoid lingering previews; pruned‑subtree cleanup cancels timers and clears retained links/permission mappings.
  - Continue pruning other roots even if one prune fails.
  - Use `CompositeSubscription` for push listener lifecycle to ensure clean teardown.
  - Share the bridge `http.Client` for push notifications; lifecycle is owned by the runtime.

<sup>Written for commit 4ff06065694187f5f2f4233fae1c545c963c7d67. Summary will update on new commits.</sup>


bridge/app/lib/src/push/push_session_state_tracker_mutations.dart
```diff
@@ -0,0 +1,129 @@
...
+  final sessionState = sessions.remove(sessionId);
+  for (final sessionState in sessions.values) {
+    sessionState.childIds.remove(sessionId);
+    if (sessionState.parentId == sessionId) {
+      sessionState.parentId = null;
+    }
+  }
+
+  permissionRequestToSession.removeWhere((_, value) => value == sessionId);
+  (sessionState?.messageIds ?? const <String>{}).forEach(messageRoles.remove);
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 4bee4e0. deleteTrackedSession now uses the removed session parent and child indexes for cleanup instead of scanning every session, and the shadowing local is gone. Verified with the push tracker tests and the full bridge analyze/test pass.

.github/workflows/bridge-release.yml
```diff
@@ -150,17 +162,37 @@ jobs:
 
   publish-platform:
     name: Publish platform npm packages
-    needs: release
+    needs: [build, release]
     runs-on: ubuntu-latest
-    if: github.event.inputs.dry_run != 'true' && startsWith(github.ref_name, 'bridge-v')
+    if: github.event.inputs.dry_run != 'true' && (startsWith(github.ref_name, 'bridge-v') || github.event.inputs.publish_npm == 'true')
```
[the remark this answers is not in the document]
AlexDochioiu: Not addressed in this PR after the latest branch direction. I merged current main and then reverted the branch-local workflow and docs divergence, so this PR no longer introduces workflow-specific changes relative to main. If this release-workflow issue still needs work, it should be handled in follow-up release-pipeline changes rather than in the push-memory branch.

.github/workflows/bridge-release.yml
```diff
@@ -150,17 +162,37 @@ jobs:
 
   publish-platform:
     name: Publish platform npm packages
-    needs: release
+    needs: [build, release]
```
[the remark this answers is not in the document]
AlexDochioiu: Same status here. After merging latest main, I explicitly reverted the branch-local workflow and docs divergence from this PR, so I am not carrying a manual publish path workflow change in this branch anymore. Any remaining workflow issue belongs in separate release-pipeline follow-up work.

.github/workflows/bridge-release.yml
```diff
@@ -60,7 +70,9 @@ jobs:
         id: version
         shell: bash
         run: |
-          if [[ "${GITHUB_REF_NAME}" == bridge-v* ]]; then
+          if [[ "${{ github.event.inputs.release_tag }}" == bridge-v* ]]; then
+            VERSION="${{ github.event.inputs.release_tag#bridge-v }}"
```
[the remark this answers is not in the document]
AlexDochioiu: Same status here as well. After merging latest main, I reverted the branch-local workflow and docs divergence, so I am not changing the release workflow further in this PR. If this expression issue still needs to be fixed, it should be handled in dedicated release-workflow follow-up work rather than in the push-memory branch.

bridge/app/lib/src/push/push_maintenance_loop.dart
```diff
@@ -0,0 +1,63 @@
...
+    _timer = Timer.periodic(_maintenanceInterval, (_) => runNow());
+  }
+
+  PushMaintenanceTelemetrySnapshot? get lastSnapshot => _lastSnapshot;
+
+  void dispose() {
+    _timer.cancel();
+  }
+
+  void runNow() {
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 4bee4e0. The maintenance loop now isolates each phase behind a guarded helper, so one failing step is logged and the later steps still run. I also added a regression test with a throwing tracker to prove telemetry and rate-limiter maintenance still execute after a prune failure.

bridge/app/lib/src/push/push_maintenance_loop.dart
```diff
@@ -0,0 +1,63 @@
...
+    final prunableRoots = _tracker.findPrunableRoots();
+    for (final prunableRoot in prunableRoots) {
+      final prunedSubtree = _tracker.pruneRootSubtree(rootSessionId: prunableRoot.rootSessionId);
+      _completionNotifier.cleanupPrunedRootSubtree(
+        rootSessionId: prunableRoot.rootSessionId,
+        prunedSessionIds: prunedSubtree.prunedSessionIds,
+      );
+    }
+
+    _rateLimiter.pruneStaleEntries();
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 4bee4e0. PushMaintenanceLoop.runNow now invokes pruneMessageRoleMetadata during the periodic maintenance pass, and there is regression coverage in push_notification_service_test.dart so the retention policy is exercised by the maintenance loop.

bridge/app/lib/src/push/push_session_state_tracker.dart
```diff
@@ -156,135 +152,118 @@ class PushSessionStateTracker {
...
+    subtreeSessionIds.forEach(_sessions.remove);
 
-    if (nextParentId != null) {
-      final parentState = _sessions[nextParentId];
-      if (parentState != null) {
-        parentState.childIds.add(sessionId);
-      }
+    for (final sessionState in _sessions.values) {
+      sessionState.childIds.removeWhere(subtreeSessionIds.contains);
     }
```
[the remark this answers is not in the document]
AlexDochioiu: I rechecked this on the merged-main branch and I am intentionally leaving this cleanup loop in place. It is not required for reachability correctness because traversal is backlink-valid, but it is still useful defensive hygiene to strip stale child IDs after a prune instead of retaining them indefinitely.

bridge/app/lib/src/push/push_session_state_tracker_maintenance.dart
```diff
@@ -0,0 +1,80 @@
...
+    pendingQuestionCount: sessions.values.where((sessionState) => sessionState.hasPendingQuestion).length,
+    pendingPermissionCount: sessions.values.where((sessionState) => sessionState.hasPendingPermission).length,
+    permissionRequestCount: permissionRequestCount,
+    previouslyBusyCount: sessions.values.where((sessionState) => sessionState.previouslyBusy).length,
+    latestAssistantTextCount: sessions.values.where((sessionState) => sessionState.latestAssistantText != null).length,
+    latestAssistantTextCharCount: sessions.values
+        .map((sessionState) => sessionState.latestAssistantText?.length ?? 0)
+        .fold(0, (sum, count) => sum + count),
+    messageRoleCount: messageRoles.length,
+    assistantMessageRoleCount: messageRoles.values.where((messageRole) => messageRole.role == "assistant").length,
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 4bee4e0. The session-derived telemetry counters are now aggregated in a single pass over sessions while preserving the existing snapshot fields and behavior. Verified with targeted push tests plus the full bridge analyze/test run.

bridge/app/lib/src/push/push_session_state_tracker_mutations.dart
```diff
@@ -0,0 +1,136 @@
...
+}
+
+void deleteTrackedSession({
+  required String sessionId,
+  required Map<String, PushTrackedSessionState> sessions,
+  required Map<String, PushTrackedMessageRole> messageRoles,
+  required Map<String, String> permissionRequestToSession,
+}) {
+  final removedSessionState = sessions.remove(sessionId);
+  if (removedSessionState != null) {
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 38046c7. deleteTrackedSession now clears stale child parent links even when the deleted session itself is not currently tracked, so later projectsSummary hydration can repair the relationship instead of being blocked by a leftover non-null parentId. Added regression coverage in push_session_state_tracker_test.dart.

bridge/app/lib/src/push/push_session_state_tracker_mutations.dart
```diff
@@ -0,0 +1,136 @@
...
+
+void applyTrackedProjectsSummaryChildLinks({
+  required List<ProjectActivitySummary> projects,
+  required Map<String, PushTrackedSessionState> sessions,
+}) {
+  for (final project in projects) {
+    for (final activeSession in project.activeSessions) {
+      stateForTrackedSession(sessionId: activeSession.id, sessions: sessions).projectId = project.id;
+      for (final childId in activeSession.childSessionIds) {
+        final childState = stateForTrackedSession(sessionId: childId, sessions: sessions);
```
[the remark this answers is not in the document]
AlexDochioiu: Fixed in 38046c7. projectsSummary-linked sessions now get touchedAt stamps through applyTrackedProjectsSummaryChildLinks, so roots discovered only from summary snapshots can age into the idle-prune path instead of being retained forever. Added a regression test that proves a projectsSummary-only root becomes prunable after the TTL.

bridge/app/lib/src/push/push_notification_service.dart
```diff
@@ -16,20 +19,35 @@ class PushNotificationService {
...
   late final StreamSubscription<String> _completionSubscription;
 
   PushNotificationService({
     required PushNotificationClient client,
     required PushRateLimiter rateLimiter,
     required PushSessionStateTracker tracker,
     required CompletionNotifier completionNotifier,
+    Duration maintenanceInterval = const Duration(minutes: 10),
+    int? Function()? rssBytesReader,
+    void Function(String)? debugLogger,
```
AlexDochioiu: just do the same logging like the rest of the app. No callback needed.

bridge/app/lib/src/push/push_notification_service_helpers.dart
```diff
@@ -0,0 +1,101 @@
+import "package:sesori_shared/sesori_shared.dart";
+
+String truncateTitle(String title, {int maxChars = 50}) {
```
AlexDochioiu: I hate the idea of public helper functions that have a very limited use-case. You should Wrap this in a `PushNotificationServiceHelpers` class and inject an instance as constructor arg for PushNotificationService.

Also update AGENTS.md based on this info to never do this unless the functions are genuinely reusable and not just targeted helpers for a given class/etc. Additionally, update aristotle's definition to also see this as an arch violation

bridge/app/lib/src/push/push_session_state_tracker_events.dart
```diff
@@ -0,0 +1,111 @@
+import "package:sesori_shared/sesori_shared.dart";
+
+import "push_session_state_tracker_mutations.dart";
+import "push_session_state_tracker_state.dart";
+
+void handleTrackedEvent({
```
AlexDochioiu: same comment as before about global top-level targeted functions not being allowed.

bridge/app/lib/src/push/push_session_state_tracker_graph.dart
```diff
@@ -0,0 +1,106 @@
+import "dart:collection";
+
+import "push_session_state_tracker_state.dart";
+
+Set<String> collectTrackedSubtreeSessionIds({
```
AlexDochioiu: same comment as before about global top-level targeted functions not being allowed.

bridge/app/lib/src/push/push_session_state_tracker_maintenance.dart
```diff
@@ -0,0 +1,106 @@
+import "push_session_state_tracker_graph.dart";
+import "push_session_state_tracker_models.dart";
+import "push_session_state_tracker_state.dart";
+
+List<PushPrunableRoot> findTrackedPrunableRoots({
```
AlexDochioiu: same comment as before about global top-level targeted functions not being allowed.

bridge/app/lib/src/push/push_session_state_tracker_mutations.dart
```diff
@@ -0,0 +1,147 @@
+import "package:sesori_shared/sesori_shared.dart";
+
+import "push_session_state_tracker_state.dart";
+
+void upsertTrackedSession({
```
AlexDochioiu: same comment as before about global top-level targeted functions not being allowed.

bridge/app/lib/src/bridge/routing/request_router.dart
```diff
@@ -114,7 +114,7 @@ class RequestRouter {
         sessionPersistenceService: sessionPersistenceService,
       ),
       SendPromptHandler(plugin),
-      AbortSessionHandler(plugin, onSessionAborted: onSessionAborted),
+      abortSessionHandler,
       GetProvidersHandler(providerRepository),
       GetAgentsHandler(plugin),
```
AlexDochioiu, line 119: It's weird that we create all handlers here except for `AbortSessionHandler` which is passed in constructor. Let's also just create it here instead.

bridge/app/lib/src/bridge/orchestrator.dart
```diff
@@ -140,21 +154,29 @@ class OrchestratorSession {
...
   final TokenRefresher _tokenRefresher;
   final StreamController<int> _bytesSentController;
   final FailureReporter _failureReporter;
   final PrSyncService _prSyncService;
   final SessionEventEnrichmentService _sessionEventEnrichmentService;
+  final SessionAbortService _sessionAbortService;
   StreamSubscription<dynamic>? _eventSubscription;
+  StreamSubscription<String>? _abortStartedSessionsSubscription;
+  StreamSubscription<String>? _abortedSessionsSubscription;
+  StreamSubscription<String>? _abortFailedSessionsSubscription;
```
AlexDochioiu: I'd rather just have a single final `CompositeSubscription`. Subscriptions just get added to it and then we cancel the entire CompositeSubscription on dispose. It's cleaner. I'd update AGENTS.md to include this as general preference.

bridge/app/lib/src/bridge/orchestrator.dart
```diff
@@ -168,19 +190,24 @@ class OrchestratorSession {
...
        _roomKey = roomKey,
        _sseManager = sseManager,
        _bytesSentController = bytesSentController,
        _failureReporter = failureReporter,
        _prSyncService = prSyncService,
+       _sessionAbortService = sessionAbortService,
        _router = RequestRouter(
          plugin: plugin,
          sessionRepository: sessionRepository,
+         abortSessionHandler: AbortSessionHandler(sessionAbortService: sessionAbortService),
```
AlexDochioiu, line 208: related to prev comment. I'd rather just pass `sessionAbortService` to `RequestRouter` and let it create the handler internally.

bridge/app/lib/src/bridge/orchestrator.dart
```diff
@@ -211,45 +237,58 @@ class OrchestratorSession {
...
+                  information: const [],
+                ),
+              );
+            },
+            onDone: () {
+              Log.w("[dbg] plugin event stream closed");
+            },
+          );
+      Log.d("[dbg] plugin event stream subscribed");
+      prChangesSubscription = _prSyncService.prChanges.listen((String projectId) {
```
AlexDochioiu: prob best to put all those subs in a composite subscription too.

bridge/app/lib/src/push/completion_push_listener.dart
```diff
@@ -0,0 +1,90 @@
...
+import "push_session_state_tracker.dart";
+
+class CompletionPushListener {
+  final PushSessionStateTracker _tracker;
+  final CompletionNotifier _completionNotifier;
+  final PushNotificationContentBuilder _contentBuilder;
+  final PushDispatcher _dispatcher;
+
+  // ignore: cancel_subscriptions, this listener stores the subscription and cancels it in dispose()
+  StreamSubscription<String>? _completionSubscription;
```
AlexDochioiu: use composite subscription. also get rid of the lint

bridge/app/lib/src/push/push_notification_client.dart
```diff
@@ -9,12 +10,21 @@ import "push_send_exception.dart";
...
-  }) : _tokenRefreshManager = tokenRefreshManager;
+  }) : _tokenRefreshManager = tokenRefreshManager,
+       _client = http.Client();
+
+  @visibleForTesting
+  PushNotificationClient.withClient({
+    required this.authBackendURL,
+    required TokenRefresher tokenRefreshManager,
+    required http.Client client,
+  }) : _tokenRefreshManager = tokenRefreshManager,
```
AlexDochioiu, line 18: We should have a single constructor that always takes the `required http.Client client`. The app should use a single http client for everything unless you have a good reason to create a dedicated http client for some stuff.

bridge/app/lib/src/push/push_notification_client.dart
```diff
@@ -46,4 +56,8 @@ class PushNotificationClient {
       body: jsonEncode(payload.toJson()),
     );
   }
+
+  Future<void> dispose() async {
+    _client.close();
```
AlexDochioiu: won't be needed if we share the same client as rest of the app

bridge/app/AGENTS.md
```diff
@@ -52,9 +52,12 @@ modules/
 | Process mgmt     | `lib/src/server/`                  | Spawns OpenCode, health poll, SIGTERM cleanup         |
 
 ## FILE SIZE
```
AlexDochioiu: file size requirements were pretty much removed from all other AGENTS.md so we should remove it from here too.

PR #132, opened 2026-04-26
feat: improve wake lock lid-close behavior and add laptop warnings

AlexDochioiu:
## Summary

Improves wake lock behavior around laptop lid-close sleep and adds platform-specific warnings when lid-close sleep cannot be prevented.

### Platform wake lock improvements

- **macOS**: Added `-s` flag to `caffeinate` (`caffeinate -s -w <pid>`) to prevent system sleep on AC power. Lid-close sleep on battery still occurs due to macOS hardware behavior.
- **Linux**: Added `handle-lid-switch` to `systemd-inhibit` (`--what=idle:sleep:handle-lid-switch`) so the inhibitor lock also blocks lid-close sleep events.
- **Windows**: No API change — `SetThreadExecutionState` cannot prevent lid-close sleep per Microsoft documentation.

### New capabilities

- **`DeviceTypeDetector`**: New utility that detects whether the current device is a laptop vs desktop/server:
  - macOS: checks `sysctl hw.model` for "MacBook"
  - Windows: checks `Win32_Battery` via PowerShell
  - Linux: checks `/sys/class/power_supply/BAT*`

- **`preventsLidCloseSleep`** property on `WakeLockClient`: Each platform implementation declares whether it can prevent lid-close sleep.
  - macOS: `false` (cannot reliably prevent lid-close)
  - Windows: `false` (cannot prevent lid-close)
  - Linux: `true` (uses `handle-lid-switch` inhibitor)

### Startup log warning

When wake lock is enabled and the device is a laptop on a platform that cannot prevent lid-close sleep, the bridge now logs:

```
[SleepPreventionService] wake lock enabled, but this platform cannot prevent the system from sleeping when the laptop lid is closed.
```

This warning only appears when:
- Sleep prevention mode is `always`
- The device is detected as a laptop
- The platform's wake lock does not claim lid-close prevention

## Testing

- Updated `MacOSWakeLockApiTest` for new `-s` flag and `preventsLidCloseSleep`
- Updated `LinuxWakeLockApiTest` for new `handle-lid-switch` flag and `preventsLidCloseSleep`
- Updated `WindowsWakeLockApiTest` for `preventsLidCloseSleep`
- Updated `WakeLockRepositoryTest` for `preventsLidCloseSleep` delegation
- Updated `SleepPreventionServiceTest` with new `DeviceTypeDetector` dependency and warning log coverage
- Added new `DeviceTypeDetectorTest` for all three platforms

---
## Summary by cubic
Improves wake lock behavior around laptop lid-close sleep and adds clear laptop warnings when the platform can’t block lid-close sleep. macOS now uses caffeinate -i -s -w <pid>; Linux blocks lid-close via systemd-inhibit; Windows unchanged; warnings fire only after wake lock is enabled.

- **New Features**
  - Linux: use systemd-inhibit --what=idle:sleep:handle-lid-switch to block lid-close.
  - macOS: use caffeinate -i -s -w <pid> to prevent idle and AC system sleep; lid-close not prevented.
  - Windows: no change; cannot prevent lid-close.
  - DeviceTypeDetector: detects laptops on macOS/Windows/Linux; caches isLaptop(); configurable Linux power supply path; requires PlatformChecker; logs on failures.
  - WakeLockClient.preventsLidCloseSleep: exposed via WakeLockRepository.
  - SleepPreventionService: warns only on laptops when lid-close can’t be prevented, after enabling wake lock; supports injected warningLogger; catches and logs detection errors so wake lock enable still succeeds.

<sup>Written for commit d491d824fe0f98b4a358bcf7e1ea3aed5ce87d68. Summary will update on new commits.</sup>

### DIFF
--- a/bridge/app/lib/src/bridge/foundation/device_type_detector.dart
+++ b/bridge/app/lib/src/bridge/foundation/device_type_detector.dart
@@ -1,5 +1,7 @@
 import "dart:io";
 
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show Log;
+
 import "process_runner.dart";
 
 /// Abstracts platform checks so they can be mocked in tests.
@@ -9,7 +11,7 @@ abstract class PlatformChecker {
   bool get isLinux;
 }
 
-class _DefaultPlatformChecker implements PlatformChecker {
+class DefaultPlatformChecker implements PlatformChecker {
   @override
   bool get isMacOS => Platform.isMacOS;
 
@@ -25,35 +27,50 @@ class _DefaultPlatformChecker implements PlatformChecker {
 class DeviceTypeDetector {
   final ProcessRunner _processRunner;
   final PlatformChecker _platformChecker;
+  final String _linuxPowerSupplyPath;
+  bool? _isLaptop;
 
   DeviceTypeDetector({
     required ProcessRunner processRunner,
-    PlatformChecker? platformChecker,
+    required PlatformChecker platformChecker,
+    String linuxPowerSupplyPath = "/sys/class/power_supply/",
   }) : _processRunner = processRunner,
-       _platformChecker = platformChecker ?? _DefaultPlatformChecker();
+       _platformChecker = platformChecker,
+       _linuxPowerSupplyPath = linuxPowerSupplyPath;
 
   Future<bool> isLaptop() async {
+    if (_isLaptop != null) {
+      return _isLaptop!;
+    }
     if (_platformChecker.isMacOS) {
-      return _isMacOSLaptop();
+      _isLaptop = await _isMacOSLaptop();
+    } else if (_platformChecker.isWindows) {
+      _isLaptop = await _isWindowsLaptop();
+    } else if (_platformChecker.isLinux) {
+      _isLaptop = await _isLinuxLaptop();
+    } else {
+      _isLaptop = false;
     }
-    if (_platformChecker.isWindows) {
-      return _isWindowsLaptop();
-    }
-    if (_platformChecker.isLinux) {
-      return _isLinuxLaptop();
-    }
-    return false;
+    return _isLaptop!;
   }
 
   Future<bool> _isMacOSLaptop() async {
     try {
-      final result = await _processRunner.run("sysctl", <String>["-n", "hw.model"]);
+      final result = await _processRunner.run(
+        "sysctl",
+        <String>["-n", "hw.model"],
+      );
       if (result.exitCode != 0) {
+        Log.w(
+          "[device-type] sysctl failed (exit ${result.exitCode}): "
+          "${result.stderr}",
+        );
         return false;
       }
       final model = (result.stdout as String).trim().toLowerCase();
       return model.contains("macbook");
-    } on Object {
+    } on Object catch (error) {
+      Log.w("[device-type] failed to detect macOS laptop: $error");
       return false;
     }
   }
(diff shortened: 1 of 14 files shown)


bridge/app/lib/src/bridge/foundation/device_type_detector.dart
```diff
@@ -0,0 +1,93 @@
...
+  Future<bool> _isMacOSLaptop() async {
+    try {
+      final result = await _processRunner.run("sysctl", <String>["-n", "hw.model"]);
+      if (result.exitCode != 0) {
+        return false;
+      }
+      final model = (result.stdout as String).trim().toLowerCase();
+      return model.contains("macbook");
+    } on Object {
+      return false;
```
AlexDochioiu, line 74: Let's also log some warning with the error before returning false

bridge/app/lib/src/bridge/foundation/device_type_detector.dart
```diff
@@ -0,0 +1,93 @@
...
+      return _isLinuxLaptop();
+    }
+    return false;
+  }
+
+  Future<bool> _isMacOSLaptop() async {
+    try {
+      final result = await _processRunner.run("sysctl", <String>["-n", "hw.model"]);
+      if (result.exitCode != 0) {
+        return false;
```
AlexDochioiu, line 68: Let's also log some warning before returning false. Include failure message if available

bridge/app/lib/src/bridge/foundation/device_type_detector.dart
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      final result = await _processRunner.run(
+        "powershell.exe",
+        <String>[
+          "-Command",
+          "Get-CimInstance Win32_Battery -ErrorAction SilentlyContinue | Measure-Object | Select-Object -ExpandProperty Count",
+        ],
+      );
+      if (result.exitCode != 0) {
+        return false;
```
AlexDochioiu, line 92: Let's also log some warning before returning false. Include failure message if available

bridge/app/lib/src/bridge/foundation/device_type_detector.dart
```diff
@@ -0,0 +1,93 @@
...
+      }
+      final count = int.tryParse((result.stdout as String).trim()) ?? 0;
+      return count > 0;
+    } on Object {
+      return false;
```
AlexDochioiu, line 98: Let's also log some warning with the error before returning false

AlexDochioiu:
Addressing review feedback:

**gemini (cache + config path):** Added isLaptop cache field and made linuxPowerSupplyPath configurable via constructor.

**gemini (async I/O):** Switched to await dir.exists() and await dir.list().toList() with justified ignore comment.

**codex (-i flag):** Added -i alongside -s so caffeinate prevents both idle sleep and system sleep on AC power.

**codex (required PlatformChecker):** Made platformChecker required and exposed DefaultPlatformChecker for explicit wiring.

**cubic (try/catch):** Moved _warnIfLidCloseSleepNotPrevented inside the wake lock try block, and added its own try/catch around isLaptop() so detection failures are logged separately and do not masquerade as wake lock errors.

**cubic (async sync):** Same as gemini - switched to async dir operations.

**cubic (warning assertion):** Added injectable warningLogger to SleepPreventionService and asserted the exact warning message in all warning tests.

**Alex (all 4 log warnings):** Added Log.w() calls in all failure paths of DeviceTypeDetector - non-zero exit codes include stderr, catch blocks include the error.

PR #216, opened 2026-06-10
fix: restore agent & model pickers by scoping GET /agent to a directory

AlexDochioiu:
## Problem

On the mobile new-session screen, the agent and model pickers stopped showing up entirely.

Root cause chain:

1. Newer OpenCode releases (verified on 1.17.1) resolve agents per project and return `500 UnknownError` when `GET /agent` carries no directory context (header or query param). Older releases tolerated the bare call.
2. The bridge's `OpenCodeApi.listAgents()` called `GET /agent` with no directory at all, so every agent fetch failed with a 500.
3. The failure was invisible: the bridge request handlers convert `PluginApiException` to an error response without logging, and the mobile `NewSessionCubit` maps the error to an empty agent list (plus a `catch (_) {}`).
4. The new-session screen hides the whole `AgentModelButtons` header when the agent list is empty (`new_session_screen.dart`), so **both** the agent and the model picker disappeared even though `POST /provider` still worked.

## Fix

**Bridge**
- `OpenCodeApi.listAgents` now sends the `x-opencode-directory` header (verified against a live OpenCode 1.17.1: bare call 500s, with header returns the agent list).
- Agent listing is threaded through `OpenCodeRepository` → `OpenCodeService` → plugin impl per the layering rules. The service falls back to the bridge CWD when no `projectId` is given, so older mobile clients keep working.
- `GetAgentsHandler` accepts an optional `projectId` query parameter (the project worktree directory), mirroring how providers/commands are already project-scoped.

**Mobile**
- `listAgents` passes the project id end to end (new-session composer and session-detail load), so project-scoped agents (`.opencode/agent`) resolve correctly per project.

**Logging** (the failure produced zero log lines before)
- `OpenCodeApiException` / `PluginApiException` now carry the upstream response body, so the OpenCode error payload (e.g. `{"name":"UnknownError",...}`) shows up in bridge logs.
- Bridge request handlers log upstream and unexpected failures before converting them to error responses.
- The new-session composer logs agent/provider/command load failures instead of silently emitting empty lists.

## Testing

- New unit tests: `listAgents` forwards the directory header / omits it when absent / surfaces the upstream body; `OpenCodeService.getAgents` projectId pass-through and CWD fallback; `GetAgentsHandler` forwards the `projectId` query param.
- All suites pass: bridge app (1168), opencode plugin (169), mobile module_core (311), mobile app (419). `dart analyze` clean across bridge and mobile.
- Live-verified the new API→repository→service path against the running OpenCode 1.17.1 instance: returns all 20 agents both project-scoped and via the CWD fallback.

---
## Summary by cubic
Restores the mobile agent and model pickers by scoping agent listing to a project. Adds POST /agent (ProjectIdRequest), deprecates GET /agent (CWD fallback), always sends the x-opencode-directory header, and updates mobile to query agents per project.

- **Bug Fixes**
  - Bridge: Implemented POST /agent (project-scoped); deprecated GET /agent with CWD fallback. AgentRepository forwards projectId or CWD; `OpenCodeApi.listAgents(directory)` always sets `x-opencode-directory`. Plugin `getAgents({projectId})` is required. Request handlers log upstream/unexpected errors; exceptions include upstream response bodies.
  - Mobile: `SessionApi.listAgents({projectId})` now POSTs and requires a project id. `SessionService.listAgents` is a non-null pass-through. `SessionDetailLoadService` resolves project from route/session and returns an empty list when unknown. New-session and session-detail pass projectId; the composer logs agent/provider/command load failures.

<sup>Written for commit 7ac6ce75b19752f72b1cc1521887f701ffb73873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/app/lib/src/bridge/repositories/agent_repository.dart
+++ b/bridge/app/lib/src/bridge/repositories/agent_repository.dart
@@ -1,5 +1,7 @@
+import "dart:io" as io;
+
 import "package:sesori_plugin_interface/sesori_plugin_interface.dart" show BridgePluginApi;
-import "package:sesori_shared/sesori_shared.dart" show Agents;
+import "package:sesori_shared/sesori_shared.dart" show Agents, StringExtensions;
 
 import "mappers/plugin_agent_mapper.dart";
 
@@ -8,8 +10,12 @@ class AgentRepository {
 
   AgentRepository({required BridgePluginApi plugin}) : _plugin = plugin;
 
-  Future<Agents> getAgents({String? projectId}) async {
-    final pluginAgents = await _plugin.getAgents(projectId: projectId);
+  Future<Agents> getAgents({required String? projectId}) async {
+    // A null/blank projectId comes from the deprecated GET /agent route,
+    // which carries no project context. Fall back to the bridge CWD, which
+    // plugins treat as the active project.
+    final resolvedProjectId = projectId?.normalize() ?? io.Directory.current.path;
+    final pluginAgents = await _plugin.getAgents(projectId: resolvedProjectId);
     final agents = pluginAgents.map((a) => a.toAgentInfo()).toList();
     return Agents(agents: agents);
   }
--- a/bridge/app/lib/src/bridge/repositories/agent_repository.dart
+++ b/bridge/app/lib/src/bridge/repositories/agent_repository.dart
@@ -8,8 +8,8 @@ class AgentRepository {
 
   AgentRepository({required BridgePluginApi plugin}) : _plugin = plugin;
 
-  Future<Agents> getAgents() async {
-    final pluginAgents = await _plugin.getAgents();
+  Future<Agents> getAgents({String? projectId}) async {
+    final pluginAgents = await _plugin.getAgents(projectId: projectId);
     final agents = pluginAgents.map((a) => a.toAgentInfo()).toList();
     return Agents(agents: agents);
   }
--- a/bridge/app/lib/src/bridge/routing/get_agents_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/get_agents_handler.dart
@@ -4,6 +4,10 @@ import "../repositories/agent_repository.dart";
 import "request_handler.dart";
 
 /// Handles `GET /agent` — returns all available agents from the plugin.
+///
+/// Accepts an optional `projectId` query parameter (the project worktree
+/// directory) so the plugin can resolve project-scoped agents. Older clients
+/// that omit it get the plugin's fallback behaviour.
 class GetAgentsHandler extends GetRequestHandler<Agents> {
   final AgentRepository _repository;
 
@@ -16,6 +20,6 @@ class GetAgentsHandler extends GetRequestHandler<Agents> {
     required Map<String, String> queryParams,
     required String? fragment,
   }) async {
-    return _repository.getAgents();
+    return _repository.getAgents(projectId: queryParams["projectId"]);
   }
 }
--- a/bridge/app/lib/src/bridge/routing/get_agents_handler.dart
+++ b/bridge/app/lib/src/bridge/routing/get_agents_handler.dart
@@ -5,12 +5,13 @@ import "request_handler.dart";
 
 /// Handles `GET /agent` — returns all available agents from the plugin.
 ///
-/// Accepts an optional `projectId` query parameter (the project worktree
-/// directory) so the plugin can resolve project-scoped agents. Older clients
-/// that omit it get the plugin's fallback behaviour.
+/// Carries no project context, so the repository falls back to the bridge
+/// CWD as the active project.
+@Deprecated("Use POST /agent with a ProjectIdRequest body (PostAgentsHandler)")
 class GetAgentsHandler extends GetRequestHandler<Agents> {
   final AgentRepository _repository;
 
+  @Deprecated("Use POST /agent with a ProjectIdRequest body (PostAgentsHandler)")
   GetAgentsHandler(this._repository) : super("/agent");
 
   @override
@@ -20,6 +21,6 @@ class GetAgentsHandler extends GetRequestHandler<Agents> {
     required Map<String, String> queryParams,
     required String? fragment,
   }) async {
-    return _repository.getAgents(projectId: queryParams["projectId"]);
+    return _repository.getAgents(projectId: null);
   }
 }
--- a/bridge/app/test/bridge/routing/request_router_test.dart
+++ b/bridge/app/test/bridge/routing/request_router_test.dart
@@ -280,6 +280,19 @@ void main() {
       expect(response.status, equals(200));
     });
 
+    test("routes POST /agent to PostAgentsHandler and forwards the project body", () async {
+      final response = await router.route(
+        makeRequest(
+          "POST",
+          "/agent",
+          body: jsonEncode({"projectId": "/repo"}),
+        ),
+      );
+
+      expect(response.status, equals(200));
+      expect(plugin.lastAgentsProjectId, equals("/repo"));
+    });
+
     test("routes POST /session/questions to GetSessionQuestionsHandler", () async {
       final response = await router.route(
         makeRequest(
--- a/bridge/app/test/bridge/routing/request_router_test.dart
+++ b/bridge/app/test/bridge/routing/request_router_test.dart
@@ -14,6 +14,7 @@ import "package:sesori_bridge/src/bridge/routing/abort_session_handler.dart";
 import "package:sesori_bridge/src/bridge/routing/get_agents_handler.dart";
 import "package:sesori_bridge/src/bridge/routing/get_commands_handler.dart";
 import "package:sesori_bridge/src/bridge/routing/get_session_diffs_handler.dart";
+import "package:sesori_bridge/src/bridge/routing/post_agents_handler.dart";
 import "package:sesori_bridge/src/bridge/routing/request_router.dart";
 import "package:sesori_bridge/src/bridge/routing/send_prompt_handler.dart";
 import "package:sesori_bridge/src/bridge/services/session_abort_service.dart";
@@ -100,6 +101,7 @@ void main() {
         projectRepository: projectRepository,
         providerRepository: providerRepository,
         getAgentsHandler: GetAgentsHandler(AgentRepository(plugin: plugin)),
+        postAgentsHandler: PostAgentsHandler(AgentRepository(plugin: plugin)),
         permissionRepository: permissionRepository,
         sessionPersistenceService: sessionPersistenceService,
         worktreeService: worktreeService,
@@ -419,6 +421,7 @@ void main() {
         projectRepository: projectRepository,
         providerRepository: providerRepository,
         getAgentsHandler: GetAgentsHandler(AgentRepository(plugin: plugin)),
+        postAgentsHandler: PostAgentsHandler(AgentRepository(plugin: plugin)),
         permissionRepository: permissionRepository,
         sessionPersistenceService: sessionPersistenceService,
         worktreeService: worktreeService,
--- a/bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
+++ b/bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
@@ -97,7 +97,12 @@ abstract class BridgePluginApi {
 
   Future<void> abortSession({required String sessionId});
 
-  Future<List<PluginAgent>> getAgents();
+  /// Returns the agents available for the given project.
+  ///
+  /// [projectId] is the project worktree directory. When `null`, the plugin
+  /// falls back to a directory it considers reasonable (e.g. its own CWD) —
+  /// kept for callers that predate project-scoped agent listing.
+  Future<List<PluginAgent>> getAgents({String? projectId});
 
   Future<List<PluginPendingQuestion>> getPendingQuestions({required String sessionId});
 
--- a/bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
+++ b/bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
@@ -99,10 +99,8 @@ abstract class BridgePluginApi {
 
   /// Returns the agents available for the given project.
   ///
-  /// [projectId] is the project worktree directory. When `null`, the plugin
-  /// falls back to a directory it considers reasonable (e.g. its own CWD) —
-  /// kept for callers that predate project-scoped agent listing.
-  Future<List<PluginAgent>> getAgents({String? projectId});
+  /// [projectId] identifies the project; its format is plugin-defined.
+  Future<List<PluginAgent>> getAgents({required String projectId});
 
   Future<List<PluginPendingQuestion>> getPendingQuestions({required String sessionId});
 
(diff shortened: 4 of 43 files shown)


bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -97,7 +97,12 @@ abstract class BridgePluginApi {
 
   Future<void> abortSession({required String sessionId});
 
-  Future<List<PluginAgent>> getAgents();
+  /// Returns the agents available for the given project.
+  ///
+  /// [projectId] is the project worktree directory. When `null`, the plugin
+  /// falls back to a directory it considers reasonable (e.g. its own CWD) —
+  /// kept for callers that predate project-scoped agent listing.
```
AlexDochioiu: "[projectId] is the project worktree directory" -- true only for OpenCode. This is not an assumption we should make for the plugin interface.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
```diff
@@ -97,7 +97,12 @@ abstract class BridgePluginApi {
...
+  Future<List<PluginAgent>> getAgents({String? projectId});
```
AlexDochioiu: let's make it required and non-nullable

mobile/module_core/lib/src/api/session_api.dart
```diff
@@ -17,10 +17,11 @@ class SessionApi {
 
   SessionApi({required RelayHttpApiClient client}) : _client = client;
 
-  Future<ApiResponse<Agents>> listAgents() {
+  Future<ApiResponse<Agents>> listAgents({String? projectId}) {
```
AlexDochioiu:
```suggestion
  Future<ApiResponse<Agents>> listAgents({required String? projectId}) {
```

mobile/module_core/lib/src/capabilities/session/session_service.dart
```diff
@@ -82,8 +82,8 @@ class SessionService {
     return _repository.getSessionStatuses();
   }
 
-  Future<ApiResponse<Agents>> listAgents() {
-    return _repository.listAgents();
+  Future<ApiResponse<Agents>> listAgents({String? projectId}) {
```
AlexDochioiu:
```suggestion
  Future<ApiResponse<Agents>> listAgents({required String? projectId}) {
```

mobile/module_core/lib/src/repositories/session_repository.dart
```diff
@@ -93,8 +93,8 @@ class SessionRepository {
     return _api.getSession(sessionId: sessionId);
   }
 
-  Future<ApiResponse<Agents>> listAgents() {
-    return _api.listAgents();
+  Future<ApiResponse<Agents>> listAgents({String? projectId}) {
```
AlexDochioiu:
```suggestion
  Future<ApiResponse<Agents>> listAgents({required String? projectId}) {
```

bridge/app/lib/src/bridge/repositories/agent_repository.dart
```diff
@@ -8,8 +8,8 @@ class AgentRepository {
 
   AgentRepository({required BridgePluginApi plugin}) : _plugin = plugin;
 
-  Future<Agents> getAgents() async {
-    final pluginAgents = await _plugin.getAgents();
+  Future<Agents> getAgents({String? projectId}) async {
```
AlexDochioiu:
```suggestion
  Future<Agents> getAgents({required String? projectId}) async {
```

bridge/app/lib/src/bridge/routing/get_agents_handler.dart
```diff
@@ -16,6 +20,6 @@ class GetAgentsHandler extends GetRequestHandler<Agents> {
     required Map<String, String> queryParams,
     required String? fragment,
   }) async {
-    return _repository.getAgents();
+    return _repository.getAgents(projectId: queryParams["projectId"]);
   }
```
AlexDochioiu, line 25: project id cannot safely be query param. Let's
1. Keep `GetAgentsHandler` and mark as deprecated. Send null projectId so that we use CWD or similar in the end.
2. Create a new `PostAgentsHandler` (similar to other request) taking ProjectIdRequest body (like other handlers) and update the app to start using this.

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
```diff
@@ -327,8 +327,14 @@ class OpenCodeApi {
...
   }
 
-  Future<List<AgentInfo>> listAgents() async {
-    final response = await _client.get(Uri.parse("$serverURL/agent"), headers: _authHeaders);
+  Future<List<AgentInfo>> listAgents({required String? directory}) async {
+    // OpenCode resolves agents per project; newer releases reject requests
+    // that carry no directory context with a 500.
+    final headers = <String, String>{
+      ..._authHeaders,
+      _directoryOpenCodeHeader: ?directory,
```
AlexDochioiu: We must make sure to always send some directory cause it just doesn't work otherwise.

```suggestion
  Future<List<AgentInfo>> listAgents({required String directory}) async {
    // OpenCode resolves agents per project; newer releases reject requests
    // that carry no directory context with a 500.
    final headers = <String, String>{
      ..._authHeaders,
      _directoryOpenCodeHeader: directory,
```

bridge/sesori_plugin_opencode/lib/src/opencode_repository.dart
```diff
@@ -64,6 +66,11 @@ class OpenCodeRepository {
     return commands.map<PluginCommand>(_mapCommand).toList();
   }
 
+  Future<List<PluginAgent>> getAgents({required String? directory}) async {
+    final agents = await _api.listAgents(directory: directory?.normalize());
+    return agents.map((agent) => agent.toPlugin()).toList();
```
AlexDochioiu:
```suggestion
  Future<List<PluginAgent>> getAgents({required String directory}) async {
    final agents = await _api.listAgents(directory: directory.normalize());
    return agents.map((agent) => agent.toPlugin()).toList();
```

bridge/sesori_plugin_opencode/lib/src/opencode_service.dart
```diff
@@ -33,6 +36,21 @@ class OpenCodeService {
...
+  /// When [projectId] is `null` (callers that predate project-scoped agent
+  /// listing) we fall back to the bridge's own CWD, which yields the global
+  /// agent set.
+  Future<List<PluginAgent>> getAgents({required String? projectId}) {
+    var directory = projectId;
+    if (directory == null) {
+      directory = io.Directory.current.path;
+      Log.d("getAgents: no projectId given, falling back to bridge CWD: $directory");
+    }
+    return repository.getAgents(directory: directory);
```
AlexDochioiu, line 53: `var` should be pretty much forbidden. Rarely it's a good idea to use mutable stuff like this.
```suggestion
  Future<List<PluginAgent>> getAgents({required String? projectId}) {
    final String directory;
    if (projectId == null) {
      directory = io.Directory.current.path;
      Log.d("getAgents: no projectId given, falling back to bridge CWD: $directory");
    } else {
      directory = projectId;
    }
    return repository.getAgents(directory: directory);
```

bridge/app/lib/src/bridge/repositories/agent_repository.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Applied — `projectId` is now `required String?` on `AgentRepository.getAgents`.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

`BridgePluginApi.getAgents` now takes `required String projectId` (non-nullable). The null/fallback handling lives above the interface: the deprecated GET route resolves the bridge CWD in `AgentRepository` before calling the plugin.

bridge/sesori_plugin_interface/lib/src/bridge_plugin.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Reworded to "[projectId] identifies the project; its format is plugin-defined" — no worktree-directory assumption in the interface docs anymore.

bridge/app/lib/src/bridge/routing/get_agents_handler.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Done as suggested:
1. `GetAgentsHandler` (GET /agent) is kept and marked `@Deprecated`; it now passes a null projectId, which `AgentRepository` resolves to the bridge CWD.
2. Added `PostAgentsHandler` (POST /agent) taking a `ProjectIdRequest` body, mirroring `GetProvidersHandler`, and registered it in the router. The app now POSTs when it has a projectId and only falls back to the deprecated GET when it has no project context at all.

bridge/sesori_plugin_opencode/lib/src/opencode_api.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

`listAgents` now takes `required String directory` and always sends the `x-opencode-directory` header. The nullable handling moved up to `OpenCodeService`, which resolves a CWD fallback before calling down.

bridge/sesori_plugin_opencode/lib/src/opencode_repository.dart
AlexDochioiu: [Sesori reply] Partially addressed (in commit 778dce50)

Made `directory` `required String` as suggested. I did not call `.normalize()` here: it returns `String?` (null for blank strings), which no longer fits the non-nullable API parameter. Normalization happens in `OpenCodeService.getAgents` instead, where a blank/null projectId now triggers the CWD fallback rather than sending an empty header.

bridge/sesori_plugin_opencode/lib/src/opencode_service.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Replaced `var` with a `final String directory` + if/else as suggested, and folded `.normalize()` in so a blank projectId also takes the CWD fallback.

mobile/module_core/lib/src/api/session_api.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

`projectId` is now required. The method also switched to POST /agent with a `ProjectIdRequest` body per the handler discussion; it only uses the deprecated GET route when projectId is null (no project context available).

mobile/module_core/lib/src/capabilities/session/session_service.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Applied — `required String? projectId`.

mobile/module_core/lib/src/repositories/session_repository.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Applied — `required String? projectId`.

mobile/module_core/lib/src/services/session_detail_load_service.dart
```diff
@@ -57,7 +57,7 @@ class SessionDetailLoadService {
         _repository.getPendingPermissions(),
         _repository.getChildren(sessionId: sessionId),
         _repository.getSessionStatuses(),
-        _repository.listAgents(),
+        _repository.listAgents(projectId: routeProjectId),
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

Implemented the suggested deferral: `agentsFuture` is started eagerly when the route carries a projectId, otherwise the agents request waits for the session's project context and uses `effectiveProjectId` — same pattern as commands. Sessions opened from deep links/notifications now get project-scoped agents instead of the global fallback.

mobile/module_core/test/helpers/test_helpers.dart
```diff
@@ -122,7 +122,7 @@ void delegateSessionRepositoryToService({
     (invocation) => service.getChildren(sessionId: invocation.namedArguments[#sessionId]! as String),
   );
   when(() => repository.getSessionStatuses()).thenAnswer((_) => service.getSessionStatuses());
-  when(() => repository.listAgents()).thenAnswer((_) => service.listAgents());
+  when(() => repository.listAgents(projectId: any(named: "projectId"))).thenAnswer((_) => service.listAgents());
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 778dce50)

The delegation now forwards the captured argument via `invocation.namedArguments[#projectId] as String?`, consistent with the `listProviders`/`listCommands` delegations. Applied to both `module_core/test/helpers/test_helpers.dart` and `app/test/helpers/test_helpers.dart`.

bridge/app/test/bridge/routing/request_router_test.dart
```diff
@@ -100,6 +101,7 @@ void main() {
         projectRepository: projectRepository,
         providerRepository: providerRepository,
         getAgentsHandler: GetAgentsHandler(AgentRepository(plugin: plugin)),
+        postAgentsHandler: PostAgentsHandler(AgentRepository(plugin: plugin)),
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 0439ed62)

Added a routing test that POSTs /agent with a `ProjectIdRequest` body through the router and asserts a 200 plus that the projectId reaches the plugin.

mobile/module_core/lib/src/api/session_api.dart
```diff
@@ -17,10 +17,19 @@ class SessionApi {
...
-    return _client.get(
+  Future<ApiResponse<Agents>> listAgents({required String? projectId}) {
+    if (projectId == null) {
+      // Deprecated bridge route for callers with no project context; the
+      // bridge resolves a fallback project (its CWD) itself.
+      return _client.get(
+        "/agent",
+        fromJson: Agents.fromJson,
+      );
+    }
```
AlexDochioiu: can we not have a non-nullable projectId as arg? I want to not use the deprecated `GET /agent` anymore

AlexDochioiu: [Sesori reply] Addressed (in commit add0297f)

`SessionApi.listAgents` (and `SessionRepository`) now take `required String projectId` and always POST — the GET branch is gone, so the app never touches the deprecated route. For the one caller that can lack project context (session detail when neither the route nor the session context yields a project), `SessionDetailLoadService`/`SessionService` return an empty agent list instead, mirroring the existing `listCommands` pattern.

PR #234, opened 2026-06-12
fix(mobile): return split detail pane to empty state after deleting the open session

AlexDochioiu:
## Problem

In the tablet/desktop split view, long-pressing the currently open session in the left list pane and deleting it left the right detail pane rendering the just-deleted session. The pane is driven purely by the route, and the delete flow never updated it.

## Fix

After a successful delete, `_deleteSession` now calls a new `_closeDeletedSessionRoute` helper: if the current location's `sessionId` path param matches the deleted session, it navigates back to the sessions route, which the split shell renders as the empty "select a session" panel.

- Also covers the diffs subroute (right pane showing diffs of the deleted session)
- Covers the force-delete (409 → retry) path, which re-enters `_deleteSession`
- No-op in narrow layouts (sessions route has no `sessionId` param) and when deleting a session that isn't the one open
- Preserves the `?name=` query param so the left pane header keeps the project name

## Tests

Two regression tests added to `adaptive_session_list_navigation_test.dart` using the existing `AdaptiveSessionRouterTestHarness`:
- deleting the selected session returns the wide detail pane to the empty state
- deleting an unselected session keeps the open detail pane

Full `flutter test` suite in `mobile/app`: 481/481 passing. `dart analyze` clean on changed files. Architectural review (aristotle-impl-review): approved.

---
## Summary by cubic
Fixes the split view so deleting the open session returns the right pane to the empty “select a session” state. Also standardizes route path-param names via shared constants.

- **Bug Fixes**
  - Close the deleted session route after a successful delete.
  - If the current route’s `sessionId` matches, navigate back to the sessions route and preserve the `?name=` query; covers detail and diffs, including force-delete retries; no-op on narrow layouts or when deleting a different session.
  - Added two widget tests for wide mode to verify clearing the detail pane vs. keeping it when deleting another session.

- **Refactors**
  - Extracted `projectId` and `sessionId` path-param names into shared consts and updated route definitions and lookups to use them.

<sup>Written for commit 1ebb7203c4959e018571fa61a1afa4426298d75b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mobile/app/lib/features/session_list/session_list_actions.dart
+++ b/mobile/app/lib/features/session_list/session_list_actions.dart
@@ -217,9 +217,9 @@ Future<void> _deleteSession({
 /// already current when deleting from the list, so this is a no-op there.
 void _closeDeletedSessionRoute({required BuildContext context, required String sessionId}) {
   final routeState = GoRouterState.of(context);
-  if (routeState.pathParameters["sessionId"] != sessionId) return;
+  if (routeState.pathParameters[sessionIdPathParam] != sessionId) return;
 
-  final projectId = routeState.pathParameters["projectId"];
+  final projectId = routeState.pathParameters[projectIdPathParam];
   if (projectId == null) return;
 
   context.goRoute(
--- a/mobile/app/lib/features/session_list/session_list_actions.dart
+++ b/mobile/app/lib/features/session_list/session_list_actions.dart
@@ -185,6 +185,7 @@ Future<void> _deleteSession({
     ScaffoldMessenger.of(context)
       ..clearSnackBars()
       ..showSnackBar(SnackBar(content: Text(loc.sessionListDeleted)));
+    _closeDeletedSessionRoute(context: context, sessionId: sessionId);
     return;
   }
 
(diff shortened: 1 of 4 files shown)


mobile/app/lib/features/session_list/session_list_actions.dart
```diff
@@ -206,3 +207,25 @@ Future<void> _deleteSession({
...
+/// In the split layout the detail pane would otherwise keep rendering the
+/// deleted session; returning to the sessions route swaps it for the empty
+/// "select a session" panel. In the narrow layout the sessions route is
+/// already current when deleting from the list, so this is a no-op there.
+void _closeDeletedSessionRoute({required BuildContext context, required String sessionId}) {
+  final routeState = GoRouterState.of(context);
+  if (routeState.pathParameters["sessionId"] != sessionId) return;
+
+  final projectId = routeState.pathParameters["projectId"];
+  if (projectId == null) return;
```
AlexDochioiu, line 223: I don't like magic strings. Can those path parameters be defined in `const` variables and referenced everywhere from the same variable?

AlexDochioiu: [Sesori reply] Addressed (in commit 1ebb7203)

Added public `projectIdPathParam` and `sessionIdPathParam` consts in `app_routes.dart` next to the existing `projectNameQueryParam`, and referenced them everywhere: the `AppRouteDef` route templates, the per-route-class private aliases, the route segments and `pathParameters` lookups in `app_router.dart`, and the new `_closeDeletedSessionRoute` helper.

PR #260, opened 2026-06-14
feat(bridge): selectable release track (stable|internal) for the auto-updater

AlexDochioiu:
## Summary

Lets a bridge user choose which GitHub release channel the **runtime auto-updater** follows, via a new CLI command. The internal pre-release channel (`vX.Y.Z-internal.N`) is already produced by CI on every `main` merge but was invisible to the updater until now.

```
sesori-bridge config track internal   # follow internal pre-releases
sesori-bridge config track stable      # back to stable (default)
sesori-bridge config track             # print the current track
```

## Behavior

- **`stable`** (default, unchanged): only stable `vX.Y.Z` releases.
- **`internal`**: newest of `{stable} ∪ {vX.Y.Z-internal.N}` by semver. Only the `-internal.*` line is honored — other pre-release kinds (`-rc`, `-beta`) are ignored.
- **Forward-only**: switching back to `stable` never auto-downgrades a binary that's ahead of latest stable (avoids an older binary opening a newer Drift schema). You converge to stable once it catches up.
- Track is read once at startup and injected into `ReleaseRepository`; a change applies on **next restart** (consistent with the existing "updates apply at next startup" model). The active track is logged at startup.
- The 10-min release cache is now **track-aware**, so a recent stable check can't mask an internal build.
- Installers (`install.sh` / `install.ps1` / npm bootstrap) remain **stable-only** by design.

## Design

The full design was driven by an interview and captured before implementation. Decisions: internal = superset of stable; runtime-updater-only scope; forward-only (no downgrade); `stable`/`internal` naming; `config` becomes a command group (`config track`, `config edit`); next-restart application; no env override; warn-on-internal + startup log; `-internal.*`-only matching.

## Implementation

- New `ReleaseTrack { stable, internal }` enum (updater foundation), persisted in `BridgeSettings` (`releaseTrack`, default `stable`, always serialized, tolerant parse).
- `BridgeSettingsRepository.updateReleaseTrack` (read-modify-write).
- `ReleaseRepository` gains a `track` + `_isEligible` predicate; `CachedRelease` gains a `track` field (old caches invalidate gracefully).
- `config` restructured into a branch command (`config track` / `config edit`); bare `config` now lists subcommands.
- Docs updated (README / INSTALL / RELEASING).

No DB schema change.

## Review & verification

- ✅ `aristotle-plan-review`: approved.
- ✅ `aristotle-impl-review`: approved (one magic-string finding fixed in 49a7381b).
- ✅ `dart analyze --fatal-infos` clean; `dart test` 1329 passing (incl. new stable/internal selection, forward-only, cache-track, settings, and `updateReleaseTrack` tests).

## Out of scope (deferred)

Installers honoring the track; `SESORI_TRACK` env override; auto-downgrade; non-`internal` pre-release channels.

---
## Summary by cubic
Adds a selectable update track for the runtime auto-updater: choose `stable` (default) or `internal` via `sesori-bridge config track`. Internal builds can auto‑update without affecting stable users; updates remain forward‑only.

- **New Features**
  - New CLI: `sesori-bridge config track [stable|internal]` (no args prints the current track); invalid args show a usage error; `config edit` opens the config.
  - Persisted `releaseTrack` with default `stable`; read at startup and logged (warning for `internal`, debug for `stable`). Missing/unreadable value resolves silently to `stable`; unexpected value falls back to `stable` with an error on stderr so stdout stays clean and startup never blocks.
  - Updater honors track: `internal` = newest of stable and `vX.Y.Z-internal.N`; other prereleases ignored; no auto‑downgrade when switching back to `stable`.
  - Release cache is track‑aware and stores the track; legacy cache files default to `stable`; installers (shell/PowerShell/`@sesori/bridge`) remain stable‑only.

- **Migration**
  - If you used `sesori-bridge config` to open the config, use `sesori-bridge config edit`.
  - After changing the track, restart the bridge to apply.

<sup>Written for commit 366323f385595e17a621ba55e278bff4d9499414. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -30,6 +30,7 @@ import 'package:sesori_bridge/src/server/repositories/terminal_prompt_repository
 import 'package:sesori_bridge/src/server/services/bridge_instance_service.dart';
 import 'package:sesori_bridge/src/services/bridge_config_service.dart';
 import 'package:sesori_bridge/src/services/sleep_prevention_service.dart';
+import 'package:sesori_bridge/src/updater/foundation/release_track.dart';
 import 'package:sesori_bridge/src/version.dart';
 import 'package:sesori_plugin_interface/sesori_plugin_interface.dart'
     show BridgePluginDescriptor, Log, LogLevel, PluginConfig, PluginConfigException, ProcessUser, ServerClock;
@@ -271,6 +272,19 @@ class ConfigCommand extends cli.Command<void> {
   final name = 'config';
 
   @override
+  final description = 'Manage bridge configuration';
+
+  ConfigCommand() {
+    addSubcommand(ConfigTrackCommand());
+    addSubcommand(ConfigEditCommand());
+  }
+}
+
+class ConfigEditCommand extends cli.Command<void> {
+  @override
+  final name = 'edit';
+
+  @override
   final description = 'Open the bridge configuration file in your default editor';
 
   @override
@@ -289,6 +303,52 @@ class ConfigCommand extends cli.Command<void> {
 
     final configFilePath = await configService.openConfigFile();
     stdout.writeln('Opening config file at $configFilePath');
+  }
+}
+
+class ConfigTrackCommand extends cli.Command<void> {
+  @override
+  final name = 'track';
+
+  @override
+  final description = 'Show or set the bridge update track (stable|internal)';
+
+  @override
+  Future<void> run() async {
+    final rest = argResults!.rest;
+    final repository = BridgeSettingsRepository(api: BridgeSettingsApi());
+
+    if (rest.isEmpty) {
+      final settings = await repository.loadSettings();
+      stdout.writeln('Release track: ${settings.releaseTrack.wireValue}');
+      return;
+    }
+
+    if (rest.length > 1) {
+      usageException('Expected a single track value: stable or internal.');
+    }
+
+    final ReleaseTrack track = _parseTrackArgument(rest.single);
+    await repository.updateReleaseTrack(track: track);
+    stdout.writeln('Release track set to ${track.wireValue}.');
+    if (track == ReleaseTrack.internal) {
+      stdout.writeln(
+        'Warning: internal builds are pre-release and may be unstable. '
+        'The bridge will auto-update to the latest internal build on next start.',
+      );
+    }
+    stdout.writeln('Restart sesori-bridge to apply.');
+  }
+
+  ReleaseTrack _parseTrackArgument(String value) {
+    switch (value) {
+      case 'stable':
+        return ReleaseTrack.stable;
+      case 'internal':
+        return ReleaseTrack.internal;
+      default:
+        usageException('Track must be "stable" or "internal".');
+    }
   }
 }
 
--- a/bridge/app/bin/bridge.dart
+++ b/bridge/app/bin/bridge.dart
@@ -315,7 +315,11 @@ class ConfigTrackCommand extends cli.Command<void> {
 
   @override
   Future<void> run() async {
-    final rest = argResults!.rest;
+    final results = argResults;
+    if (results == null) {
+      usageException('Unable to read command arguments.');
+    }
+    final rest = results.rest;
     final repository = BridgeSettingsRepository(api: BridgeSettingsApi());
 
     if (rest.isEmpty) {
--- a/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
+++ b/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
@@ -184,11 +184,12 @@ class BridgeRuntimeRunner {
         return 1;
       }
 
-      final releaseTrack = await _resolveReleaseTrack();
+      final settingsRepository = BridgeSettingsRepository(api: BridgeSettingsApi());
+      final releaseTrack = await _readReleaseTrack(settingsRepository: settingsRepository);
       if (releaseTrack == ReleaseTrack.internal) {
         Log.w("Release track: internal (pre-release auto-updates enabled)");
       } else {
-        Log.i("Release track: ${releaseTrack.wireValue}");
+        Log.d("Release track: ${releaseTrack.wireValue}");
       }
 
       final updateService = _createUpdateService(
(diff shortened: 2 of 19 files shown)


bridge/app/bin/bridge.dart
```diff
@@ -292,6 +306,52 @@ class ConfigCommand extends cli.Command<void> {
...
+class ConfigTrackCommand extends cli.Command<void> {
+  @override
+  final name = 'track';
+
+  @override
+  final description = 'Show or set the bridge update track (stable|internal)';
+
+  @override
+  Future<void> run() async {
+    final rest = argResults!.rest;
```
AlexDochioiu: No bang operators allowed. This can crash runtime. Handle properly the null cases with proper user facing messages

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -181,10 +184,18 @@ class BridgeRuntimeRunner {
         return 1;
       }
 
+      final releaseTrack = await _resolveReleaseTrack();
+      if (releaseTrack == ReleaseTrack.internal) {
+        Log.w("Release track: internal (pre-release auto-updates enabled)");
+      } else {
+        Log.i("Release track: ${releaseTrack.wireValue}");
```
AlexDochioiu: No need to log this for regular users on stable. Make this a debug log

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -431,10 +442,21 @@ class BridgeRuntimeRunner {
     return attemptStart(attempt: 1);
   }
 
+  static Future<ReleaseTrack> _resolveReleaseTrack() async {
+    try {
+      final settings = await BridgeSettingsRepository(api: BridgeSettingsApi()).loadSettings();
```
AlexDochioiu: Awful practice for every method invocation to create new repo and api

bridge/app/lib/src/updater/foundation/release_track.dart
```diff
@@ -0,0 +1,23 @@
...
+  String get wireValue => name;
+
+  /// Maps a persisted/CLI string to a track, defaulting unknown or null
+  /// values to [stable] so a malformed config can never silently opt a user
+  /// into pre-releases.
+  static ReleaseTrack fromWire(String? value) {
+    return switch (value) {
+      'stable' => ReleaseTrack.stable,
+      'internal' => ReleaseTrack.internal,
+      _ => ReleaseTrack.stable,
```
AlexDochioiu: Should also log a warning of an unexpected value is used. Must handle the "no value" case correctly, which is stable release

AlexDochioiu (commented):
Left more things to address. Also, make sure when track is internal, our API call to GitHub returns internal releases too (not sure if it does by default?)

bridge/app/bin/bridge.dart
AlexDochioiu: [Sesori reply] Addressed (in commit eb7f6353)

Removed the argResults! null-assertion in ConfigTrackCommand; it now guards argResults and raises a usage error instead of risking a runtime crash.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
AlexDochioiu: [Sesori reply] Addressed (in commit eb7f6353)

Demoted the stable-track startup line to Log.d; the internal track stays a warning so opted-in users still see it.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
AlexDochioiu: [Sesori reply] Addressed (in commit eb7f6353)

BridgeSettingsRepository/BridgeSettingsApi is now constructed once in the runtime composition root (run()) and injected into the track read, instead of being built inside the helper.

bridge/app/lib/src/updater/foundation/release_track.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 2894edcd)

fromWire now treats a missing value (null) as the silent stable default, and logs an unrecognized non-null value before falling back to stable. I used Log.e (stderr) rather than Log.w because track parsing runs during plugin selection for every command, and Log.w writes to stdout which must stay machine-clean for --version/--help (same reason peekSettings uses Log.e for bad configs). Added tests.

bridge/app/lib/src/updater/models/cached_release.dart
```diff
@@ -19,6 +19,11 @@ sealed class CachedRelease with _$CachedRelease {
     /// The release asset name this cache entry was resolved for.
     required String assetName,
 
+    /// The release track ([ReleaseTrack.wireValue]) this cache entry was
+    /// resolved for. A cache entry from a different track is ignored so a
+    /// recent stable check can't mask an internal build (or vice versa).
+    required String track,
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 30cde4ea)

Switched to @Default('stable'). Minor correction: UpdateCacheApi.read() already catches TypeError, so an old cache would not have crashed startup; it would have been discarded and refetched. The default is still better since it avoids throw/catch as migration control-flow and keeps a stable-track user's pre-upgrade cache usable (every legacy cache was the stable channel). Added a regression test.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -181,10 +184,19 @@ class BridgeRuntimeRunner {
         return 1;
       }
 
+      final settingsRepository = BridgeSettingsRepository(api: BridgeSettingsApi());
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 366323f3)

Good catch. The BridgeSettingsApi() construction and the settings read are now both inside the try/catch in the composition root, so an unset HOME (or any settings failure) falls back to the stable track instead of aborting startup.

AlexDochioiu:
[Sesori reply] Re: your review note — "make sure when track is internal, our API call to GitHub returns internal releases too (not sure if it does by default?)".

Yes — confirmed, no fetch change needed. GitHubReleasesApi.fetchReleases() calls GET /repos/sesori-ai/sesori_apps_monorepo/releases (the list endpoint), which returns ALL releases including pre-releases; there is no server-side prerelease filter. I verified against the live repo and the response currently includes v1.0.9-internal.91 with prerelease=true alongside the 10 stable releases. (The endpoint that WOULD exclude pre-releases is /releases/latest, which we deliberately do not use.)

So pre-releases are always fetched; the track only changes client-side selection: on internal, _isEligible keeps stable releases plus -internal.* pre-releases; on stable it keeps stable only.

PR #386, opened 2026-07-06
feat(app): migrate all bottom sheets to PregoBottomSheet

daniil-shumko:
## Summary

Migrates every bottom sheet in `client/app` to the design-system `PregoBottomSheet`/`showPregoBottomSheet` component (glass nav header, rounded `bgSecondary` surface, scroll-edge fade, drag-to-dismiss), one commit per sheet:

- **Pickers:** command, model (agent + variant picker sheets deleted as dead code — replaced by the composer glass menus; the model sheet stays as the menu's full-screen search escalation)
- **Session detail modals:** reasoning ("Thought"), permission, question
- **List actions:** session/project action menus, archive/delete confirms, rename session/project, add project
- `module_prego`: exposes `PregoBottomSheet.contentTopInset` so sheets hosting their own scroll view can compute a bounded body height; removes the now-unused `showAppModalBottomSheet` helper

## Sizing & safe-area fixes found during live verification

- Content-sized sheets: short content wraps, tall content grows to the status bar then scrolls (reasoning/question/permission previously used fixed fractional heights)
- `showModalBottomSheet(useSafeArea: false)` strips the top inset from **both** `padding` and `viewPadding` inside the route, so full-height sheets drew into the status bar; the presenting context's inset is now captured in each `show()` and threaded through
- Bottom-edge scrollables (reasoning sheet) were hard-clipped at the safe-area line on rounded devices; the body now runs to the screen edge with the home-indicator inset as trailing scroll padding, so content scrolls behind the indicator (matches the picker pattern)

## Verification

- Live E2E on iPhone 17 Pro Max simulator via a local bridge: every migrated sheet exercised (menus, renames, archive+undo, model search + keyboard, command filter, short/medium/long reasoning, live question round-trip)
- `flutter test`: 536/536 green; `flutter analyze` clean
- Safe-area regression tests drive the real `show()` route under simulated view insets; each half of the geometry fix is mutation-tested
- `aristotle-impl-review`: APPROVED (no architectural violations)

https://claude.ai/code/session_01P4bCZtUJPWKGPJedc4FvJM

---
## Summary by cubic
Migrates all bottom sheets in `client/app` to `PregoBottomSheet`/`showPregoBottomSheet` for a consistent glass header, rounded `bgSecondary` surface, and safer insets. Removes legacy picker sheets and `showAppModalBottomSheet`; exposes `PregoBottomSheet.contentTopInset` to size scroll bodies correctly.

- **Refactors**
  - Move command/model pickers, session detail modals (reasoning, permission, question), and list actions (session/project menus, archive/delete, rename, add project) to `PregoBottomSheet`.
  - Titles move to the header; bodies go full‑bleed where tiles/fields self‑pad; drag‑to‑dismiss.
  - Use `MediaQuery.heightOf` where only height is read to avoid rebuilds on width-only metric changes.

- **Bug Fixes**
  - Size sheets to content and cap just below the status bar; capture the presenting context’s top inset to avoid header overlap.
  - Let content scroll behind the home indicator with trailing scroll padding (reasoning); remove hard bottom clip.
  - Keep action rows pinned under long content in permission/question sheets; restore contrast for unselected question option cards on the `bgSecondary` surface.
  - Command picker: cap the inner list at the space below the sheet header so the outer sheet doesn’t gain its own scroll.

<sup>Written for commit 522f5802188b7b6d7bf4111e0d6f7efcf02134f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/386?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


client/app/lib/features/session_detail/widgets/permission_modal.dart
```diff
@@ -39,13 +51,19 @@ class PermissionModal extends StatelessWidget {
...
+      // PregoBottomSheet paints the rounded surface; keep the route
+      // transparent. The sheet caps itself below the status bar.
       backgroundColor: Colors.transparent,
+      useSafeArea: false,
       builder: (_) => PermissionModal(
         permission: permission,
         onReply: onReply,
+        topInset: topInset,
       ),
     );
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed: dismissibility here is pre-existing behavior (main presents this sheet with the same defaults), and a swiped-away request is not lost — it stays in pendingPermissions and a persistent SessionDetailPendingBanner pinned under the app bar re-opens this sheet on tap (session_detail_loaded_view.dart). Swipe/scrim dismissal is the deliberate defer affordance so the user can read the transcript behind the sheet before deciding; the explicit decline lives on the header X, which rejects. Making the sheet non-dismissible would remove that defer path without adding safety.

client/app/lib/features/session_detail/widgets/question_modal.dart
```diff
@@ -18,30 +19,46 @@ class QuestionModal extends StatefulWidget {
...
+      // transparent. The sheet caps itself below the status bar.
       backgroundColor: Colors.transparent,
+      useSafeArea: false,
       builder: (_) => QuestionModal(
         question: question,
         onReply: onReply,
         onReject: onReject,
+        topInset: topInset,
       ),
     );
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed: same defer design as the permission sheet — a dismissed question stays in pendingQuestions and the pinned question banner re-opens this sheet on tap (session_detail_loaded_view.dart), so the session cannot hang silently. Dismissibility is unchanged from main; forcing isDismissible: false would trap the user when they need transcript context to answer.

client/app/lib/features/project_list/project_list_screen.dart
```diff
@@ -79,10 +79,20 @@ class _ProjectListBodyState extends State<_ProjectListBody> {
     final scaffoldMessenger = ScaffoldMessenger.of(context);
     final cubit = context.read<ProjectListCubit>();
     final loc = context.loc;
+    // Same display-name resolution as _ProjectTile, so the sheet is titled
+    // by the project it acts on.
+    final lastSegment = project.id.split("/").last;
+    final displayName = project.name ?? (lastSegment.isNotEmpty ? lastSegment : loc.projectListDefaultName);
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed: this deliberately mirrors the identical resolution in _ProjectTile (project_tile.dart:21) so the sheet title always matches the tile it was opened from (the comment above the line notes this). Project ids are bridge-normalized directory paths with / separators (bridge hosts are macOS/Linux), and package:path's default context on mobile is already posix, so p.basename would be functionally identical here. Switching only this call site would risk divergence from the tile; unifying both sites behind a shared helper is a sensible follow-up outside this migration PR.

client/app/lib/features/session_detail/widgets/question_modal.dart
```diff
@@ -196,148 +213,130 @@ class _QuestionModalState extends State<QuestionModal> {
...
+    final bottomInset = keyboard > 0 ? keyboard : MediaQuery.paddingOf(context).bottom;
+    // Size to content: a short question set wraps; a tall one caps just under
+    // the sheet's own cap and scrolls inside the Flexible list while the
+    // actions stay pinned.
+    final maxBody = size.height - widget.topInset - PregoBottomSheet.contentTopInset - bottomInset;
+
+    return PregoBottomSheet(
+      title: info.header.isNotEmpty ? info.header : loc.questionModalTitle,
+      topInset: widget.topInset,
+      onClose: _dismissModal,
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed: the X is intentionally defer, matching the sheet's pre-existing swipe-to-dismiss semantics (on main this sheet had no header X and could already be swiped away unanswered). A dismissed question is not lost — it stays in pendingQuestions and the pinned question banner re-opens the sheet on tap (session_detail_loaded_view.dart). Wiring X to _onReject would make a visible close affordance destructive: it sends an explicit decline for the whole question set when the user may only want to check transcript context. The permission sheet's X=reject differs on purpose — denying one tool call is recoverable (the assistant can re-request), while rejecting a question consumes it.

client/app/lib/core/widgets/command_picker_sheet.dart
```diff
@@ -23,24 +23,23 @@ class CommandPickerSheet extends StatefulWidget {
...
-          borderRadius: const BorderRadius.vertical(top: Radius.circular(16)),
-          clipBehavior: Clip.antiAlias,
-          child: SizedBox(
-            height: height,
-            child: CommandPickerSheet(commands: commands),
-          ),
+        // The body hosts its own scroll view, so it needs a bounded height.
+        // Shrink above the keyboard (the sheet re-adds the keyboard inset
+        // below the body) so the search field stays visible while typing.
+        final size = MediaQuery.sizeOf(sheetContext);
```
daniil-shumko: Does using this impact the app performance? Seems that we are accessing the MediaQuery too often. Could there be a better way to solve this?

client/app/lib/core/widgets/command_picker_sheet.dart
```diff
@@ -110,126 +109,113 @@ class _CommandPickerSheetState extends State<CommandPickerSheet> {
...
-                        Padding(
-                          padding: const EdgeInsetsDirectional.only(top: 4),
-                          child: Text(
-                            hints,
+              final filtered => ListView.separated(
+                // Extend the scrollable underneath the home indicator: with
+                // handleBottomSafeArea: false the sheet no longer pads for it, so
+                // the bottom inset is added as scroll padding here (mirroring the
+                // model picker) instead of clipping the last command above it.
+                padding: EdgeInsetsDirectional.fromSTEB(8, 4, 8, 4 + MediaQuery.paddingOf(context).bottom),
```
daniil-shumko, line 169: Do you think it is wise to extract the `MediaQuery.paddingOf(context).bottom)` somewhere where we read it once? If the user did not change the orientation of the device, then I think this value will not be changing often. Only if the size of the app actually changes, could there be a future proof for this bottom padding/safe area app wide?

client/app/lib/core/widgets/model_picker_sheet.dart
```diff
@@ -53,47 +53,42 @@ class ModelPickerSheet extends StatefulWidget {
...
-                context.pop();
-              },
-            ),
+        final keyboard = MediaQuery.viewInsetsOf(sheetContext).bottom;
+        // The body hosts its own scroll view, so it needs a bounded height.
+        // The keyboard inset is subtracted (the sheet re-adds it below the
+        // body) so the search field stays visible while typing. Full screen:
+        // fill from just below the sheet header to the bottom edge.
+        final maxBody = size.height - topInset - PregoBottomSheet.contentTopInset - keyboard;
+        final height = fullScreen ? maxBody : math.min(size.height * 0.7 - keyboard, maxBody);
```
daniil-shumko: Does all this calculation needed here? Could there be a bug on different screen sizes? Especially small phones or large desktop

client/app/lib/features/session_detail/widgets/reasoning_modal.dart
```diff
@@ -72,7 +112,16 @@ class _ReasoningModalState extends State<ReasoningModal> {
 
     final prego = context.prego;
     final loc = context.loc;
-    final height = MediaQuery.of(context).size.height * 0.7;
+    final size = MediaQuery.sizeOf(context);
```
daniil-shumko: Could we not just use `MediaQuery.heightOf` if we only need height? Also, review all other usage of `MediaQuery.` maybe there are more optimized ways to use it. Only make sure you stay within the context and code changes related to this PR

daniil-shumko (commented):
@claude see my comments and questions

client/app/lib/core/widgets/command_picker_sheet.dart
daniil-shumko: [Sesori reply] Not addressed (no code change needed — answering)

No measurable impact. `MediaQuery.sizeOf`/`viewInsetsOf`/`paddingOf` are the aspect-scoped accessors Flutter added exactly for this: each is an O(1) inherited-model lookup (no platform query — the values are computed once per metrics change at the root view), and the caller registers a dependency on only that aspect, so these builders rebuild only when the height/insets actually change (keyboard animation, rotation, window resize) — precisely the moments the sheet must re-measure anyway. Flutter's own SafeArea/Scaffold internals read MediaQuery the same way on every build, and the builder only runs while the sheet is open. Per your comment on the reasoning modal, these call sites are now further narrowed to `MediaQuery.heightOf` in 522f5802.

client/app/lib/core/widgets/command_picker_sheet.dart
daniil-shumko: [Sesori reply] Not addressed

I'd advise against a read-once holder, for three reasons. (1) The value is context-scoped by design: modal routes strip the top inset, Scaffold consumes padding for its bars, and `padding.bottom` auto-zeroes while the keyboard covers the home-indicator region — behaviour this sheet relies on. A global value would be right in some scopes and wrong in others; the status-bar bug fixed earlier on this branch (607289ec) came exactly from reading insets in the wrong scope. (2) It effectively is read-once already: MediaQuery is the cache — `paddingOf` doesn't query the OS, it's a constant-time lookup on data recomputed only when window metrics change. (3) Staleness: rotation, iPad multitasking/Stage Manager, and desktop window resize all change it; MediaQuery invalidates dependents automatically, while a manual cache would need its own invalidation — a worse reimplementation of the same mechanism. If the goal is fewer call sites re-deriving this, the design-system lever would be PregoBottomSheet computing the insets once and passing resolved metrics (max body height + bottom scroll padding) into its builder — happy to do that as a follow-up.

client/app/lib/core/widgets/model_picker_sheet.dart
daniil-shumko: [Sesori reply] Not addressed (no code change needed — answering)

Each term is needed for a distinct constraint: `0.7 * height` is the sheet's default size (same as before the migration); subtracting `keyboard` keeps the search field visible while typing (the sheet re-adds that inset below the body); `min(…, maxBody)` stops the body exceeding the space under the sheet header, which would give the outer sheet scroll range on top of the inner list; `max(…, 0.3 * height)` keeps the sheet usable when the keyboard eats most of the screen. On small phones (SE-class, 667pt): keyboard up leaves a ~200pt list with the search field visible — checked the arithmetic across portrait sizes and the cap/floor never conflict there. On large desktop windows there's no keyboard, so it's simply 70% of the window capped under the header. One real inconsistency did exist — the command picker was missing the `maxBody` cap (cubic flagged it too) and now mirrors this code exactly (e6c6e654), with a regression test. If you'd rather not have this computed per sheet, the follow-up I'd suggest is PregoBottomSheet passing resolved metrics into its builder so the formula lives once in module_prego.

client/app/lib/features/session_detail/widgets/reasoning_modal.dart
daniil-shumko: [Sesori reply] Addressed (in commit 522f5802)

Swapped every `MediaQuery.sizeOf` that only read `.height` to `MediaQuery.heightOf` — all seven call sites across the migrated sheets (reasoning/question/permission modals, model/command pickers, add-project dialog) and `PregoBottomSheet` itself, so these builders no longer rebuild on width-only metric changes. Also audited the remaining `MediaQuery.` uses in the PR-touched files: everything else is `paddingOf`/`viewInsetsOf`, which are already the finest-grained accessors Flutter offers (only `size` has per-dimension variants; there is no per-edge accessor for padding/insets).

client/app/lib/core/widgets/command_picker_sheet.dart
```diff
@@ -23,24 +23,23 @@ class CommandPickerSheet extends StatefulWidget {
...
-          child: SizedBox(
-            height: height,
-            child: CommandPickerSheet(commands: commands),
-          ),
+        // The body hosts its own scroll view, so it needs a bounded height.
+        // Shrink above the keyboard (the sheet re-adds the keyboard inset
+        // below the body) so the search field stays visible while typing.
+        final size = MediaQuery.sizeOf(sheetContext);
+        final keyboard = MediaQuery.viewInsetsOf(sheetContext).bottom;
+        final height = math.max(size.height * 0.7 - keyboard, size.height * 0.3);
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit e6c6e654)

Captured the status-bar inset before presenting and capped the body at `min(0.7 * height - keyboard, maxBody)`, mirroring the model picker's formula, so the outer sheet never gains scroll range of its own on top of the inner list. Added a regression test that simulates a status bar tall enough for the cap to bite and pins the body height to `maxBody` (mutation-checked: reverting the cap fails it).

client/app/lib/features/session_detail/widgets/permission_modal.dart
```diff
@@ -39,13 +51,19 @@ class PermissionModal extends StatelessWidget {
     })
     onReply,
   }) {
-    return showAppModalBottomSheet<void>(
+    // Capture before presenting: inside the route the top inset reads as 0.
+    final topInset = MediaQuery.paddingOf(context).top;
+    return showModalBottomSheet<void>(
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed

Same assessment as the earlier thread on this line: scrim/swipe dismissal is pre-existing behavior (main presents this sheet with the same route defaults), and a dismissed request is not lost — it stays in pendingPermissions and a persistent SessionDetailPendingBanner pinned under the app bar re-opens this sheet on tap (session_detail_loaded_view.dart). Dismissal is the deliberate defer affordance so the user can consult the transcript before deciding; the header X is the explicit reject path. Auto-rejecting on any dismissal would destroy a recoverable request, and disabling dismissal would trap the user in the sheet.

client/app/lib/features/session_detail/widgets/question_modal.dart
```diff
@@ -196,148 +213,130 @@ class _QuestionModalState extends State<QuestionModal> {
...
+      // Full-bleed body; the step indicator, list, and actions pad themselves.
+      contentPadding: EdgeInsetsDirectional.zero,
+      child: ConstrainedBox(
+        constraints: BoxConstraints(maxHeight: math.max(maxBody, size.height * 0.3)),
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed

The 0.3-height floor is the shared pattern across the migrated sheets (model picker, reasoning modal, add-project dialog — the same reference this PR's other comments cite), and it exists so the sheet cannot collapse to a sliver under a large keyboard. It only overrides maxBody when keyboard > 0.7 * height − topInset − header, which is unreachable in portrait on any supported phone (worst case ~0.55 * height of headroom vs ~0.4 * height keyboards) and only marginal in landscape-with-keyboard. Even then the actions are not lost: the keyboard inset is the sheet's trailing sliver, so the outer scroll bottoms out with the action row sitting directly above the keyboard. Clamping to zero instead would squeeze the question content and text field into a few points of height in exactly those cases.

PR #398, opened 2026-07-08
fix(bridge): key moved-folder project id off the opened directory

AlexDochioiu:
## Problem

Follow-up to #396, which only surfaced moved/deleted folders in the list but didn't fix the underlying re-add regression the user reported:

1. Open a project at a folder on the machine.
2. Move that folder to another directory on disk.
3. Hide the project in the app.
4. Open a new project from the folder's new location.
5. Sesori still tries to use the folder's **original** location — the new location can't actually be added.

## Root cause

The bridge uses a project's directory path as its canonical project id; for the OpenCode plugin `PluginProject.id = OpenCode Project.worktree`.

OpenCode identifies a git project by **git identity** (remote URL / root commit / cached id), not by path. So opening a repo that was moved on disk from `/A` to `/B` resolves to the **same** project and returns the stale primary `worktree = /A` (now gone), while recording the freshly-opened `/B` in the project's `sandboxes`.

`ProjectRepository.openProject` (native branch) then keyed `unhideProject`, the `directoryMissing` check, and the returned `Project.id` off that stale `/A` — never off the `/B` path the user actually opened. Result: `/B` could never be added and the hidden `/A` project was silently unhidden.

## Fix

Resolve the quirk **inside the OpenCode plugin** (per "backend quirks live in the plugin"). When `getProject`/`renameProject` are asked about a directory that is not the project's `worktree` but **is** one of its `sandboxes`, `PluginModelMapper.mapProject` now uses that requested directory as the project id, so the bridge keys off the live location the user opened. The list flow (`getProjects`) passes no requested directory and is unchanged.

No changes to bridge `app/`, `shared/`, or the client — the plugin boundary is preserved.

## Tests

- New `plugin_model_mapper_test.dart` unit tests for the id resolution (worktree match, sandbox match, unknown-dir fallback, path-separator normalization).
- `opencode_plugin_impl_test.dart`: `/project/current` mock is now sandbox-aware; new `getProject` group asserts a moved folder keys off the new location.
- `dart analyze --fatal-infos` clean; full bridge suites pass (plugin + app).

## Review

- Plan reviewed by `aristotle-plan-review` — approved.
- Implementation reviewed by `aristotle-impl-review` — approved.

---
## Summary by cubic
Fixes the regression when reopening a Git project after its folder was moved: the project id now keys off the directory you opened, not the stale worktree, so the new location adds correctly and the old one stays hidden.

- **Bug Fixes**
  - Map `PluginProject.id` to the requested directory when it matches a sandbox and not the primary worktree.
  - `getProject` and `renameProject` pass the requested directory to ensure id stability for moved folders.
  - Normalize path separators when comparing directories.

<sup>Written for commit e5b6ff68ae106fcccc0bf6f9fe7b4ccef846e8d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
+++ b/bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
@@ -515,7 +515,10 @@ class OpenCodePlugin implements OpenCodeManagedApi {
         directory: projectId,
       ),
     );
-    return _pluginModelMapper.mapProject(project);
+    // projectId is the directory the caller asked to open. When a git repo was
+    // moved on disk, OpenCode still returns the stale primary worktree; pass the
+    // requested directory so the mapper can key off the live location instead.
+    return _pluginModelMapper.mapProject(project, requestedDirectory: projectId);
   }
 
   @override
@@ -559,7 +562,9 @@ class OpenCodePlugin implements OpenCodeManagedApi {
         body: {"name": name},
       ),
     );
-    return _pluginModelMapper.mapProject(updated);
+    // Keep the id stable with the directory the caller referenced, so a rename
+    // of a moved-folder project doesn't snap the id back to the stale worktree.
+    return _pluginModelMapper.mapProject(updated, requestedDirectory: projectId);
   }
 
   void _handleRawSseEvent(String rawData) {
--- a/bridge/sesori_plugin_opencode/lib/src/plugin_model_mapper.dart
+++ b/bridge/sesori_plugin_opencode/lib/src/plugin_model_mapper.dart
@@ -47,10 +47,23 @@ class PluginModelMapper {
     );
   }
 
-  PluginProject mapProject(Project project) {
+  /// Maps an OpenCode [Project] to the plugin-interface [PluginProject], whose
+  /// id is the project directory the bridge keys everything off.
+  ///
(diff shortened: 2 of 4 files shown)


AlexDochioiu:
Implementation is not a great fix. Will do a proper fix instead.

AlexDochioiu:
Closing without merging. This interim fix remapped the project id to the opened directory inside the OpenCode plugin, which only fixes the one-shot getProject: the list path still maps back to the stale canonical worktree (the P1 flagged here — the re-opened project drops off the next list refresh), and making the id follow the folder orphans durable per-project state (hidden/rename/base branch) on every move.

The proper fix separates identity from location instead: the plugin keeps reporting the canonical worktree as the stable id, and the bridge stores the live directory alongside it, resolving id → path for every backend/git operation. That work supersedes this branch.

PR #440, opened 2026-07-13
fix(client): reduce chat-screen battery drain during agent streaming

daniil-shumko:
## Problem

Battery drain / phone heating on the session chat screen while an agent streams. Profiled live on a physical iPhone (profile mode, VM-service CPU samples + timeline): during any agent turn the screen rendered at display refresh (~65fps) nonstop — the indeterminate Material spinners (nav-bar busy, tool cards, background tasks) each keep a vsync ticker hot for the entire turn. Compounding it:

- every streaming flush re-parsed the **full accumulated markdown document** (regex interpreter topped every CPU profile — quadratic over a long reply);
- the reasoning preview laid out the whole reasoning doc under a ShaderMask per flush to show ~3 lines;
- PR-sync `sessions.updated` bursts each triggered a full ~10-round-trip encrypted snapshot refetch;
- per-rebuild `TextPainter` overflow measurement and `ChatTheme` derivation on every visible row per flush.

## Fixes

- **`IsolatedActivityIndicator`** (new, `client/app` core widgets): the stock indeterminate spinner behind its own `RepaintBoundary`, confining per-frame repaints to the spinner's small layer instead of re-rasterizing the screen; honours OS reduce-motion (static arc). Swapped into all long-lived spinner sites on the chat screen.
- **Adaptive stream throttle**: `StreamingTextBuffer` stretches its flush interval from 50ms up to 300ms as buffered text grows, capping the quadratic markdown-re-parse tail.
- **Reasoning preview** renders only a ~700-char tail slice of the streaming text.
- **Silent-refresh coalescing**: event-driven refreshes (`sessions.updated`, `command.executed`, `dataMayBeStale`) go through a 5s leading-edge cooldown with a trailing catch-up refresh; reconnect/resume refreshes bypass it deliberately.
- **Memoized** tool-output overflow measurement and chat theme derivation.
- Xcode **Profile** build config now dev-signs (was pinned to Apple Distribution, breaking `flutter run --profile`); Release untouched.

## Measured result (same tool-heavy live scenario, on device)

| Metric | Before | After |
| --- | --- | --- |
| Sustained frame rate during agent turn | ~65fps | ~20fps (~8fps between content updates) |
| Main-isolate CPU mid-stream | ~14% of a core | ~10% of a core |

## Testing

- `client/module_core`: 484 tests green (new: adaptive flush interval, refresh-coalescing burst behaviour)
- `client/app`: 588 tests green (new: `ThrottledActivityIndicator` timer stepping / reduce-motion / unmount)
- `dart analyze` / `flutter analyze` clean
- Live E2E re-verified on iPhone via bridge + cursor plugin

Bridge-side follow-ups (SSE per-session scoping/coalescing, full-part snapshot payloads, main-isolate crypto offload) are documented and intentionally out of scope.

https://claude.ai/code/session_01Vpg35AQbS663Amn8jcpWmQ

---
## Summary by cubic
Cuts chat-screen battery drain during agent streaming by isolating spinner repaints, backing off stream flushes, tailing the reasoning preview, and coalescing background refreshes with safe retries that defer while disconnected.

- **Performance**
  - Replace indeterminate spinners with `IsolatedActivityIndicator` (behind a `RepaintBoundary`); respects `TickerMode`; reduced motion shows a static arc with loading‑spinner semantics.
  - Adaptive `StreamingTextBuffer` flush: 50 ms base up to 300 ms as text grows; `removePart()` cancels/reschedules pending flushes and cancels when empty.
  - Reasoning preview renders only a ~700‑char tail; never starts on a UTF‑16 low surrogate; aligns to first newline unless it would collapse the 56px preview.
  - Silent refresh coalescing: 5 s cooldown with a trailing catch‑up; staleness is consumed only on a successful snapshot; retain/chain signals during in‑flight refreshes and across pause/resume; hold while hidden and defer while disconnected until reconnect; if the window elapses mid‑refresh, run the trailing refresh on completion; reconnect/resume bypass.
  - Memoize tool‑output overflow measurement and chat theme derivation.
  - Measured on iPhone: ~65fps → ~20fps during agent turns; main‑isolate CPU ~14% → ~10%.

- **Refactors**
  - Dev‑sign Xcode Profile and align entitlements so `flutter run --profile` works on device; Release unchanged.
  - Fix tests to pass required `pluginId` and cover spinner semantics, tail slicing, buffer backoff, and stale‑refresh coalescing.

<sup>Written for commit d23eb469e541378528695530ef7bae5cdaf2ab1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


client/app/lib/core/widgets/throttled_activity_indicator.dart
```diff
@@ -0,0 +1,81 @@
...
+  Timer? _timer;
+  int _step = 0;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    if (context.isReducedMotion) {
+      _timer?.cancel();
+      _timer = null;
+    } else {
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

The timer-stepped indicator was replaced wholesale (at the maintainer's direction) with IsolatedActivityIndicator — the stock indeterminate spinner behind its own RepaintBoundary. There is no raw Timer anymore, so muting on covered routes comes from the framework's own TickerMode handling of the spinner's ticker.

client/app/lib/core/widgets/throttled_activity_indicator.dart
```diff
@@ -0,0 +1,81 @@
...
+      _timer ??= Timer.periodic(_tickInterval, (_) {
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

The raw Timer is gone: the widget was replaced (at the maintainer's direction) with IsolatedActivityIndicator, the stock indeterminate spinner isolated behind a RepaintBoundary, so TickerMode muting on covered routes applies via the framework's own ticker handling.

client/app/lib/features/session_detail/widgets/reasoning_part_card.dart
```diff
@@ -150,6 +150,22 @@ class _ReasoningPartCardState extends State<ReasoningPartCard> {
...
+  static const int _kStreamingTailChars = 700;
+
+  static String _streamingTail(String text) {
+    if (text.length <= _kStreamingTailChars) return text;
+    final slice = text.substring(text.length - _kStreamingTailChars);
+    // Start at a line boundary when one exists so the slice doesn't begin
+    // with a mid-line fragment that wraps differently from the real text.
+    final newline = slice.indexOf('\n');
+    return newline >= 0 && newline + 1 < slice.length ? slice.substring(newline + 1) : slice;
+  }
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

The cut point now nudges past an orphaned low surrogate before slicing (with a regression test pinning the exact output). I did not take the `characters`-based suggestion: `text.characters`/`takeLast` re-walk the entire accumulated document on every streaming flush, which is exactly the O(document) per-flush cost class this PR removes; the O(1) boundary check fixes the malformed-Unicode case without reintroducing it.

client/app/lib/features/session_detail/widgets/reasoning_part_card.dart
```diff
@@ -150,6 +150,22 @@ class _ReasoningPartCardState extends State<ReasoningPartCard> {
...
+  /// The streaming preview is bottom-aligned and clipped to 56px, so only the
+  /// last few lines are ever visible — but laying out the whole accumulated
+  /// reasoning document costs O(text) under a saveLayer on every streaming
+  /// flush. Hand the layout only a tail slice that comfortably overfills the
+  /// viewport at the preview's small text size.
+  static const int _kStreamingTailChars = 700;
+
+  static String _streamingTail(String text) {
+    if (text.length <= _kStreamingTailChars) return text;
+    final slice = text.substring(text.length - _kStreamingTailChars);
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

Took the 'checking for a surrogate boundary' variant: the cut point nudges past an orphaned low surrogate before slicing, with a regression test. The `characters` route was deliberately avoided since grapheme iteration re-walks the whole accumulated document per flush — the cost class this PR exists to remove.

client/app/lib/features/session_detail/widgets/tool_part_widget.dart
```diff
@@ -135,6 +136,38 @@ class _ToolOutputBlockState extends State<_ToolOutputBlock> {
...
+      textDirection: Directionality.of(context),
+      textScaler: textScaler,
+    )..layout(maxWidth: textWidth);
+    _measuredOutput = output;
+    _measuredWidth = textWidth;
+    _measuredScaler = textScaler;
+    _isExpandable = painter.didExceedMaxLines;
+    painter.dispose();
+    return _isExpandable;
+  }
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

The memo key now includes the text style (and text direction, which the sibling cubic review also flagged). In practice the style is pinned — it always derives as textXs with an explicit fontSize 11 — but the cache is now correct against any future styling change for free.

client/app/lib/features/session_detail/widgets/tool_part_widget.dart
```diff
@@ -135,6 +136,38 @@ class _ToolOutputBlockState extends State<_ToolOutputBlock> {
...
+  bool _isExpandable = false;
+
+  bool _measureIsExpandable({
+    required BuildContext context,
+    required String output,
+    required double textWidth,
+    required TextStyle monoStyle,
+  }) {
+    final textScaler = MediaQuery.textScalerOf(context);
+    if (output == _measuredOutput && textWidth == _measuredWidth && textScaler == _measuredScaler) {
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

Both `monoStyle` and `Directionality.of(context)` are now part of the cache key and the stored measurement inputs.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -162,6 +169,36 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
+  /// collapse into a single trailing refresh. Reconnect and app-resume
+  /// refreshes bypass this on purpose: they must run promptly and re-assert
+  /// the bridge-side view declaration.
+  void _requestEventDrivenRefresh() {
+    if (state is! SessionDetailLoaded) return;
+    if (_eventRefreshCooldown != null) {
+      _eventRefreshQueued = true;
+      return;
+    }
+    _silentRefresh();
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 9eb22518)

Valid catch. `_requestEventDrivenRefresh` now queues the signal behind a cooldown window when a refresh is already in flight (instead of letting `_silentRefresh`'s `??=` coalesce it into the stale run), and the cooldown callback retains the queue and re-arms when it still can't start. Added a regression test where a refresh outlives the cooldown window and the queued signal must still produce a trailing refresh.

AlexDochioiu (approved):

client/app/lib/features/session_detail/widgets/reasoning_part_card.dart
```diff
@@ -21,6 +21,32 @@ class ReasoningPartCard extends StatefulWidget {
...
+    // (emoji etc.): an orphaned low surrogate is malformed and renders as a
+    // replacement character. Dropping the split character entirely is
+    // invisible in a tail preview, and stays O(1) where grapheme-aware
+    // slicing would re-walk the whole document on every flush.
+    if (_isLowSurrogate(text.codeUnitAt(start))) start++;
+    final slice = text.substring(start);
+    // Start at a line boundary when one exists so the slice doesn't begin
+    // with a mid-line fragment that wraps differently from the real text.
+    final newline = slice.indexOf('\n');
+    return newline >= 0 && newline + 1 < slice.length ? slice.substring(newline + 1) : slice;
```
AlexDochioiu: **P2: A newline near the end can collapse the preview to almost nothing.** This drops everything through the first newline anywhere in the 700-code-unit window. For example, if that newline is followed by one newly streamed character, the 56 px preview renders only that character and appears mostly blank until the new paragraph grows. The bounded raw slice already overfills the viewport, so keeping all 700 units is safe; alternatively only align to a newline when enough text remains to fill the preview. Please add a near-end-newline case rather than only the current 100-character suffix case.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -162,6 +169,54 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
+    _eventRefreshCooldown = null;
+    if (!_eventRefreshQueued) return;
+    if (isClosed || state is! SessionDetailLoaded) {
+      _eventRefreshQueued = false;
+      return;
+    }
+    if (_activeRefresh != null) {
+      // The previous refresh is still in flight: keep the signal queued and
+      // check again after another window so it is never silently dropped.
+      _eventRefreshCooldown = Timer(eventRefreshMinInterval, _onEventRefreshCooldownElapsed);
```
AlexDochioiu: **P2: A slow refresh can push the trailing update to almost two cooldown windows.** If refresh A starts at t=0, a newer signal queues, the timer observes A still active at t=5, and A completes at t=5.1, this re-arms a full five-second timer and the catch-up snapshot does not start until t=10. On a slow relay the screen can remain stale for another five seconds even though the minimum interval already elapsed. Please trigger the queued run when the active future completes after the deadline, or schedule only the remaining delay relative to the previous refresh start.

client/module_core/lib/src/cubits/session_detail/streaming_text_buffer.dart
```diff
@@ -45,9 +45,30 @@ class StreamingTextBuffer {
   }
 
   void _scheduleFlush() {
-    _timer ??= Timer(_throttle, () {
+    _timer ??= Timer(_flushInterval(), () {
```
AlexDochioiu, line 60: **P2: A finalized long part leaves its 300 ms timer throttling the next part.** `removePart()` removes the buffer but neither cancels nor reschedules `_timer`. If a large part schedules the capped interval, finalizes, and the next part starts before that timer fires, `_timer ??=` prevents the new part from getting its normal 50 ms flush, so its first visible delta can be delayed by up to 300 ms. With no next part, the old timer still causes a redundant empty snapshot emit. Please cancel/reschedule the timer when the buffered set changes and cover `large p1 -> remove p1 -> small p2` with fake time.

client/app/lib/core/widgets/isolated_activity_indicator.dart
```diff
@@ -0,0 +1,42 @@
...
+    super.key,
+    required this.strokeWidth,
+    required this.color,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return RepaintBoundary(
+      child: CircularProgressIndicator(
+        value: context.isReducedMotion ? _staticArcSweep : null,
```
AlexDochioiu: **P2: Reduced-motion mode exposes a fake 75% completion value to assistive technology.** Flutter treats any non-null progress value as determinate, changes the semantics role from `loadingSpinner` to `progressBar`, and derives the value `75`. These agent/tool/transcription operations have no measurable completion percentage. Please exclude the determinate arc semantics and wrap the static visual in loading-spinner semantics, with a semantics regression assertion.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -162,6 +169,54 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
+      // A refresh is already in flight (e.g. the reconnect path): its
+      // snapshot may predate this signal, so queue a trailing refresh behind
+      // a cooldown window instead of letting _silentRefresh coalesce the
+      // signal into the stale in-flight run.
+      _eventRefreshQueued = true;
+      _eventRefreshCooldown = Timer(eventRefreshMinInterval, _onEventRefreshCooldownElapsed);
+      return;
+    }
+    _silentRefresh();
+    _eventRefreshCooldown = Timer(eventRefreshMinInterval, _onEventRefreshCooldownElapsed);
```
AlexDochioiu, line 256: **P2: The new cooldown can spend radio/CPU while hidden and can survive a newer resume refresh.** The timer is only cancelled by `close()`, not on `paused`/`hidden`. A queued signal can therefore start the full roughly 10-request snapshot while the app is backgrounding. Also, if resume/reconnect runs a newer bypass refresh before this timer expires, `_eventRefreshQueued` remains set and this timer performs another obsolete full snapshot afterward. Please defer/cancel the cooldown across hidden states and consume signals older than a successful bypass refresh while preserving signals that arrive during that refresh.

client/app/lib/core/widgets/isolated_activity_indicator.dart
```diff
@@ -0,0 +1,42 @@
...
+  const IsolatedActivityIndicator({
+    super.key,
+    required this.strokeWidth,
+    required this.color,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return RepaintBoundary(
+      child: CircularProgressIndicator(
```
AlexDochioiu: **P1: This still schedules frames at display refresh for the whole agent turn.** `RepaintBoundary` narrows the dirty raster region, but an indeterminate `CircularProgressIndicator` still owns a repeating `AnimationController`, ticks every vsync, paints its layer, and causes the frame to be composited/presented. The nav-bar instance alone is visible for the full turn, so the final implementation cannot retain the earlier timer-stepped 65 fps to 20 fps result; that implementation was replaced after profiling. Please profile this exact head and use either a lifecycle/TickerMode-aware low-frequency painter or a static busy indicator if the goal is to let the render pipeline sleep between content updates.

client/app/lib/core/widgets/isolated_activity_indicator.dart
daniil-shumko: [Sesori reply] Addressed (in commit 9c67d033)

The reduced-motion arc now excludes the derived determinate semantics and wraps the visual in a `SemanticsRole.loadingSpinner` node, matching what the animated variant exposes. Regression test asserts the role and an empty value.

client/app/lib/features/session_detail/widgets/reasoning_part_card.dart
daniil-shumko: [Sesori reply] Addressed (in commit 9c67d033)

The tail now only aligns to a newline when at least half the window (350 units) remains after it; otherwise the whole slice is kept. Added the near-end-newline regression case (newline 10 chars from the end → full slice retained) and moved the existing suffix case above the threshold.

client/module_core/lib/src/cubits/session_detail/streaming_text_buffer.dart
daniil-shumko: [Sesori reply] Addressed (in commit 9c67d033)

`removePart()` now cancels the pending flush and reschedules with a recomputed interval when parts remain — so a finalized large part can't leave its 300ms timer throttling the next part, and removing the last part emits no empty snapshot (both call sites emit their own state on finalize). Covered `large p1 → remove p1 → small p2` plus the cancel and reschedule cases with fake time.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
daniil-shumko: [Sesori reply] Addressed (in commit 9c67d033)

Paused/hidden now cancels the cooldown timer (and staleness signals arriving while backgrounded queue without starting a refresh), so no snapshot runs while hidden. Consumption is anchored at refresh start: any refresh that begins — including the resume/reconnect bypass — clears the queued flag, since its snapshot covers every signal seen so far, while signals arriving during that refresh re-queue behind it. Tests cover held-while-hidden, consumed-by-resume, and no duplicate afterwards.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
daniil-shumko: [Sesori reply] Addressed (in commit 9c67d033)

Went with your first option: when the cooldown elapses while a refresh is still in flight, the elapsed handler chains onto the active future and runs the queued refresh as soon as it completes (guarded against a newer signal having armed its own cooldown in the meantime). Test pins the promptness: slow refresh completes → trailing refresh starts within milliseconds, not another window.

client/app/lib/core/widgets/isolated_activity_indicator.dart
daniil-shumko: [Sesori reply] Question

You're technically right: RepaintBoundary bounds the raster region, but the indeterminate spinner still owns a repeating AnimationController, so frames are scheduled/composited at display refresh for the whole turn. The history matters here though — the first implementation of this PR was exactly the low-frequency painter you describe (a timer-stepped arc at 8Hz, TickerMode/reduce-motion aware, profiled on device at 65fps → ~20fps), and it was replaced with the RepaintBoundary wrapper at the maintainer's explicit direction. So this thread is a product/design call between the two of you rather than something I should silently flip back: smooth stock Material motion (with frames still scheduled but cheap raster) vs. letting the pipeline sleep between content updates (~20fps measured, visibly stepped arc). Happy to re-profile this exact head on device and/or reinstate the low-frequency painter if that's the decision.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -804,6 +879,10 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
       case LifecycleState.paused:
       case LifecycleState.hidden:
         _wasPaused = true;
+        // Don't let a queued trailing refresh spend radio/CPU while hidden;
+        // the queue is preserved and consumed by the resume bypass refresh.
+        _eventRefreshCooldown?.cancel();
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 8aeb72ea)

Real gap — pause cancelled the only armed trailing trigger while the resume bypass returned early on the in-flight refresh, stranding the queue. The bypass path now chains the queue drain onto the in-flight refresh's completion whenever a signal is queued with no cooldown armed (same guarded continuation the elapsed handler uses, so a newer signal's cooldown still owns the queue). Regression test covers the exact race: signal queued behind an in-flight refresh + pause/resume → trailing refresh still fires on completion.

client/app/lib/core/widgets/isolated_activity_indicator.dart
AlexDochioiu: [Sesori reply] Follow-up required

The current head still schedules/composites every vsync for the full agent turn; `RepaintBoundary` only limits raster scope. The quoted 65 fps to ~20 fps measurement belongs to the removed timer-stepped implementation, so it is not evidence for `8aeb72ea`. Please profile this exact head and show that compositor/GPU/frame activity is acceptable, or restore a `TickerMode`-aware low-frequency/static busy indicator. The earlier visual-direction choice does not by itself satisfy the battery acceptance criterion.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
AlexDochioiu: [Sesori reply] Follow-up required

The success path is fixed, but queued staleness is cleared when the resume bypass refresh starts, not when it succeeds. If that refresh returns `SessionDetailLoadResultFailed`, the failure branch neither restores the queue nor schedules another attempt, so a signal held while hidden can still be lost indefinitely. Please add the held-signal + failed-resume regression case and preserve/re-arm the queue until a successful snapshot consumes it.

client/app/lib/core/widgets/isolated_activity_indicator.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 52d40eb2)

Replaced the stock indeterminate spinner with an 8 Hz determinate arc behind its own repaint boundary. Its timer stops when `TickerMode` is disabled or reduced motion is requested, so this head no longer schedules spinner frames at display refresh; focused tests cover the 125 ms cadence, suspension/resumption, static reduced-motion behavior, and loading semantics.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
AlexDochioiu: [Sesori reply] Addressed (in commit 52d40eb2)

A queued signal is now removed from the live queue only while its refresh runs and is permanently consumed only when the snapshot is successfully applied. Failure or connection blocking restores and re-arms that signal, while signals arriving during the refresh remain independently queued; the new hidden-signal + failed-resume regression test verifies the retry succeeds without an extra refresh.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -159,18 +166,125 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
+    // Remove prior staleness from the live queue while this refresh runs.
+    // Success consumes it; failure restores it. Signals that arrive while the
+    // refresh is in flight independently re-queue behind it.
+    final consumedQueuedSignal = _eventRefreshQueued;
+    _eventRefreshQueued = false;
+    late final Future<void> refresh;
+    refresh = _doSilentRefresh()
+        .then((appliedSnapshot) {
+          if (!appliedSnapshot && consumedQueuedSignal) {
+            _restoreQueuedRefreshAfterFailure();
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit bae0ec0c)

The leading event-driven signal is now entered into the existing queue before its refresh starts. A successful snapshot consumes it; failure restores it and the existing cooldown retries. The new regression test verifies first-load failure, one successful retry, and no third refresh.

client/app/lib/features/session_detail/widgets/prompt_input.dart
```diff
@@ -535,7 +536,7 @@ class _TranscribingIndicator extends StatelessWidget {
           SizedBox(
             width: 16,
             height: 16,
-            child: CircularProgressIndicator(
+            child: IsolatedActivityIndicator(
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Not addressed

The animated branch uses Flutter's stock indeterminate `CircularProgressIndicator`, whose semantics node already exposes `SemanticsRole.loadingSpinner`. The normal-motion test in `isolated_activity_indicator_test.dart` explicitly enables semantics and asserts that role (lines 18-39), and it passes in the current 10/10 CI run. Adding a second wrapper would duplicate semantics rather than restore a missing role.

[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit a36da288)

Wrapped both animated and reduced-motion indicator modes in explicit loading-spinner semantics and added a normal-motion semantics regression assertion.

client/app/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -396,7 +396,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = Sesori;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/RunnerRelease.entitlements;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "Apple Distribution";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "Apple Development";
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 67242f35)

Profile now uses the development APNs entitlement file alongside Apple Development signing; Release remains paired with production entitlements and Apple Distribution.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -159,18 +166,126 @@ class SessionDetailCubit extends Cubit<SessionDetailState> {
...
+    // While backgrounded the queue is held for the resume bypass refresh.
+    if (_wasPaused) return;
+    final active = _activeRefresh;
+    if (active != null) {
+      // The minimum interval has already elapsed, so run the queued refresh
+      // as soon as the in-flight one completes instead of waiting another
+      // full window.
+      _drainQueueWhenRefreshCompletes(active);
+      return;
+    }
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit d23eb469)

The cooldown drain now preserves queued staleness while disconnected and lets the existing reconnect path perform the refresh. Added an in-flight-disconnect regression test that verifies no repeated offline refresh state and one refresh after reconnect.

PR #519, opened 2026-07-21
fix(client): commit email autofill before navigation

AlexDochioiu:
## Summary

- defer Projects navigation while the email login sheet completes its success path
- let the sheet commit the accepted credentials to the autofill context before it pops
- replace the placeholder login screen test with a widget regression test that verifies the sheet pops before navigation

## Bug

`LoginCubit.loginWithEmail` emits `LoginSuccess` before returning `true`. The login screen listener immediately replaced the login route, disposing the email sheet with `AutofillContextAction.cancel`. The sheet therefore never reached its explicit `TextInput.finishAutofillContext()` call, so password managers were not offered the successfully authenticated credentials.

## Verification

- `flutter test test/features/login/login_screen_test.dart test/features/login/email_login_sheet_test.dart`
- `dart analyze lib/features/login/login_screen.dart test/features/login/login_screen_test.dart`

---
## Summary by cubic
Ensure the email login sheet commits autofill before navigation so password managers can save credentials. Navigation to Projects now occurs after the sheet closes.

- **Bug Fixes**
  - Defer navigation to `/projects` until the email sheet closes and calls `TextInput.finishAutofillContext()`, gating the `LoginSuccess` listener while `_isEmailSheetOpen`.
  - Add a widget regression test that verifies the sheet pops before navigating to Projects.

<sup>Written for commit ef555b2970699a2c5e90d980439b37bd56a30793. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/519?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/app/lib/features/login/login_screen.dart
+++ b/client/app/lib/features/login/login_screen.dart
@@ -135,6 +135,10 @@ class _LoginScreenBodyState extends State<_LoginScreenBody> {
     } finally {
       if (mounted) setState(() => _isEmailSheetOpen = false);
     }
+    if (!mounted) return;
+    if (context.read<LoginCubit>().state is LoginSuccess) {
+      context.goRoute(const AppRoute.projects());
+    }
   }
 
   @override
@@ -150,6 +154,9 @@ class _LoginScreenBodyState extends State<_LoginScreenBody> {
             current is LoginSuccess || current is LoginFailed || current is LoginTimeout || current is LoginIdle,
         listener: (context, state) {
           if (state is LoginSuccess) {
+            // The email sheet must commit its autofill context before this
+            // route is replaced. It navigates after the sheet has closed.
+            if (_isEmailSheetOpen) return;
             // Relay connection is handled reactively: AuthManager emits
             // AuthState.authenticated → ConnectionService connects. The
             // connection overlay shows progress; navigation proceeds immediately.
--- a/client/app/test/features/login/login_screen_test.dart
+++ b/client/app/test/features/login/login_screen_test.dart
@@ -1,13 +1,30 @@
+import "package:flutter/material.dart";
 import "package:flutter_test/flutter_test.dart";
 import "package:get_it/get_it.dart";
+import "package:go_router/go_router.dart";
 import "package:mocktail/mocktail.dart";
 import "package:rxdart/rxdart.dart";
 import "package:sesori_auth/sesori_auth.dart";
 import "package:sesori_dart_core/sesori_dart_core.dart";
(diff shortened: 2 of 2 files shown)


daniil-shumko:
I tried to verify this end-to-end and could not reproduce the underlying bug — the credential-save prompt appears **with and without** this change.

**Live A/B on an iPhone 17e simulator (iOS 26, Flutter 3.44.7), email login with the demo review account:**

- **This branch:** sign in via the email sheet → iOS shows the "Save Password?" prompt (over the Projects screen). ✓
- **main:** reverted only `login_screen.dart` to main's version (hot restart, same build otherwise), logged out, repeated the identical flow → the **same "Save Password?" prompt appears**. So password managers are offered the credentials even without the gate.

**The regression test doesn't discriminate either:** running this branch's `login_screen_test.dart` against main's `login_screen.dart` passes, including the `poppedPopupRoute` assertion. As written it would not catch a regression of the behavior this PR targets.

**Why the described race doesn't materialize:** `LoginCubit.loginWithEmail` does emit `LoginSuccess` before returning, and the screen's listener does call `context.goRoute(...)` first — but `go` only schedules the navigator rebuild for the next frame. The sheet's `_submit` continuation resumes in a microtask *before* that frame, so it always reaches `TextInput.finishAutofillContext()` while the `AutofillGroup` is still alive; the dispose-with-`cancel` runs a frame later, after the context has already been committed.

Given that, the `_isEmailSheetOpen` gate plus the post-sheet navigation in `_showEmailLogin` add a second navigation path for a timing window that doesn't seem to exist in practice. If there's a repro where the prompt genuinely doesn't appear (a specific device/OS — Android?), could you share it? Happy to re-verify against it. Otherwise I'd drop the `lib/` change; the test-file modernization (real widget test instead of the placeholder) could stand on its own, but it would need an assertion that main doesn't already satisfy.

AlexDochioiu:
[Sesori reply] Thanks for the live A/B and timing analysis. You're right: the route replacement is applied on the next frame, so the sheet's continuation commits autofill first, and the test does not discriminate against main. The production gate therefore adds complexity without fixing a demonstrated bug. I'm closing this PR rather than preserving an invalid change.

PR #639, opened 2026-07-31
[output-image-support] refactor(codex): seal rollout content [step 2/13]

AlexDochioiu:
## Summary

- replace the flattened rollout content type plus nullable text field with generated sealed input-text, output-text, summary-text, input-image, and unknown variants
- require each known variant's valid data and retain forward-compatible unknown fallback plus per-item malformed-content recovery
- use generated union serialization in the list converter, including preservation of unknown discriminators
- parse persisted `input_image.image_url` without exposing images before the later bounded image-mapping step
- move `CodexRolloutToolMapper` into the repository mapping layer
- inject one mapper instance from the `CodexPlugin` composition root into both live rollout enrichment and history replay
- preserve existing message, reasoning, tool-output, clipping, status, and malformed-record behavior

## Series

- Step 2/13 of `output-image-support`
- depends on merged plan PR #638
- 680 changed lines including generated output and tests, below the 1,500-line soft cap; no neighboring scope was combined

## Verification

- `dart pub get`
- focused rollout, event-mapper, session-service, and metadata tests
- full `dart test` in `bridge/sesori_plugin_codex` — 210 tests passed
- review follow-up: `dart test test/codex_rollout_api_test.dart` — 33 tests passed
- `dart analyze --fatal-infos` in `bridge/sesori_plugin_codex`
- `git diff --check origin/main...HEAD`
- `aristotle-impl-review` — approved with no findings

### DIFF
--- a/bridge/sesori_plugin_codex/lib/src/api/models/codex_rollout_dto.dart
+++ b/bridge/sesori_plugin_codex/lib/src/api/models/codex_rollout_dto.dart
@@ -42,16 +42,6 @@ enum CodexRolloutRole {
   unknown,
 }
 
-enum CodexRolloutContentType {
-  @JsonValue("input_text")
-  inputText,
-  @JsonValue("output_text")
-  outputText,
-  @JsonValue("summary_text")
-  summaryText,
-  unknown,
-}
-
 @Freezed(fromJson: true, toJson: false)
 sealed class CodexSessionIndexEntryDto with _$CodexSessionIndexEntryDto {
   const factory CodexSessionIndexEntryDto({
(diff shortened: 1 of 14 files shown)


bridge/sesori_plugin_codex/lib/src/api/models/codex_rollout_dto.dart
```diff
@@ -142,15 +154,24 @@ class CodexRolloutContentListConverter implements JsonConverter<List<CodexRollou
...
+            "type": "summary_text",
+            "text": text,
+          },
+          CodexRolloutInputImageDto(:final imageUrl) => {
+            "type": "input_image",
+            "image_url": imageUrl,
           },
-          "text": content.text,
+          CodexRolloutUnknownContentDto() => {"type": "unknown"},
         },
```
AlexDochioiu: isn't just calling `content.toJson()` yielding exactly the same? (once you enable toJson generation)

AlexDochioiu: [Sesori reply] Addressed (in commit b75a1e45)

Yes. Enabled generated `toJson`, replaced the manual switch with `content.toJson()`, and added converter serialization coverage. The generated fallback also preserves the original unknown discriminator.

PR #696, opened 2026-08-02
🚧 [relay-request-concurrency] refactor(bridge): coordinate routed request shutdown [step 3/10]

AlexDochioiu:
## Complexity

Complex: this introduces one cross-transport lifecycle owner whose acceptance and drain barrier must protect shared routing collaborators across relay, debug HTTP, session teardown, and runtime composition.

## What

- Add a concrete `RoutedRequestDispatcher` around the sole `RequestRouter`.
- Register every accepted relay/debug route completion synchronously.
- Stop route acceptance idempotently during shutdown and return a typed 503 rejection afterward.
- Make session and debug drains await the same memoized route barrier.
- Keep debug HTTP lifecycle tracking and serial relay response handling with their transport owners.
- Remove direct router injection and the debug-to-session mutation-drain callback.

## Why

Relay and debug routes share repositories and services that session teardown currently owns. A debug request can still be routing while session teardown disposes those collaborators. The shared barrier closes that ownership gap before relay request completion becomes concurrent in later steps.

## Risk and test focus

Risk is high around accepting work after a drain snapshot, missing a transport's route, deadlocking repeated drains, returning the wrong shutdown response, or disposing route collaborators before debug and relay work settles. Tests cover synchronous acceptance closure, repeated begin/drain calls, mapped route failure, overlapping encrypted relay and debug routes, post-stop debug 503 behavior, release in either order, runtime identity wiring, and teardown ordering.

## Expected result

- **User-visible:** No ordinary behavior change. A request racing after shutdown begins receives 503 rather than entering shared routing work.
- **Persisted/database:** None.
- **Internal:** One dispatcher owns route acceptance and completion lifetime for both transports. Relay request execution remains serial.

## Verification

- 63 focused dispatcher, router, debug, relay, runtime, and shutdown tests passed.
- `dart analyze --fatal-infos` from `bridge/app` passed.
- `git diff --check` passed.
- `aristotle-impl-review` approved the Step 3 changes with no blocking findings.
- 680 changed lines, below the 1,500-line soft cap.

### DIFF
--- a/.plan/active/relay-request-concurrency/PLAN.md
+++ b/.plan/active/relay-request-concurrency/PLAN.md
@@ -6,12 +6,13 @@
 - **Status:** Active — Step 1 PR
   [#687](https://github.com/sesori-ai/sesori_apps_monorepo/pull/687) and
   post-merge correction PR
-  [#688](https://github.com/sesori-ai/sesori_apps_monorepo/pull/688) merged;
-  Step 2 PR [#690](https://github.com/sesori-ai/sesori_apps_monorepo/pull/690) open
+  [#688](https://github.com/sesori-ai/sesori_apps_monorepo/pull/688) and Step 2 PR
+  [#690](https://github.com/sesori-ai/sesori_apps_monorepo/pull/690) merged;
+  Step 3 implementation ready for review
 - **Plan date:** 2026-08-02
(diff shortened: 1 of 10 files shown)


AlexDochioiu:
@kody start-review

PR #737, opened 2026-08-04
🌱 [claude-code-plugin] docs: plan Claude Code harness plugin [step 1/17]

daniil-shumko:
## Complexity

🌱 Trivial — documentation only. Three new plan files and one MCP config entry; no production code, no generated code, no tests.

## What

Raises the durable plan for adding **Claude Code** as a first-class Sesori harness, driven by the `claude` CLI's headless **stream-json** protocol over stdio.

- `.plan/active/claude-code-plugin/PLAN.md` — goal, success criteria, current-behavior evidence, locked scope, final architecture (package layout, component contracts, complete `BridgePluginApi` method mapping), two-wave registration, client touchpoints, cleanup assessment, delivery rules, the fixed 17-step sequence with per-step verification, the 20-check live matrix, compatibility, and risks with evidence levels.
- `.plan/active/claude-code-plugin/TRACKER.md` — fixed exact PR titles, execution rules, verification log, and the findings/deltas record.
- `.plan/active/claude-code-plugin/PROTOCOL.md` — skeleton ground-truth document. Every wire section is explicitly marked **UNVERIFIED** with its research hypothesis; Step 2 replaces each with an observed shape or a recorded absence.
- `.mcp.json` — registers `mobile-mcp` (the same server `opencode.json` already declares, translated to this file's schema) so the Step 16 simulator verification can run.

## Why

The README harness table still lists Claude Code as "Coming soon" — it is the flagship missing backend. The bridge already runs three plugins across the two shapes this needs (Cursor's port-less `SteadyPluginLifecycle`, Codex's global on-disk enumeration, ACP's stdio transport / approval registry / turn queue), and the client renders an unknown harness gracefully, so the groundwork exists.

The work is large enough to need a durable plan and a fixed PR series before implementation starts. The protocol skeleton exists separately because the stream-json control protocol is SDK-internal and unversioned: pinning its ground truth once, from direct observation, keeps thirteen later steps from each re-deriving wire shapes.

## Risk and test focus

**Risk: none.** No production code, no shared contracts, no generated output, no database or wire change.

- **User-visible behavior:** none.
- **Database:** no change.
- **Analytics:** none. Passive planning work with no authoritative user action.

Review focus is the plan itself: whether the step boundaries are independently valid and merge-ordered, whether the changed-line estimates are realistic against this repo's test-to-source ratio and committed Freezed output, and whether the architecture honors the layering, plugin-encapsulation, and no-speculative-machinery rules.

Two deliberate constraints worth checking:

- The plugin is registered only in **Step 14**, so no intermediate state ships a partial `BridgePluginApi`. But the workspace / Makefile / CI plumbing lands with the scaffold in **Step 2**, because all three lists are explicit — a package missing from them is locally unbuildable *and* invisible to CI.
- No new `MessagePartType` value is introduced anywhere in the series. That enum has no unknown fallback, so a new value would crash released clients.

## Expected result

`.plan/active/claude-code-plugin/` exists with three documents; `.mcp.json` declares `mobile-mcp` alongside `figma`. Nothing else changes and nothing builds differently.

## Verification

- `.mcp.json` parses (`python3 -m json.tool`).
- `git diff --check` passes.
- Every repository symbol cited in `PLAN.md` was verified to exist before being referenced: `SteadyPluginLifecycle`, `BridgeDerivedProjectsPluginApi`, `PersistedSessionCleanupApi`, `PluginSetupAuthenticationRequired`, `SemanticVersion`, `resolveUserHomeDirectory`, `maxToolOutputLength`, `maxInlineMessageAttachmentBytes`, `BufferedUntilFirstListener`, `PluginAuthenticationRequiredException`, `PluginStartAbortedException`, `HostProcessCommandExecutor`, `PluginValueOption`, `PluginQuestionInfo`.
- Diff is 997 changed lines. The initial 500-800 estimate was corrected to 900-1,100 in both `PLAN.md` and `TRACKER.md` rather than published stale.
- No Dart or Flutter suites were run: this is a documentation-only step.

https://claude.ai/code/session_01HuG5CqV5bHvpfS9aRcKMSN

---
## Summary by cubic
Introduces a durable plan for the Claude Code harness plugin using the `claude` CLI stream‑json protocol, with the protocol skeleton and a Step 1 tracker, and enforces a one‑open‑PR rule. Applies review feedback (fix delivery count, narrow scope guardrail, verify and make init `session_id` authoritative, add respawn‑durability checks), keeps the existing attachment‑gate decision, and updates Step 1 diff to ~1,300 lines; no production code.

- **Dependencies**
  - `.mcp.json`: add `mobile-mcp` via `npx -y @mobilenext/mobile-mcp@1.0.0` for the Step 16 simulator run.

<sup>Written for commit 529f7f2382005b73e3e6fbca5132687908185336. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -3,6 +3,10 @@
     "figma": {
       "type": "http",
       "url": "http://127.0.0.1:3845/mcp"
+    },
+    "mobile-mcp": {
+      "command": "npx",
+      "args": ["@mobilenext/mobile-mcp@latest"]
     }
   }
 }
--- a/.mcp.json
+++ b/.mcp.json
@@ -6,7 +6,7 @@
     },
     "mobile-mcp": {
       "command": "npx",
-      "args": ["@mobilenext/mobile-mcp@latest"]
+      "args": ["-y", "@mobilenext/mobile-mcp@1.0.0"]
     }
   }
 }
--- a/.plan/active/claude-code-plugin/PLAN.md
+++ b/.plan/active/claude-code-plugin/PLAN.md
@@ -1,0 +1,675 @@
+# Claude Code Harness Plugin
+
+## Status
+
+- **Plan slug:** `claude-code-plugin`
+- **Status:** Step 1/17 plan PR open
+- **Plan date:** 2026-08-04
+- **Plan delivery:** this document, `TRACKER.md`, and the `PROTOCOL.md` skeleton
+  are Step 1/17
+- **Implementation base:** `origin/main` at
+  `ca7470fd6ead8f7e1ff0d58e3591e7ce25a5314d`
+- **Repository:** `sesori-ai/sesori_apps_monorepo`
+- **Delivery:** one planning PR, thirteen sequential bridge PRs, one activation
+  PR, one client PR, one live-verification PR, and one plan-retirement PR
+
+## Goal
+
+Add Claude Code as a first-class Sesori harness by driving the `claude` CLI's
+headless **stream-json** protocol directly from a new bespoke bridge plugin, at
+full parity with the existing harnesses: sessions, live streaming, tools,
+permissions and questions, plan mode, model selection, interrupt, resume with
+history replay, slash commands, and image attachments.
+
+The plugin speaks the same ndjson stdio protocol the official Agent SDK speaks
+internally, so it needs no Node runtime — the `claude` CLI is a self-contained
+binary the user already has installed and logged in.
+
+## Success Criteria
+
+1. `GET /plugin` advertises a `claude` plugin with displayName `Claude Code`,
+   correct setup state, and a working brand mark in the mobile picker.
+2. A new Claude session created from the phone streams assistant text
+   token-by-token, renders thinking, and renders tool cards through their
+   pending → running → completed lifecycle.
+3. Tool permission requests reach the phone as permission cards; once, always,
+   and reject all round-trip correctly, and `always` never escalates beyond the
+   permissions the backend itself suggested.
+4. `AskUserQuestion` and `ExitPlanMode` render as question cards and their
+   answers reach the backend in the exact shapes the CLI expects.
+5. Model catalog and mid-session model switching work, and the client nav-bar
+   subtitle reflects the switch because every assistant message envelope carries
+   both `providerID` and `modelID`.
+6. Interrupt stops a running turn promptly and clears any pending permission or
+   question card; no stale prompt survives an abort, a session stop, or a
+   process exit.
+7. History survives a bridge restart: sessions enumerate from the on-disk
+   transcript index, open with a full replay whose shapes match live rendering,
+   and accept a follow-up turn through `--resume`.
+8. Sessions started in an external terminal appear in Sesori after a refresh,
+   under the project matching their working directory.
+9. No Claude-specific concept — tool name, model id, permission mode, transcript
+   path — escapes the plugin package into shared, bridge, or client code.
+10. No new `MessagePartType` value and no breaking wire change; the client
+    remains data-driven from `GET /plugin` apart from the brand asset and the
+    attachment gate.
+11. Every implementation PR is independently buildable and targets no more than
+    1,500 changed lines as a soft cap, counting additions plus deletions,
+    generated output, and tests against that PR's base.
+
+## Current Behavior And Evidence
+
+### There is no prior Claude work in this repository
+
+- Branch `claude-code-support` is 0 commits ahead of `main`; `.plan/` contains
+  nothing for Claude.
+- The single mention is `README.md:147`, the harness table row
+  `| [Claude Code](https://claude.com) | Coming soon | ... |`.
+- `shared/sesori_shared/lib/src/models/sesori/plugin_identity.dart` declares
+  `enum Harness { opencode, codex, cursor }`.
+- `bridge/app/lib/src/bridge/runtime/plugin_registry.dart` registers exactly
+  three descriptors.
+
+### The plugin seam is already sufficient
+
+Three shipped plugins cover the two shapes this work needs:
+
+- **Cursor** (`bridge/sesori_plugin_cursor/`) is the closest lifecycle model: a
+  CLI subprocess with no port and no ownership file, wrapped in
+  `SteadyPluginLifecycle` rather than `ManagedProcessService`, with
+  `projectOwnership = bridgeDerived` and `sessionOptionsScope = plugin`.
+  `cursor_plugin_descriptor.dart` is the template for probe-then-start.
+- **ACP** (`bridge/sesori_plugin_acp/`) owns the mechanics this plugin needs
+  verbatim: `AcpStdioClient` line framing and teardown, `AcpApprovalRegistry`
+  permission/question bifurcation, `AcpPlugin`'s per-session turn queue with
+  generation fencing, and `FakeAcpProcess` as the test seam pattern.
+- **Codex** (`bridge/sesori_plugin_codex/`) owns the enumeration model: a global
+  on-disk transcript index read through an injected-environment API, scanned in
+  `Isolate.run`, with `primeSessionDirectory` a documented no-op.
+
+### The client already renders an unknown harness gracefully
+
+Sessions, messages, tools, permissions, questions, models, and agents are all
+data-driven from `GET /plugin`. The only client work is the brand asset, its two
+lookup cases, and widening the temporary composer attachment gate.
+
+### Local runtime evidence
+
+The development machine has `claude` 2.1.221 on PATH. Step 2 pins the verified
+floor version and records the exact wire shapes observed at that version.
+
+## Protocol Reference
+
+The stream-json control protocol is SDK-internal and not formally versioned.
+`PROTOCOL.md` in this directory is the durable ground truth: Step 2 fills it from
+direct observation of the pinned CLI plus the SDK's own type declarations, and
+every later step cites it rather than re-deriving wire shapes.
+
+Condensed research carried into Step 2 (all marked verify-at-implementation in
+`PROTOCOL.md`):
+
+- Invocation: `claude --input-format stream-json --output-format stream-json
+  --verbose --include-partial-messages [--model <id>] [--permission-mode <mode>]
+  [--session-id <uuid> | --resume <uuid>]`, cwd = the session directory, one
+  process per session kept alive by streaming stdin.
+- Stdout types: `system` (`init`, `api_retry`, `permission_denied`),
+  `assistant`, `user`, `stream_event` (raw Anthropic SSE deltas), `result`,
+  `control_request`, `control_response`.
+- Permissions arrive as `can_use_tool` control requests answered with
+  `{"behavior":"allow"|"deny", ...}`; `AskUserQuestion` and `ExitPlanMode` are
+  ordinary tools whose approval is a user question rather than a permission.
+- Transcripts live at `$CLAUDE_CONFIG_DIR ?? ~/.claude` +
+  `/projects/<munged-cwd>/<session-id>.jsonl`; the filename is the session id
+  and each record carries its own `cwd` and timestamps, so the munged directory
+  name is never un-munged.
+- Auth is the user's existing `claude` login. The plugin never runs a login
+  flow and never overrides `HOME`, which would break macOS keychain lookup.
+
+## Locked Scope And Product Decisions
+
+### Included
+
+- A bespoke `bridge/sesori_plugin_claude/` package (pub name `claude_plugin`)
+  driving stream-json over stdio, one child process per live session.
+- Full `BridgePluginApi` surface plus `PersistedSessionCleanupApi`.
+- Global transcript enumeration, history replay, and `--resume` continuation.
+- Permission modes surfaced as two agents, **Default** and **Plan**.
+- Model catalog with mid-session `set_model` switching.
+- Text and image prompt parts; slash commands.
+- Brand assets and the two client lookups.
+
+### Excluded
+
+- The `claude-agent-acp` Node adapter route (decided against: adds a Node
+  dependency and hides half the protocol behind an adapter).
+- `--fork-session`, subagent sessions as first-class children (subagent work
+  renders inside its Task tool card), and MCP server management.
+- Any managed runtime, download, or version provisioning. The plugin resolves an
+  existing binary and reports setup state; it never installs one.
+- Any login flow, credential storage, or auth mutation.
+- Reasoning-effort variants unless Step 11 finds first-party support; the
+  decision is recorded there with evidence rather than assumed now.
+- New shared wire types, new `MessagePartType` values, relay changes, database
+  migrations, and product analytics. This harness adds no new authoritative user
+  action beyond the ones the existing harnesses already instrument.
+
+### Naming
+
+`Harness.claude` with id `claude`; displayName `Claude Code`; package directory
+`bridge/sesori_plugin_claude/`; pub name `claude_plugin`; plan slug
+`claude-code-plugin`.
+
+## Final Architecture
+
+### Lifecycle shape
+
+Cursor-style lifecycle, codex-style enumeration, per-session processes:
+
+- `SteadyPluginLifecycle`, not `ManagedProcessService` — there is no port to
+  reclaim and no ownership file to arbitrate.
+- `ensureRuntime` stays the inherited no-op. There is no `RuntimeManifest`
+  because there is no first-party download this plugin may legitimately perform.
+- The binary is `--claude-bin` (declared as the bare option name `bin`) else
+  `claude` on PATH, gated in `inspectSetup` by a typed `SemanticVersion` floor.
+- `start()` is cheap: construct, `markReady()`. Authentication failures surface
+  at use time as `PluginAuthenticationRequiredException`, which
+  `PluginRuntime.use()` already converts into a setup-state flip.
+- Per-session child processes spawn lazily on the first turn, are reaped after
+  an idle window, and respawn transparently through `--resume`.
+
+### Type
+
+`ClaudePlugin extends BridgeDerivedProjectsPluginApi implements
+PersistedSessionCleanupApi`. Claude has no project concept, so projects are
+bridge-derived from session working directories. `listAllSessions` ignores
+`knownDirectories` because the transcript index is global, and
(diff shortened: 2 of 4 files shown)


AlexDochioiu (commented):
Deep analysis of the plan (verified every cited repo symbol: `SteadyPluginLifecycle`, `BridgeDerivedProjectsPluginApi`, `PersistedSessionCleanupApi`, the five `PluginSetupStatus` variants, `SemanticVersion`, `resolveUserHomeDirectory`, `maxToolOutputLength`/`maxInlineMessageAttachmentBytes`, `BufferedUntilFirstListener`, the ACP stdio/approval/turn-state pattern, Codex's `Isolate.run` scan + `primeSessionDirectory` no-op + `_redactForLog`, the `allowMalformed` stderr decoder, the workspace/Makefile/CI wave-1 claim, the client brand-logo fallback and the both-non-null `providerID`/`modelID` subtitle requirement, and the no-unknown-fallback `MessagePartType` claim — all accurate). Six comments below: one factual miscount, one internal-contradiction, and three protocol-gap/robustness suggestions.

.plan/active/claude-code-plugin/TRACKER.md
```diff
@@ -0,0 +1,113 @@
...
+  changed-line estimates were cross-checked. Referenced repository symbols were
+  verified to exist before being cited: `SteadyPluginLifecycle`,
+  `BridgeDerivedProjectsPluginApi`, `PersistedSessionCleanupApi`,
+  `PluginSetupAuthenticationRequired`, `SemanticVersion`,
+  `resolveUserHomeDirectory`, `maxToolOutputLength`,
+  `maxInlineMessageAttachmentBytes`, `BufferedUntilFirstListener`,
+  `PluginAuthenticationRequiredException`, `PluginStartAbortedException`,
+  `HostProcessCommandExecutor`, `PluginValueOption`, and `PluginQuestionInfo`.
+  `.mcp.json` parses and `git diff --cached --check` passes. The staged diff is
+  997 changed lines; the original 500-800 estimate was corrected to 900-1,100
```
AlexDochioiu: Factual check: the recorded changed-line count doesn't match the diff. `git diff <merge-base> HEAD --numstat` sums to **1,011** (4 `.mcp.json` + 675 PLAN + 219 PROTOCOL + 113 TRACKER), and it was 1,011 at both commits on this branch. The '997 changed lines' figure appears here, in the PLAN Verification section, and in the PR body. The 900–1,100 band is fine either way, but since this step's whole purpose is verification discipline, the recorded number should be truthful — please re-measure or state what was measured (e.g. `--cached` against a partial staging).

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,675 @@
...
+
+- **Plan slug:** `claude-code-plugin`
+- **Status:** Step 1/17 plan PR open
+- **Plan date:** 2026-08-04
+- **Plan delivery:** this document, `TRACKER.md`, and the `PROTOCOL.md` skeleton
+  are Step 1/17
+- **Implementation base:** `origin/main` at
+  `ca7470fd6ead8f7e1ff0d58e3591e7ce25a5314d`
+- **Repository:** `sesori-ai/sesori_apps_monorepo`
+- **Delivery:** one planning PR, thirteen sequential bridge PRs, one activation
```
AlexDochioiu: Delivery-count math: 'one planning PR, **thirteen** sequential bridge PRs, **one** activation PR, one client PR, one live-verification PR, and one plan-retirement PR' sums to 18, but the series has exactly 17 steps. The thirteen bridge PRs are Steps 2–14, which already include the activation (Step 14 is 'Register The Harness'). The separate 'one activation PR' double-counts it — suggest 'twelve sequential bridge PRs plus one activation PR'.

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,675 @@
...
+   both `providerID` and `modelID`.
+6. Interrupt stops a running turn promptly and clears any pending permission or
+   question card; no stale prompt survives an abort, a session stop, or a
+   process exit.
+7. History survives a bridge restart: sessions enumerate from the on-disk
+   transcript index, open with a full replay whose shapes match live rendering,
+   and accept a follow-up turn through `--resume`.
+8. Sessions started in an external terminal appear in Sesori after a refresh,
+   under the project matching their working directory.
+9. No Claude-specific concept — tool name, model id, permission mode, transcript
```
AlexDochioiu: Success criterion 9 as worded contradicts what Steps 14 and 15 themselves do. Criterion: 'No Claude-specific concept … escapes the plugin package into shared, bridge, or client code.' But Step 14 adds `Harness.claude` to `shared/sesori_shared`, and Step 15 adds the Claude brand cases to client code plus the `claude` case in the client-side composer attachment gate. Presumably the criterion means no *backend-shaped* concept (tool names, model ids, permission modes, transcript paths) with identity/branding/gate as deliberate, listed exceptions — but as written it's an absolute the plan itself violates. Suggest narrowing the wording to 'no backend payload concept' or explicitly exempting the identity/branding/gate items.

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,675 @@
...
+  interrupted-not-errored classification for signalled exits.
+- Cover turn serialization, fencing, resume-versus-new spawn arguments, idle reap
+  followed by transparent resume, and approval cancellation on exit.
+- Verify: focused and full package tests, fatal analysis, implementation review.
+
+### Step 11/17 — Add The Model And Agent Catalog
+
+- Add `ClaudeCatalogService` owning models, agents, providers, and commands, plus
+  `set_model` and `set_permission_mode` application with an applied-selection
+  cache cleared on respawn.
```
AlexDochioiu: Respawn-state durability is the one behavioral area I'd add verification for. Steps 10/11 reap idle processes and respawn via `--resume`, and this step clears the applied-selection cache on respawn. Two things then need pinning down in Step 2 (or at least covered in Step 10/16):

1. Does `--resume` restore the session's last-used model? Claude Code persists the model per session; if the resumed process honors it while the plugin believes 'default', the plugin's applied-model bookkeeping drifts from reality — the nav-bar subtitle is server-stamped (both-non-null check is real, verified in `session_detail_cubit.dart:950`) so the UI would show one model while the next `set_model` decision is made against another.

2. Does an `always` permission grant survive a respawn? E2E-08 only exercises `always` inside one process lifetime. If grants are in-memory session state, users re-see permission cards after every idle window — a visible regression that E2E-18 (transparent resume) would not catch because it only checks the turn. Suggest either an E2E row ('after idle-reap, same-tool call proceeds without a prompt') or an explicitly recorded decision in Step 2.

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,675 @@
...
+binary level, and `markReady` on the next success.
+
+### BridgePluginApi method mapping
+
+| Method | Implementation |
+|---|---|
+| `getSessions(projectId)` | catalog scan filtered by normalized directory |
+| `getCommands` | slash commands from `system/init`; `PluginCommandSource.command` |
+| `getSessionOptions` | aggregate agents, providers, and commands; `refresh` re-probes |
+| `createSession` | pre-generate the session UUID, spawn with `--session-id`, dispatch the first prompt through the turn queue |
```
AlexDochioiu: `createSession`'s identity chain rests on an UNVERIFIED property: that `--session-id <uuid>` reliably pre-binds the transcript filename (PROTOCOL §5 flags exactly this). Everything downstream — enumeration, replay, `--resume`, delete — keys on that id, and a mismatch would silently split one session into two views. Cheap robustness: `system/init` already carries `session_id` (PROTOCOL §2), so make **init's reported id the source of truth**, cross-checked against the pre-generated UUID, and return that from `createSession`. Suggest recording that as the contract when Step 2 resolves the question rather than trusting the pre-bind.

.mcp.json
```diff
@@ -3,6 +3,10 @@
     "figma": {
       "type": "http",
       "url": "http://127.0.0.1:3845/mcp"
+    },
```
AlexDochioiu, line 6: Minor: `@mobilenext/mobile-mcp@latest` is unpinned. Every Claude Code session opened in this repo resolves the latest server at spawn time, which can drift out from under Step 16's verification (and from anyone else using the repo's `.mcp.json`). The plan pins the `claude` CLI floor for exactly this reproducibility reason — consider pinning the mobile-mcp version the same way.

.plan/active/claude-code-plugin/PLAN.md
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

You're right, the activation PR was double-counted. Now reads "twelve sequential bridge PRs, one activation PR, one client PR, one live-verification PR, and one plan-retirement PR — seventeen in total".

.plan/active/claude-code-plugin/PLAN.md
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Criterion 9 now scopes to backend *payload* concepts (tool name, model id, permission mode, transcript path, wire shape) and enumerates the three deliberate exceptions — `Harness.claude` in sesori_shared, the brand asset and display name in module_prego, and the harness id in the dated attachment gate — noting every existing harness carries the same three. It was an absolute the plan broke itself.

.plan/active/claude-code-plugin/PLAN.md
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Good catch, and I verified the property rather than leaving it unverified: a probe with a pre-generated UUID produced `<that uuid>.jsonl` and every record inside reported the same `sessionId`, so `--session-id` does pre-bind.

I still took your suggestion, because the identity chain is load-bearing. A new "Session identity" section makes init's reported `session_id` the source of truth, cross-checked against the pre-generated UUID, with a mismatch logged and the reported id winning since it's what the transcript is named after.

One wrinkle worth flagging: `system/init` turns out to be emitted when a *turn* starts, not when the process spawns (verified — a second init arrived with a second turn). So the cross-check lands on the first turn, which is exactly when `createSession` dispatches its first prompt. Fine here, but it also means nothing in the connect path can depend on init.

.plan/active/claude-code-plugin/TRACKER.md
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

You're right, and the number was wrong twice. The 997 came from `git diff --cached --numstat` against a partially staged tree. My first correction to 1,784 was also wrong — measured against a stale local `main` two commits behind the remote, which pulled in unrelated diff.

The tracker now records the authoritative command, `git diff $(git merge-base origin/main HEAD) --numstat`, and the figure derived from it, so it can be re-checked rather than trusted. Both earlier numbers are explicitly marked superseded.

.plan/active/claude-code-plugin/PLAN.md
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Both are now recorded as pre-Step-10 evidence items in a new "Respawn state durability" section, plus two E2E rows:

- E2E-18a — after an idle reap, the nav-bar model still matches what the plugin believes is applied.
- E2E-18b — after an idle reap, a tool previously granted `always` runs without prompting again.

Your point 1 also shaped the fix direction: if `--resume` does restore the persisted model, the process repository seeds applied-model from the first post-resume assistant message rather than assuming a default. Neither property is answered yet — they're logged as open questions rather than guessed at.

(Thread is marked outdated because the plan-review restructure moved that section; the substance still applies and is carried forward.)

.mcp.json
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Pinned to `@mobilenext/mobile-mcp@1.0.0` (current latest) and added `-y` so an unattended run can't block on install confirmation.

Also took the broader point other reviewers raised on this entry: since it exists only for the Step 16 simulator run and every ordinary Claude Code session in this repo starts the servers listed in `.mcp.json`, Step 17 now removes it as part of plan retirement.

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,860 @@
...
+    claude_permission_mode.dart   dual-spelled mode enum (CLI vs control)
+    claude_effort_level.dart      reasoning effort levels
+  lib/src/repositories/           LAYER 2 — normalization; no peer dependencies
+    claude_session_process_repository.dart  per-session client map; spawn vs
+                                            resume; per-session teardown; the
+                                            applied model/agent/mode of each
+                                            resident process; control-request
+                                            dispatch by session id
+    claude_transcript_catalog_repository.dart  transcript scan -> session records
+    claude_backend_catalog_repository.dart     initialize payload -> PluginModel,
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Valid, and a real inconsistency I introduced while applying the architecture review — I wrote the no-peer-dependency rule and then immediately broke it.

`ClaudeBackendCatalogRepository` is now collaborator-free: it's a pure mapping over an `initialize` payload handed to it. `ClaudeCatalogService` holds both repositories and passes the payload across, so the mapping stays in Layer 2 (where the architecture review required it) while the peer dependency disappears.

.plan/active/claude-code-plugin/PLAN.md
```diff
@@ -0,0 +1,675 @@
...
+direct observation of the pinned CLI plus the SDK's own type declarations, and
+every later step cites it rather than re-deriving wire shapes.
+
+Condensed research carried into Step 2 (all marked verify-at-implementation in
+`PROTOCOL.md`):
+
+- Invocation: `claude --input-format stream-json --output-format stream-json
+  --verbose --include-partial-messages [--model <id>] [--permission-mode <mode>]
+  [--session-id <uuid> | --resume <uuid>]`, cwd = the session directory, one
+  process per session kept alive by streaming stdin.
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Fair challenge, and I proved it rather than assuming. Two turns were driven over one process: it stayed alive between them (`poll()` returned None), both returned `result` frames, and it exited 0 only when stdin was closed. Multi-turn residency holds, so the architecture stays.

On `-p`: you're right that it's mandatory, and the plan's original invocation omitted it. That was already corrected during protocol verification — the launch spec now emits `-p` and has a test asserting it.

The same verification found something the docs don't mention: `--permission-prompt-tool stdio` is required or the CLI silently auto-denies every permission-gated tool. Recorded in PROTOCOL.md.

.mcp.json
```diff
@@ -3,6 +3,10 @@
...
+    "mobile-mcp": {
+      "command": "npx",
+      "args": ["@mobilenext/mobile-mcp@latest"]
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Pinned to `@mobilenext/mobile-mcp@1.0.0` rather than a range, matching how the plan pins the `claude` CLI floor for the same reproducibility reason.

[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Added `-y` and pinned the version: `["-y", "@mobilenext/mobile-mcp@1.0.0"]`.

.mcp.json
```diff
@@ -3,6 +3,10 @@
...
     }
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 529f7f23)

Added `-y` as suggested, and pinned to `@1.0.0` in the same change.

.mcp.json
```diff
@@ -3,6 +3,10 @@
     "figma": {
       "type": "http",
       "url": "http://127.0.0.1:3845/mcp"
+    },
+    "mobile-mcp": {
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Partially addressed (in commit 529f7f23)

You're right that a project-level entry loads for every non-bare session, not just Step 16. I've pinned the version and added `-y` so the per-session cost is bounded and can't hang.

Not moving it to an explicit `--mcp-config`, because the Step 16 verification is driven from an ordinary session in this repo — an external config file wouldn't be loaded by it. Instead Step 17 now removes the entry as part of plan retirement, so the cost is temporary and has a scheduled end rather than being permanent.

.mcp.json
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Partially addressed (in commit 529f7f23)

Same as the parallel cubic thread: pinned and `-y`-guarded, but kept in `.mcp.json` because the Step 16 run is driven from an ordinary session in this repo and would not pick up an external `--mcp-config`. Step 17 now removes the entry at plan retirement, so it does not outlive the verification it exists for.

.plan/active/claude-code-plugin/TRACKER.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Current State
+
+- **Plan slug:** `claude-code-plugin`
+- **Implementation base:** `origin/main` at
+  `ca7470fd6ead8f7e1ff0d58e3591e7ce25a5314d`
+- **Series state:** Step 1/17 in progress
+- **Current step:** 1/17 — raise the plan
+- **Plan PR:** pending
+- **Next action:** open the Step 1 PR, then begin Step 2 protocol ground truth
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Not addressed

No longer applicable. The tracker's "Current State" said the plan PR was pending while the steps table showed it open; that was fixed when the PR number was recorded, and the section now reflects the real state. The thread is also marked outdated against the current diff.

PR #755, opened 2026-08-04
⚙️ Map Cursor generate_image into inline Sesori images

AlexDochioiu:
## Complexity
⚙️ Moderate — Cursor extension request path + local image normalization into the existing ACP inline-image content pipeline.

## What
Handle Cursor's `cursor/generate_image` ACP extension as a JSON-RPC **request** (cursor-agent's `extMethod` / misnamed "notification"), read the local `filePath` on the bridge, and emit standard inline image message parts.

Also:
- Resolve omitted `sessionId` from the active turn / originating `toolCallId`
- Reuse shared assistant content append helper on `AcpEventMapper`
- Ack `cursor/update_todos` on the same request path (same wire bug)
- Keep a notification fallback for true `extNotification` agents

## Why
Sesori showed completed Generate Image tool cards but no preview because the bridge only listened for notifications. Live cursor-agent sends `cursor/generate_image` as a server request, so it was rejected as unhandled and the image never entered the message stream.

## Risk and test focus
- Live Cursor generate-image turns should show an inline preview in Sesori
- Confirm oversized images degrade to metadata (no transport blowup)
- No user-visible impact outside Cursor image generation; no database impact
- History reload still will not reconstruct past images unless cursor-agent replays the extension

## Expected result
Generating an image in a Cursor session streams a file part with bounded inline PNG/JPEG bytes to Sesori clients.

## Verification
- `dart test` in `bridge/sesori_plugin_cursor` for approval registry / event mapper / generate-image mapper
- Live confirmed: bridge emitted `BridgeSseMessagePartUpdated` with inline image for `sesori-test-yellow-star.png`

---
## Summary by cubic
Fixes live Cursor image previews by mapping `cursor/generate_image` into inline images and forwarding through the single notification pipeline with bounded reads, signature checks, encoded-size limits, and absolute-path enforcement. Also treats `cursor/update_todos` as fire-and-forget and resolves missing `sessionId` via explicit id, `toolCallId`, or the plugin’s active turn.

- **Bug Fixes**
  - Treat `cursor/generate_image` and `cursor/update_todos` as fire-and-forget: ack requests, forward as notifications; emit inline PNG/JPEG only with a valid image signature, otherwise degrade to metadata; enforce encoded-size limits; accept `filePath`/`path` keys but require absolute paths; accept optional `messageId`.

- **Refactors**
  - Added `CursorGeneratedImageReader` with single-descriptor bounded reads, signature-required inline, and encoded-size checks; shared raster policy and metadata builder via `AcpContentMapper`.
  - Centralized flow: `AcpApprovalRegistry.fireAndForgetExtensionMethods` + `AcpPlugin.handleAgentNotification` with contained forward on errors.
  - `AcpEventMapper.appendAssistantImageBlocks` appends image blocks to the live assistant message; `CursorEventMapper` handles both wire shapes and session resolution; `CursorApprovalRegistry` focuses on approvals only.

<sup>Written for commit 5a58422c90b56693e4255a0297082635a6036f6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
@@ -55,7 +55,7 @@ class AcpApprovalRegistry {
     required void Function(BridgeSseEvent event) emit,
     required AcpResponder respond,
     required AcpErrorResponder respondError,
-    void Function(AcpNotification notification)? onFireAndForgetNotification,
+    required void Function(AcpNotification notification) onFireAndForgetNotification,
     String Function()? idGenerator,
     String? Function()? activeSessionResolver,
   }) : _emit = emit,
@@ -69,7 +69,7 @@ class AcpApprovalRegistry {
   factory AcpApprovalRegistry.forClient({
     required AcpStdioClient client,
     required void Function(BridgeSseEvent event) emit,
-    void Function(AcpNotification notification)? onFireAndForgetNotification,
+    required void Function(AcpNotification notification) onFireAndForgetNotification,
     String Function()? idGenerator,
     String? Function()? activeSessionResolver,
   }) {
@@ -86,7 +86,7 @@ class AcpApprovalRegistry {
   final void Function(BridgeSseEvent event) _emit;
   final AcpResponder _respond;
   final AcpErrorResponder _respondError;
-  final void Function(AcpNotification notification)? _onFireAndForgetNotification;
+  final void Function(AcpNotification notification) _onFireAndForgetNotification;
   final String Function()? _injectedIdGenerator;
 
   /// Resolves the session a server request belongs to when the request itself
@@ -338,11 +338,13 @@ class AcpApprovalRegistry {
       // Fire-and-forget: the agent awaits no result. Ack so it never blocks,
       // then re-inject into the notification pipeline.
       _respond(request.id, const <String, Object?>{});
-      final forward = _onFireAndForgetNotification;
-      if (forward == null) {
-        Log.w("[acp] ${request.method} declared fire-and-forget with no notification forward wired; dropping");
-      } else {
-        forward(AcpNotification(method: request.method, params: request.params));
+      try {
+        _onFireAndForgetNotification(AcpNotification(method: request.method, params: request.params));
+      } on Object catch (error, stack) {
+        // The forward runs on the serverRequests subscription that also answers
+        // permissions/questions; a throw inside notification mapping must break
+        // only this payload, never the approval channel.
+        Log.w("[acp] ${request.method} notification forward failed", error, stack);
       }
       return;
     }
--- a/bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
@@ -55,11 +55,13 @@ class AcpApprovalRegistry {
     required void Function(BridgeSseEvent event) emit,
     required AcpResponder respond,
     required AcpErrorResponder respondError,
+    void Function(AcpNotification notification)? onFireAndForgetNotification,
     String Function()? idGenerator,
     String? Function()? activeSessionResolver,
   }) : _emit = emit,
        _respond = respond,
        _respondError = respondError,
+       _onFireAndForgetNotification = onFireAndForgetNotification,
        _injectedIdGenerator = idGenerator,
        _activeSessionResolver = activeSessionResolver;
 
@@ -67,6 +69,7 @@ class AcpApprovalRegistry {
   factory AcpApprovalRegistry.forClient({
     required AcpStdioClient client,
     required void Function(BridgeSseEvent event) emit,
+    void Function(AcpNotification notification)? onFireAndForgetNotification,
     String Function()? idGenerator,
     String? Function()? activeSessionResolver,
   }) {
@@ -74,6 +77,7 @@ class AcpApprovalRegistry {
       emit: emit,
       respond: (id, result) => client.respondToServerRequest(id: id, result: result),
       respondError: (id, code, message) => client.respondToServerRequestWithError(id: id, code: code, message: message),
+      onFireAndForgetNotification: onFireAndForgetNotification,
       idGenerator: idGenerator,
       activeSessionResolver: activeSessionResolver,
     );
@@ -82,6 +86,7 @@ class AcpApprovalRegistry {
   final void Function(BridgeSseEvent event) _emit;
   final AcpResponder _respond;
   final AcpErrorResponder _respondError;
+  final void Function(AcpNotification notification)? _onFireAndForgetNotification;
   final String Function()? _injectedIdGenerator;
 
   /// Resolves the session a server request belongs to when the request itself
@@ -145,6 +150,16 @@ class AcpApprovalRegistry {
   /// Override to handle harness-specific server requests (e.g. Cursor's
   /// `cursor/ask_question`). Return true if handled. Base handles none.
   bool handleExtensionRequest(AcpServerRequest request) => false;
+
+  /// Harness extension methods the agent sends as JSON-RPC *requests* (ACP
+  /// `extMethod`) but treats as fire-and-forget notifications (ACP
+  /// `extNotification`) — cursor-agent's `sendNonBlockingExtensionNotification`
+  /// does exactly this. The base handler acks them immediately (the agent
+  /// awaits no result, and an unanswered request would hang it) and re-injects
+  /// the payload into the plugin's notification pipeline, so one mapping path
+  /// serves both wire shapes. Do not also handle these in
+  /// [handleExtensionRequest]; this classification takes precedence.
+  Set<String> get fireAndForgetExtensionMethods => const {};
 
   // --- Lifecycle ---
 
@@ -317,6 +332,18 @@ class AcpApprovalRegistry {
   void _handle(AcpServerRequest request) {
     if (request.method == AcpMethods.sessionRequestPermission) {
       _handlePermission(request);
+      return;
+    }
+    if (fireAndForgetExtensionMethods.contains(request.method)) {
+      // Fire-and-forget: the agent awaits no result. Ack so it never blocks,
+      // then re-inject into the notification pipeline.
+      _respond(request.id, const <String, Object?>{});
+      final forward = _onFireAndForgetNotification;
+      if (forward == null) {
+        Log.w("[acp] ${request.method} declared fire-and-forget with no notification forward wired; dropping");
+      } else {
+        forward(AcpNotification(method: request.method, params: request.params));
+      }
       return;
     }
     if (handleExtensionRequest(request)) return;
--- a/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
@@ -432,32 +432,50 @@ class AcpEventMapper {
       content: update["content"],
       scope: tracker.mappingScope,
     );
-    return appendAssistantMappedBlocks(
+    return _appendAssistantBlocks(
       sessionId: sessionId,
-      identityUpdate: update,
+      identity: identity,
+      tracker: tracker,
       blocks: blocks,
       evaluateTextHaltNotice: true,
     );
   }
 
-  /// Appends pre-mapped assistant content into the same ordered message state
-  /// used by `agent_message_chunk`. Harness extensions that surface images
-  /// outside standard ACP chunks call this after local normalization.
-  List<BridgeSseEvent> appendAssistantMappedBlocks({
+  /// Appends assistant image blocks normalized locally on the bridge (a harness
+  /// extension that surfaces images outside standard ACP chunks) into the same
+  /// ordered message state used by `agent_message_chunk`.
+  List<BridgeSseEvent> appendAssistantImageBlocks({
     required String sessionId,
-    required Map<String, dynamic> identityUpdate,
+    required String? messageId,
     required Iterable<AcpMappedContentBlock> blocks,
-    required bool evaluateTextHaltNotice,
   }) {
     final identity = _chunkIdentity(
       sessionId: sessionId,
-      update: identityUpdate,
+      update: messageId == null
+          ? const <String, dynamic>{}
+          : <String, dynamic>{"messageId": messageId},
       role: _ChunkRole.assistant,
     );
     final tracker = (_contentTrackers[sessionId] ??= {}).putIfAbsent(
       identity.messageId,
       AcpContentTracker.new,
     );
+    return _appendAssistantBlocks(
+      sessionId: sessionId,
+      identity: identity,
+      tracker: tracker,
+      blocks: blocks,
+      evaluateTextHaltNotice: false,
+    );
+  }
+
+  List<BridgeSseEvent> _appendAssistantBlocks({
+    required String sessionId,
+    required ({String messageId, bool hasAcpMessageId}) identity,
+    required AcpContentTracker tracker,
+    required Iterable<AcpMappedContentBlock> blocks,
+    required bool evaluateTextHaltNotice,
+  }) {
     if (blocks.isEmpty) return const [];
     final hasTrackableContent = blocks.any(
       (block) => block is AcpMappedImageContentBlock || (block is AcpMappedTextContentBlock && block.text.isNotEmpty),
--- a/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
@@ -437,17 +437,18 @@ class AcpEventMapper {
       identity: identity,
       tracker: tracker,
       blocks: blocks,
-      evaluateTextHaltNotice: true,
     );
   }
 
   /// Appends assistant image blocks normalized locally on the bridge (a harness
   /// extension that surfaces images outside standard ACP chunks) into the same
-  /// ordered message state used by `agent_message_chunk`.
+  /// ordered message state used by `agent_message_chunk`. The image-typed
+  /// parameter also guarantees structurally that this path can never trip the
+  /// text-only halt-notice classification in [_appendAssistantBlocks].
   List<BridgeSseEvent> appendAssistantImageBlocks({
     required String sessionId,
     required String? messageId,
-    required Iterable<AcpMappedContentBlock> blocks,
+    required Iterable<AcpMappedImageContentBlock> blocks,
   }) {
     final identity = _chunkIdentity(
       sessionId: sessionId,
@@ -465,7 +466,6 @@ class AcpEventMapper {
       identity: identity,
       tracker: tracker,
       blocks: blocks,
-      evaluateTextHaltNotice: false,
     );
   }
 
@@ -474,7 +474,6 @@ class AcpEventMapper {
     required ({String messageId, bool hasAcpMessageId}) identity,
     required AcpContentTracker tracker,
     required Iterable<AcpMappedContentBlock> blocks,
-    required bool evaluateTextHaltNotice,
   }) {
     if (blocks.isEmpty) return const [];
     final hasTrackableContent = blocks.any(
@@ -484,8 +483,7 @@ class AcpEventMapper {
       _closeCurrentIdlessAssistantContent(sessionId: sessionId);
     }
 
-    if (evaluateTextHaltNotice &&
-        !identity.hasAcpMessageId &&
+    if (!identity.hasAcpMessageId &&
         tracker.snapshot.composition != AcpContentComposition.mixed &&
         blocks.every((block) => block is AcpMappedTextContentBlock)) {
       final text = blocks.whereType<AcpMappedTextContentBlock>().map((block) => block.text).join();
--- a/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
@@ -180,6 +180,16 @@ class AcpEventMapper {
   /// characters in the opaque agent-supplied ids.
   final Map<String, Map<String, _LiveTool>> _liveTools = {};
 
+  /// Resolves the live session that owns [toolCallId], when a harness extension
+  /// omits `sessionId` but carries the originating tool call id.
+  String? sessionIdForToolCallId({required String? toolCallId}) {
+    if (toolCallId == null || toolCallId.isEmpty) return null;
+    for (final entry in _liveTools.entries) {
+      if (entry.value.containsKey(toolCallId)) return entry.key;
+    }
+    return null;
+  }
+
   /// sessionId -> current turn number, advanced by [beginTurn].
   final Map<String, int> _turnSeq = {};
 
@@ -422,6 +432,32 @@ class AcpEventMapper {
       content: update["content"],
       scope: tracker.mappingScope,
     );
+    return appendAssistantMappedBlocks(
+      sessionId: sessionId,
+      identityUpdate: update,
+      blocks: blocks,
+      evaluateTextHaltNotice: true,
+    );
+  }
+
+  /// Appends pre-mapped assistant content into the same ordered message state
+  /// used by `agent_message_chunk`. Harness extensions that surface images
+  /// outside standard ACP chunks call this after local normalization.
+  List<BridgeSseEvent> appendAssistantMappedBlocks({
+    required String sessionId,
+    required Map<String, dynamic> identityUpdate,
+    required Iterable<AcpMappedContentBlock> blocks,
+    required bool evaluateTextHaltNotice,
+  }) {
+    final identity = _chunkIdentity(
+      sessionId: sessionId,
+      update: identityUpdate,
+      role: _ChunkRole.assistant,
+    );
+    final tracker = (_contentTrackers[sessionId] ??= {}).putIfAbsent(
+      identity.messageId,
+      AcpContentTracker.new,
+    );
     if (blocks.isEmpty) return const [];
     final hasTrackableContent = blocks.any(
       (block) => block is AcpMappedImageContentBlock || (block is AcpMappedTextContentBlock && block.text.isNotEmpty),
@@ -430,7 +466,8 @@ class AcpEventMapper {
       _closeCurrentIdlessAssistantContent(sessionId: sessionId);
     }
 
-    if (!identity.hasAcpMessageId &&
+    if (evaluateTextHaltNotice &&
+        !identity.hasAcpMessageId &&
         tracker.snapshot.composition != AcpContentComposition.mixed &&
         blocks.every((block) => block is AcpMappedTextContentBlock)) {
       final text = blocks.whereType<AcpMappedTextContentBlock>().map((block) => block.text).join();
--- a/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
+++ b/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
@@ -1,10 +1,18 @@
 import "package:acp_plugin/acp_plugin.dart";
 import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 
+import "cursor_event_mapper.dart";
+
 /// Adds Cursor's question extensions on top of the standard ACP permission
 /// handling: `cursor/ask_question` (multiple-choice) and `cursor/create_plan`
 /// (accept/reject), both blocking requests that the bridge surfaces as
 /// questions.
+///
+/// Also acknowledges Cursor's non-blocking extension *requests* that cursor-agent
+/// misnames as notifications (`sendNonBlockingExtensionNotification` → ACP
+/// `extMethod` → JSON-RPC request): `cursor/generate_image` and
+/// `cursor/update_todos`. Those arrive on [serverRequests], not the notification
+/// stream, so they must be handled here rather than only in [CursorEventMapper].
 ///
 /// NOTE: Cursor's exact reply payload shapes are not formally documented; the
 /// builders below are best-effort and should be confirmed against a real
@@ -13,14 +21,21 @@ class CursorApprovalRegistry extends AcpApprovalRegistry {
   CursorApprovalRegistry({
     required AcpStdioClient client,
     required super.emit,
+    required CursorEventMapper eventMapper,
+    required void Function(BridgeSseEvent event) emitContent,
     super.idGenerator,
     super.activeSessionResolver,
-  }) : super(
+  }) : _eventMapper = eventMapper,
+       _emitContent = emitContent,
+       super(
          respond: (id, result) =>
              client.respondToServerRequest(id: id, result: result),
          respondError: (id, code, message) => client
              .respondToServerRequestWithError(id: id, code: code, message: message),
        );
+
+  final CursorEventMapper _eventMapper;
+  final void Function(BridgeSseEvent event) _emitContent;
 
   @override
   bool handleExtensionRequest(AcpServerRequest request) {
@@ -31,8 +46,29 @@ class CursorApprovalRegistry extends AcpApprovalRegistry {
       case "cursor/create_plan":
         _handleCreatePlan(request);
         return true;
+      case "cursor/generate_image":
+        _handleNonBlockingExtension(request);
+        return true;
+      case "cursor/update_todos":
+        _handleNonBlockingExtension(request);
+        return true;
     }
     return false;
+  }
+
+  /// cursor-agent sends these via `extMethod` (JSON-RPC request with id) even
+  /// though it treats the call as fire-and-forget. Ack immediately and map the
+  /// payload through the same event mapper used for true notifications.
+  void _handleNonBlockingExtension(AcpServerRequest request) {
+    final params = Map<String, dynamic>.from(request.params);
+    if ((params["sessionId"] as String?)?.trim().isEmpty ?? true) {
+      final resolved = resolveSessionId(params);
+      if (resolved.isNotEmpty) params["sessionId"] = resolved;
+    }
+    _eventMapper
+        .map(AcpNotification(method: request.method, params: params))
+        .forEach(_emitContent);
+    respond(request.id, const <String, Object?>{});
   }
 
   void _handleAskQuestion(AcpServerRequest request) {
--- a/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
+++ b/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
@@ -1,18 +1,17 @@
 import "package:acp_plugin/acp_plugin.dart";
 import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 
-import "cursor_event_mapper.dart";
-
 /// Adds Cursor's question extensions on top of the standard ACP permission
 /// handling: `cursor/ask_question` (multiple-choice) and `cursor/create_plan`
 /// (accept/reject), both blocking requests that the bridge surfaces as
 /// questions.
 ///
-/// Also acknowledges Cursor's non-blocking extension *requests* that cursor-agent
-/// misnames as notifications (`sendNonBlockingExtensionNotification` → ACP
-/// `extMethod` → JSON-RPC request): `cursor/generate_image` and
-/// `cursor/update_todos`. Those arrive on [serverRequests], not the notification
-/// stream, so they must be handled here rather than only in [CursorEventMapper].
+/// Cursor's non-blocking extension *requests* (`cursor/generate_image`,
+/// `cursor/update_todos` — sent via `extMethod` even though cursor-agent
+/// treats them as fire-and-forget) are only declared in
+/// [fireAndForgetExtensionMethods]; the base registry acks and re-injects them
+/// into the notification pipeline, where [CursorEventMapper] handles both wire
+/// shapes.
 ///
 /// NOTE: Cursor's exact reply payload shapes are not formally documented; the
 /// builders below are best-effort and should be confirmed against a real
@@ -21,21 +20,21 @@ class CursorApprovalRegistry extends AcpApprovalRegistry {
   CursorApprovalRegistry({
     required AcpStdioClient client,
     required super.emit,
-    required CursorEventMapper eventMapper,
-    required void Function(BridgeSseEvent event) emitContent,
+    required super.onFireAndForgetNotification,
     super.idGenerator,
     super.activeSessionResolver,
-  }) : _eventMapper = eventMapper,
-       _emitContent = emitContent,
-       super(
+  }) : super(
          respond: (id, result) =>
              client.respondToServerRequest(id: id, result: result),
          respondError: (id, code, message) => client
              .respondToServerRequestWithError(id: id, code: code, message: message),
        );
 
-  final CursorEventMapper _eventMapper;
-  final void Function(BridgeSseEvent event) _emitContent;
+  @override
+  Set<String> get fireAndForgetExtensionMethods => const {
+    "cursor/generate_image",
+    "cursor/update_todos",
+  };
 
   @override
   bool handleExtensionRequest(AcpServerRequest request) {
@@ -46,29 +45,8 @@ class CursorApprovalRegistry extends AcpApprovalRegistry {
       case "cursor/create_plan":
         _handleCreatePlan(request);
         return true;
-      case "cursor/generate_image":
-        _handleNonBlockingExtension(request);
-        return true;
-      case "cursor/update_todos":
-        _handleNonBlockingExtension(request);
-        return true;
     }
-    return false;
-  }
-
-  /// cursor-agent sends these via `extMethod` (JSON-RPC request with id) even
-  /// though it treats the call as fire-and-forget. Ack immediately and map the
-  /// payload through the same event mapper used for true notifications.
-  void _handleNonBlockingExtension(AcpServerRequest request) {
-    final params = Map<String, dynamic>.from(request.params);
-    if ((params["sessionId"] as String?)?.trim().isEmpty ?? true) {
-      final resolved = resolveSessionId(params);
-      if (resolved.isNotEmpty) params["sessionId"] = resolved;
-    }
-    _eventMapper
-        .map(AcpNotification(method: request.method, params: params))
-        .forEach(_emitContent);
-    respond(request.id, const <String, Object?>{});
+    return super.handleExtensionRequest(request);
   }
 
   void _handleAskQuestion(AcpServerRequest request) {
--- a/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
+++ b/bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
@@ -10,8 +10,8 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 /// `cursor/update_todos` — sent via `extMethod` even though cursor-agent
 /// treats them as fire-and-forget) are only declared in
 /// [fireAndForgetExtensionMethods]; the base registry acks and re-injects them
-/// into the notification pipeline, where [CursorEventMapper] handles both wire
-/// shapes.
+/// into the notification pipeline, where the Cursor event mapper handles both
+/// wire shapes.
 ///
 /// NOTE: Cursor's exact reply payload shapes are not formally documented; the
 /// builders below are best-effort and should be confirmed against a real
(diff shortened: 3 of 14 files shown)


bridge/sesori_plugin_cursor/lib/src/cursor_approval_registry.dart
```diff
@@ -6,13 +6,21 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 /// (accept/reject), both blocking requests that the bridge surfaces as
 /// questions.
 ///
+/// Cursor's non-blocking extension *requests* (`cursor/generate_image`,
+/// `cursor/update_todos` — sent via `extMethod` even though cursor-agent
+/// treats them as fire-and-forget) are only declared in
+/// [fireAndForgetExtensionMethods]; the base registry acks and re-injects them
+/// into the notification pipeline, where [CursorEventMapper] handles both wire
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

Replaced the [CursorEventMapper] doc link with plain text in the CursorApprovalRegistry doc comment, since the import was removed when the registry returned to approvals-only.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+        _logUnavailable();
+        return const [];
+      }
+      final length = file.lengthSync();
+      if (length <= 0) {
+        _logUnavailable();
+        return const [];
+      }
+      if (length > maxInlineMessageAttachmentBytes) {
+        return [
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

mapPath now reads through a single opened descriptor capped at maxInlineMessageAttachmentBytes + 1 (RandomAccessFile.readSync), so a file that grows after any check can never pull more than the inline budget into memory. The separate lengthSync pre-check is gone — size is enforced during the read itself.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+                      mime: _mimeHintFromBasename(basename: basename),
+                      filename: basename,
+                    )
+                    as PluginMessageAttachmentMetadata,
+            reason: AcpImageDegradationReason.oversized,
+          ),
+        ];
+      }
+
+      final bytes = file.readAsBytesSync();
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

Same root cause as the qodo thread on this file: the check-then-read pair was replaced with one bounded read from a single opened descriptor (max + 1 bytes), so the TOCTOU gap is closed by construction rather than by a pre-check.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+      final mime = _mimeFromBytes(bytes: bytes) ?? _mimeHintFromBasename(basename: basename);
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

Inline bytes now require a recognized image signature from _mimeFromBytes; when sniffing fails the payload degrades to a metadata block with AcpImageDegradationReason.invalid. The extension hint is only used for the metadata display mime — it can never promote arbitrary bytes to an image. Regression test added: unknown bytes in a .png-named file yield metadata, not inline content.

bridge/sesori_plugin_cursor/lib/src/repositories/cursor_generated_image_reader.dart
```diff
@@ -0,0 +1,144 @@
...
+            filename: basename,
+            reason: AcpImageDegradationReason.oversized,
+          ),
+        ];
+      }
+
+      final bytes = file.readAsBytesSync();
+      final mime = _mimeFromBytes(bytes: bytes) ?? _mimeHintFromBasename(basename: basename);
+      if (mime == _unsupportedMime) {
+        return [
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

Good catch — the direct-construction refactor bypassed the encoded-size policy the standard ACP path applies. The mapper now checks the base64 payload with the shared isInlineMessageAttachmentWithinSizeLimit before constructing the inline attachment; raw bytes at the inline limit degrade to oversized metadata. Covered by a new test.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+      final content = <String, Object?>{
+        "type": "image",
+        "data": base64Encode(bytes),
+        "mimeType": mime,
+      };
+      if (uri != null) content["uri"] = uri;
+      return _contentMapper.map(content: content);
+    } on Object {
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

The catch-all now passes the caught error and stack trace to Log.w per the repo logging rules. Local paths in logs are acceptable here (logs stay on the user's machine); paths still never cross the plugin boundary on the wire.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+      _logUnavailable();
+      return const [];
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit 1cce3402)

The on Object handler now logs the caught error and stack trace via Log.w(..., error, stack) instead of a bare generic warning, so a recovered failure remains diagnosable.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,119 @@
...
+      filename: p.basename(normalizedPath),
+    );
+    final uri = basename == null ? null : "file:///$basename";
+
+    try {
+      final file = File(normalizedPath);
+      if (!file.existsSync()) {
+        _logUnavailable();
+        return const [];
+      }
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Partially addressed (in commit 1cce3402)

Implemented the content-trust half: unknown byte signatures now degrade to metadata (AcpImageDegradationReason.invalid) instead of being promoted to an image via the filename extension, so non-image data can no longer be forwarded as inline image content.

Not implementing the resolved-path allowlist / symlink blocking: cursor-agent is a local process running with the user's own permissions, so it can already read any file the bridge could — an allowlist adds no meaningful boundary against a compromised agent. And generated-image output locations are not confined to a knowable root (project dir, temp, or user-chosen paths), so an allowlist would break legitimate generations. The wire contract also never exposes the local path — only the basename and sniffed-image bytes cross the plugin boundary.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,94 @@
...
+  /// drop the payload (an event stamped with "" is discarded by the client).
+  String? _extensionSessionId(Map<String, dynamic> params) {
+    final explicit = params["sessionId"];
+    if (explicit is String && explicit.trim().isNotEmpty) return explicit;
+    final toolCallId = params["toolCallId"];
+    if (toolCallId is String && toolCallId.isNotEmpty) {
+      final fromTool = sessionIdForToolCallId(toolCallId: toolCallId);
+      if (fromTool != null) return fromTool;
+    }
+    return _lastTurnSessionId;
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Not addressed

The last-begun-turn fallback is a deliberate design choice from this PR's refactor: ACP provides no request→turn correlation, so the plugin's activeTurnSessionId has the same fundamental ambiguity — with multiple in-flight turns it also guesses (most recent dispatch) and logs a warning; it only biases the guess differently. The fallback is the last resort of the resolution chain: explicit sessionId wins, then the originating toolCallId (which real cursor-agent payloads carry, as seen in wire traces), and only then the last begun turn. The misattribution scenario requires concurrent cross-session turns AND a payload missing both identifiers. The injected resolver was intentionally removed in this refactor (approved direction) to eliminate the two-phase binding; reintroducing it would restore that complexity to marginally re-bias a last-resort guess.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,94 @@
...
+    required CursorGenerateImageMapper generateImageMapper,
+  }) : _generateImageMapper = generateImageMapper,
+       super(agentId: pluginId);
+
+  final CursorGenerateImageMapper _generateImageMapper;
+
+  /// The session that most recently began a turn — the last-resort attribution
+  /// for Cursor extension payloads that omit `sessionId` (cursor-agent's
+  /// extension calls carry only the originating `toolCallId`, or nothing).
+  String? _lastTurnSessionId;
```
daniil-shumko: `_lastTurnSessionId` is set in `beginTurn` but never cleared. `AcpPlugin.deleteSession` deliberately nulls its own plugin-side equivalent (`acp_plugin.dart:1253`) and then calls `eventMapper.forgetSession(sessionId)` — but this mapper-side copy survives that cleanup. A `sessionId`-less, `toolCallId`-less `cursor/generate_image` / `cursor/update_todos` arriving after the last-active session was deleted gets attributed to the deleted session and emits events for it.

Suggest mirroring the base plugin's symmetry:

```dart
@override
void forgetSession(String sessionId) {
  if (_lastTurnSessionId == sessionId) _lastTurnSessionId = null;
  super.forgetSession(sessionId);
}
```

(Alternatively host the field in `AcpEventMapper` next to `forgetSession`, so set/clear live in one class.)

This is the only pre-merge ask in my review; everything else is take-or-leave polish.

bridge/sesori_plugin_cursor/lib/src/repositories/cursor_generated_image_reader.dart
```diff
@@ -0,0 +1,163 @@
...
+      ".bmp" => "image/bmp",
+      ".gif" => "image/gif",
+      ".jpg" || ".jpeg" => "image/jpeg",
+      ".png" => "image/png",
+      ".webp" => "image/webp",
+      _ => _unsupportedMime,
+    };
+  }
+
+  static String? _mimeFromBytes({required Uint8List bytes}) {
```
daniil-shumko, line 112: This is now the third byte-identical copy of this signature table (+ `_startsWith`): `client/module_core/lib/src/repositories/message_image_repository.dart` (`_hasExpectedSignature`) and `client/app/lib/capabilities/media/composer_image_picker.dart` (`_sniffImageMime`, whose doc already notes it mirrors `MessageImageRepository`). The sync is load-bearing across the bridge/client boundary: the client validates inbound inline images, so if the tables ever drift, the bridge inlines what the client then rejects.

`sesori_shared` already owns the adjacent policy this mapper consumes (`maxInlineMessageAttachmentBytes`, `isInlineMessageAttachmentWithinSizeLimit`, `normalizePluginMessageAttachmentFilename`) — a shared `sniffImageMime(bytes:)` there would collapse all three copies. Fine as a fast-follow to keep this PR contained.

bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
```diff
@@ -422,6 +432,50 @@ class AcpEventMapper {
...
+    );
+  }
+
+  /// Appends assistant image blocks normalized locally on the bridge (a harness
+  /// extension that surfaces images outside standard ACP chunks) into the same
+  /// ordered message state used by `agent_message_chunk`.
+  List<BridgeSseEvent> appendAssistantImageBlocks({
+    required String sessionId,
+    required String? messageId,
+    required Iterable<AcpMappedContentBlock> blocks,
```
daniil-shumko: `evaluateTextHaltNotice: false` is provably unreachable for this path: the halt branch already requires `blocks.every((b) => b is AcpMappedTextContentBlock)` on a non-empty list, which is always false for image blocks. If this param were typed `Iterable<AcpMappedImageContentBlock>` (the existing supertype of the inline/metadata blocks), the invalid call becomes unrepresentable and the flag on `_appendAssistantBlocks` can be deleted — one fewer boolean knob on the shared helper.

The flag does state the policy explicitly, so this is a style call; the typed param gets the same guarantee from the signature instead.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,94 @@
...
+    );
+  }
+
+  /// The session an extension payload belongs to: its explicit `sessionId`,
+  /// else the session owning the originating `toolCallId`, else the most
+  /// recent turn's session. Null only when none is available — the caller must
+  /// drop the payload (an event stamped with "" is discarded by the client).
+  String? _extensionSessionId(Map<String, dynamic> params) {
+    final explicit = params["sessionId"];
+    if (explicit is String && explicit.trim().isNotEmpty) return explicit;
```
daniil-shumko: Nit: checks `explicit.trim().isNotEmpty` but returns the untrimmed value; the sibling resolver (`AcpApprovalRegistry.resolveSessionId`) returns the trimmed id. Harmless in practice, just inconsistent with the precedent.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,94 @@
...
-    return const [];
+    // cursor/task and other extension notifications have no sesori analog.
+    return super.mapExtension(notification);
+  }
+
+  List<BridgeSseEvent> _mapGenerateImage({required AcpNotification notification}) {
+    final params = notification.params;
+    final sessionId = _extensionSessionId(params);
+    final path = _pathFromGenerateImageParams(params: params);
+    if (sessionId == null || path == null) {
```
daniil-shumko, line 57: When neither the session nor a path resolves, the payload vanishes without a trace — and since the registry already acked the request, this is the last place a lost image could have been observed. A `Log.w` with the method plus the path (or its absence) would make a "generated image never appeared" report diagnosable from bridge logs. Local logs are the sanctioned place for path context; only the transport must stay basename-only.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,163 @@
...
+          PluginMessageAttachment.metadata(
+                mime: mime,
+                filename: filename,
+              )
+              as PluginMessageAttachmentMetadata,
+      reason: reason,
+    );
+  }
+
+  void _logUnavailable() {
```
daniil-shumko: Nit: the empty-read call site (line 42) knows the path but this log carries no context about which file was unavailable, while the catch-path log below gets it via the exception. Passing the path through here would keep the two failure logs equally diagnosable.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,94 @@
...
+    final toolCallId = params["toolCallId"];
+    if (toolCallId is String && toolCallId.isNotEmpty) {
+      final fromTool = sessionIdForToolCallId(toolCallId: toolCallId);
+      if (fromTool != null) return fromTool;
+    }
+    return _lastTurnSessionId;
+  }
+
+  static String? _pathFromGenerateImageParams({required Map<String, dynamic> params}) {
+    for (final key in const ["filePath", "path", "filename"]) {
```
daniil-shumko: `filePath` is the live-verified shape and `path` has provenance from the pre-PR test, but `filename` looks speculative: a bare filename would resolve against the bridge process CWD, which is meaningless here — and it's the one key without a test. Consider dropping it, or noting the cursor-agent trace that motivated it.

bridge/sesori_plugin_cursor/test/cursor_generate_image_mapper_test.dart
```diff
@@ -0,0 +1,83 @@
...
+import "package:cursor_plugin/src/repositories/mappers/cursor_generate_image_mapper.dart";
+import "package:sesori_shared/sesori_shared.dart" show maxInlineMessageAttachmentBytes;
+import "package:test/test.dart";
+
+void main() {
+  group("CursorGenerateImageMapper", () {
+    const mapper = CursorGenerateImageMapper();
+
+    test("maps a bounded local PNG into an inline image block", () async {
+      final file = File("${Directory.systemTemp.path}/output.png");
```
daniil-shumko: Fixed names in `Directory.systemTemp` (`output.png` here, `legacy-output.png` in `cursor_event_mapper_test.dart`) can race when two checkouts run the suite concurrently — a real workflow in this repo (parallel worktrees). The sibling tests in this PR already use microsecond-stamped names; these two are the odd ones out.

daniil-shumko (commented):
Deep review of the branch (checked out locally; read the surrounding event-mapper/registry/plugin machinery, the standard ACP image path, and the client-side image validation this must stay consistent with).

**Overall: strong PR.** The fire-and-forget classification at the registry seam with re-injection into a single mapping path is the right design, and the image path re-enters the standard tracker/envelope pipeline at exactly the right altitude (an id-less image joins the currently open assistant envelope, a `messageId`-stamped one joins its ACP message — zero client/persistence changes needed). The single-descriptor bounded read + signature-required inline + encoded-size bound is stricter than the standard ACP image path precisely where trust differs (no agent-declared MIME) and consistent with it everywhere else.

8 inline comments: **1 small lifecycle gap I'd fix pre-merge** (mapper-side `_lastTurnSessionId` never cleared on session deletion), 2 structural suggestions (shared signature sniffer; typed param instead of the `evaluateTextHaltNotice` flag), and 5 polish/test nits.

Not anchored to any line: mapper-level coverage for the changed `cursor/update_todos` fallback resolution (it used to require an explicit `sessionId`) and for a `messageId`-carrying `generate_image` joining an existing assistant envelope would be cheap confidence adds.

Verified locally: `dart analyze --fatal-infos` clean and full `dart test` green in both `sesori_plugin_cursor` (120 tests) and `sesori_plugin_acp` (216 tests).

_Review performed with Claude Fable 5 (`claude-fable-5`) at **max** reasoning effort._

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: **Design duplication — two owners of "which session does an unattributed payload belong to."**

`_lastTurnSessionId` here is a same-named, weaker copy of `AcpPlugin._lastTurnSessionId` (`acp_plugin.dart:133`), which the plugin already refines into `activeTurnSessionId`: prefer the single in-flight turn, log on ambiguity, fall back to last dispatch. Commit 2 deliberately dropped that resolver from the mapper, so the concept now has two owners with divergent semantics — and the registry still receives the good one (`activeSessionResolver: () => activeTurnSessionId`) for blocking requests.

Verified consequences (scratch tests against this branch):

- `beginTurn("sA")` → `beginTurn("sB")`, then a `cursor/generate_image` with **no** `toolCallId` → stamped `sB`, even when only `sA` is still in flight. `activeTurnSessionId` would answer `sA`.
- `beginTurn("sGone")` → `forgetSession("sGone")` → a later extension still emits 2 events stamped `sGone`; `forgetSession` is not overridden to clear this field.

Practical severity is lower than it reads: when a `toolCallId` *is* present the `_liveTools` chain resolves correctly (verified — it picks `sA` over a later `sB`), so this only bites payloads carrying neither field. But the fix is small and restores single ownership: keep the chain, make its last step a `String? Function()` the plugin supplies (the same closure the registry already gets), and override `forgetSession` to clear it.

Minor, same method: `explicit.trim().isNotEmpty` gates on the trimmed value but returns the untrimmed one, unlike `AcpApprovalRegistry.resolveSessionId`, which returns the trimmed id. A padded id would key a distinct pseudo-session across `_contentTrackers`/`_startedParts`.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: **Speculative scope, and the third key is semantically wrong.**

Only `filePath` is live-verified (per the PR body and the approval-registry test). AGENTS.md is explicit: smallest change that fully solves the demonstrated problem, no machinery for hypothetical shapes.

Worse than surplus: a *filename* is not a path. If `filename` ever fires with a bare name, `File(name)` resolves against the **bridge process** working directory — not the session's project directory — so it reads an unrelated file or nothing at all, and the failure is invisible (`mapPath` returns `[]`).

Suggest keeping `filePath` (plus `path` only if a real trace shows it) and dropping `filename`. If relative paths genuinely need supporting, resolve them against the session directory the mapper already knows via `projectForSession`.

bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
```diff
@@ -422,6 +432,50 @@ class AcpEventMapper {
...
+      evaluateTextHaltNotice: false,
+    );
+  }
+
+  List<BridgeSseEvent> _appendAssistantBlocks({
+    required String sessionId,
+    required ({String messageId, bool hasAcpMessageId}) identity,
+    required AcpContentTracker tracker,
+    required Iterable<AcpMappedContentBlock> blocks,
+    required bool evaluateTextHaltNotice,
```
daniil-shumko: **This parameter is dead — it cannot change behavior.**

The halt-notice branch it guards also requires `blocks.every((block) => block is AcpMappedTextContentBlock)`, and the only caller of `appendAssistantImageBlocks` feeds it `CursorGenerateImageMapper.mapPath`, which returns exclusively `AcpMappedInlineImageContentBlock` / `AcpMappedMetadataImageContentBlock` — or an empty list, which returns at the `blocks.isEmpty` guard above. So `every(isText)` is already false on every path that passes `false` here.

Net effect: a boolean flag added to the shared helper, plus a "why is this here?" for the next reader, for zero behavioral difference. Two ways out, both better than the flag:

1. Drop the parameter and let the existing `every(isText)` guard do the work (a one-line comment carries the intent).
2. Better — type `appendAssistantImageBlocks`' parameter as `Iterable<AcpMappedImageContentBlock>`. That makes the guarantee structural instead of conventional and documents the contract at the seam, and the flag disappears for free.

bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
```diff
@@ -319,6 +334,18 @@ class AcpApprovalRegistry {
       _handlePermission(request);
       return;
     }
+    if (fireAndForgetExtensionMethods.contains(request.method)) {
+      // Fire-and-forget: the agent awaits no result. Ack so it never blocks,
+      // then re-inject into the notification pipeline.
+      _respond(request.id, const <String, Object?>{});
+      final forward = _onFireAndForgetNotification;
+      if (forward == null) {
```
daniil-shumko: **Impossible-but-representable state.** `onFireAndForgetNotification` is nullable, yet both production call sites (`AcpPlugin.buildApprovalRegistry`, `CursorPlugin.buildApprovalRegistry`) and the one test always pass it — and `CursorApprovalRegistry` already marks it `required` while the type still admits null.

So "declared `fireAndForgetExtensionMethods` + no forward wired" can only be reached by constructing a registry that no code constructs, and it is paid for with a runtime branch and a warning log. Making the parameter required and non-nullable on the base deletes the branch, the log line, and the impossible state (AGENTS.md: *make impossible state combinations unrepresentable*).

bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
```diff
@@ -319,6 +334,18 @@ class AcpApprovalRegistry {
...
+        Log.w("[acp] ${request.method} declared fire-and-forget with no notification forward wired; dropping");
+      } else {
+        forward(AcpNotification(method: request.method, params: request.params));
```
daniil-shumko: Minor / hardening: this forward is invoked from the **serverRequests** subscription, so the notification-mapping path now shares a failure domain with approvals. A throw inside the event mapper would kill `_subscription` — after which permission requests stop being answered and turns hang — rather than only breaking the notifications stream as before.

`CursorGenerateImageMapper.mapPath` catches `Object`, so today's exposure is limited to the mapper/tracker code itself. A `try`/`catch` (log + continue) around this call keeps the blast radius on the channel that owns it.

daniil-shumko (commented):
## Deep review — architecture and elegance

_Reviewed with **Claude Opus 5** (`claude-opus-5[1m]`, 1M context) at **max** reasoning effort._

**Verdict: the core is elegant.** Reclassifying the wire shape at the registry seam, re-injecting into the single notification pipeline, and appending through the *existing* content tracker is the right architecture, and commit 3's hardening is genuinely good. Everything below is at the edges — a round of small fixes, not a redesign.

### What's well done

- **The seam is declarative and backend-neutral.** `fireAndForgetExtensionMethods` lets Cursor declare two method names while the base owns all the wire mechanics: no mapper dependency in the registry, no second emit channel, no params mutation, one mapping site for both wire shapes.
- **Reusing `appendAssistantImageBlocks` is the difference between a feature and a special case.** Verified locally that a text → image → text sequence produces **one** message with correctly ordered parts:
  ```
  MSG   s1-t1-assistant-a0
  PART  s1-t1-assistant-a0-text      type=text
  PART  s1-t1-assistant-a0-image-1   type=file
  PART  s1-t1-assistant-a0-text-1    type=text
  ```
  The image inherits the count limit, aggregate-byte budget, id-less-envelope closing, and part identity for free.
- **The commit-3 security posture is correct.** Requiring an image signature means an agent-supplied arbitrary path can never push non-image bytes inline — worst case is a basename in metadata. The single-descriptor bounded read caps memory at 5 MB + 1 regardless of file size and closes the TOCTOU.

### One PR-level point (no inline anchor)

This reverses a locked scope decision. `.plan/completed/output-image-support/PLAN.md` excluded *"Reading local files or forwarding local paths"* and `cursor/generate_image` specifically, because its payload is *"a host-local path without the standard ACP session/content identity needed for safe replay."* This PR solves the identity half for **live** rendering; the replay half is unsolved (and acknowledged in the PR body). Net user-visible result: a Cursor-generated image renders live and then vanishes on history reload — a divergence from that plan's success criterion 1 (same identity live and after reload). That's a cursor-agent limitation rather than a bridge shortcut, but it deserves its own line in the PR body / release notes rather than a bullet under "Risk".

### Verification run

`dart analyze` clean and `dart test` green in both `sesori_plugin_acp` (216 tests) and `sesori_plugin_cursor` (120 tests), plus four throwaway scratch tests used to confirm the claims in the inline comments (since removed).

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
```diff
@@ -0,0 +1,163 @@
...
+import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
+import "package:sesori_shared/sesori_shared.dart"
+    show isInlineMessageAttachmentWithinSizeLimit, maxInlineMessageAttachmentBytes;
+
+/// Reads Cursor `cursor/generate_image` host paths and maps them into the same
+/// bounded inline-image content blocks used for standard ACP assistant images.
+///
+/// Local paths never leave this mapper; only basename metadata and inline bytes
+/// cross the plugin boundary.
+final class CursorGenerateImageMapper {
```
daniil-shumko: **Layering/naming:** this is the only mapper in the entire bridge that imports `dart:io` — I scanned every file under a `*mappers*` path and it is the sole hit. Everything else in `repositories/mappers/` (including `AcpContentMapper` and `CursorCatalogMapper`) is a pure transformation; filesystem access is the API/repository role in `Foundation -> API -> Repository -> Service -> Consumer`.

Not worth a large refactor, but the name undersells what the class does. Either rename to something honest (`CursorGeneratedImageReader`), or split "read bounded bytes from a path" from "map bytes -> content blocks" — the second half then unit-tests without touching the filesystem, which would remove the 5 MB temp-file writes from the suite.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
daniil-shumko: **The failure log drops the one identifier that would make it actionable.** This fires for an empty path *and* for a zero-byte file, and the `catch` above logs error + stack but likewise omits the path — three different failures, one indistinguishable message.

AGENTS.md explicitly lists *paths* among the diagnostics to preserve in local bridge logs, and says to strip only known user data with no debugging value (prompt/transcript content). The class doc's "local paths never leave this mapper" is about the plugin/wire boundary, not local logging — logs are not auto-submitted. When a user reports "my generated image didn't show", `[cursor] generate_image source unavailable` says nothing.

Suggest including the path and distinguishing unreadable from empty. Related nit: a 4-byte garbage file degrades to a metadata chip while a 0-byte file emits nothing at all — worth making consistent one way or the other.

bridge/sesori_plugin_cursor/lib/src/repositories/cursor_generated_image_reader.dart
```diff
@@ -0,0 +1,163 @@
...
+    );
+  }
+
+  void _logUnavailable() {
+    Log.w("[cursor] generate_image source unavailable");
+  }
+
+  static const String _unsupportedMime = "application/octet-stream";
+
+  static String _mimeHintFromBasename({required String? basename}) {
```
daniil-shumko, line 104: **Duplication:** this switch and `AcpContentMapper._supportedRasterMimeEssences` are two independent enumerations of the same closed set (bmp/gif/jpeg/png/webp), in two packages. `_metadata(...)` just below is also a verbatim copy of `AcpContentMapper._metadata`. Adding a sixth inline-able type means remembering to edit both.

Small, but this is exactly the drift the shared content mapper exists to prevent. Exporting one supported-type list plus an `acpMetadataImageBlock(...)` helper from the ACP package would remove both copies.

bridge/sesori_plugin_cursor/test/cursor_event_mapper_test.dart
```diff
@@ -48,7 +53,124 @@ void main() {
...
+            )
+            .whereType<BridgeSseMessagePartUpdated>()
+            .single
+            .part
+            .type,
+        PluginMessagePartType.file,
+      );
+    });
+
+    test("cursor/generate_image resolves sessionId from toolCallId", () async {
```
daniil-shumko: **This test cannot fail for the reason its name claims.** The `mapper` is declared once at group scope and shared, and the test calls `beginTurn("s1")` before registering the tool call on `s1` — so the last-begun-turn fallback also returns `s1`. The assertion is only `part.type == file`, so it passes identically whether the `toolCallId` chain works, is broken, or is deleted.

The chain does work — I confirmed it by beginning a turn on a *different* session after registering the tool call, and the part came back stamped with the tool call's session. The shipped test should do the same: `beginTurn("sA")` → tool call on `sA` → `beginTurn("sB")` → generate_image with the `sA` tool call id → assert `part.sessionID == "sA"`.

Same shape one test up ("resolves sessionId from the last begun turn"): it uses a fresh mapper so it does exercise the fallback, but asserting `part.type` rather than `part.sessionID` leaves the actual claim unchecked.

Nit: three of the new tests use fixed temp filenames (`output.png`, `legacy-output.png`, `cursor-active-turn-image.png`) while their neighbours in the same files use a `microsecondsSinceEpoch` suffix — worth being consistent, since Dart runs test files concurrently.

bridge/sesori_plugin_cursor/test/cursor_event_mapper_test.dart
```diff
@@ -48,7 +53,124 @@ void main() {
       expect(events.whereType<BridgeSseMessagePartDelta>().single.delta, "hi");
     });
 
-    test("standard ACP images are inherited while cursor image extensions stay dropped", () {
+    test("cursor/generate_image maps to a standard inline file part", () async {
```
daniil-shumko, line 76: **Test gap: nothing covers the property the whole design rests on** — that the image lands as an *ordered part inside the live assistant message* rather than as a standalone one. That is the entire justification for routing through `appendAssistantImageBlocks` instead of emitting a sidecar message, and a regression back to a standalone message would leave every test in this file green.

I verified it by hand (text chunk -> generate_image -> text chunk yields one message `s1-t1-assistant-a0` with parts `-text`, `-image-1`, `-text-1`); a test asserting exactly that sequence would be the highest-value addition here.

Also uncovered: an unresolvable session (no `sessionId`, no `toolCallId`, no begun turn) must drop the payload — currently unasserted, and it is the guard that keeps an event stamped `""` off the wire.

bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

`onFireAndForgetNotification` is now `required` and non-nullable on both the `AcpApprovalRegistry` constructor and `forClient`, and the null branch plus its warning log are deleted. The base-registry test supplies a no-op.

bridge/sesori_plugin_acp/lib/src/acp_approval_registry.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

The forward is wrapped in try/catch (log + continue), so a throw inside notification mapping breaks only that payload and never the serverRequests subscription that answers permissions. Added a registry test that makes the forward throw and asserts the request is still acked and a subsequent permission request still surfaces.

bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Went with the typed variant: `appendAssistantImageBlocks` now takes `Iterable<AcpMappedImageContentBlock>` and the `evaluateTextHaltNotice` flag is deleted from `_appendAssistantBlocks` — the existing `every(is AcpMappedTextContentBlock)` guard carries the policy, now guaranteed structurally.

bridge/sesori_plugin_acp/lib/src/acp_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Took option 2: the parameter is typed `Iterable<AcpMappedImageContentBlock>`, the flag is gone, and `CursorGeneratedImageReader.read` (né `CursorGenerateImageMapper.mapPath`) returns `List<AcpMappedImageContentBlock>`, so the guarantee is structural at every link of the chain.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Fixed via the stronger shape from your follow-up round: the mapper-side `_lastTurnSessionId` and its `beginTurn` override are deleted entirely, and the chain's last step is now the plugin-supplied `activeTurnSessionId` closure. `AcpPlugin.deleteSession` already nulls its own field and removes the session from `_inFlightTurnSessions`, so a post-delete extension payload resolves to null and is dropped — covered by the new unresolvable-session test.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

`_mapGenerateImage` now logs a `Log.w` with the method, whether a session resolved, and the path (or its absence) before dropping. The reader also logs the path on unreadable sources, so every silent-loss path leaves a trace in bridge logs.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

`_extensionSessionId` now returns the trimmed id, matching `AcpApprovalRegistry.resolveSessionId`.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Single ownership restored: `CursorEventMapper` takes a required `String? Function() activeSessionResolver` — the same `() => activeTurnSessionId` closure the registry gets — as the chain's last step, and the mapper-side `_lastTurnSessionId`/`beginTurn` override are deleted. The construction cycle is broken with a `late final plugin` closure in the factory, so there is no two-phase bind API and the field stays non-nullable.

No `forgetSession` override turned out to be needed: the mapper holds no attribution state anymore, and `deleteSession` already clears the plugin-side field, so both verified scenarios resolve correctly (overlapping turns prefer the single in-flight turn; post-delete payloads drop — new test covers the drop). The trim inconsistency is fixed as well.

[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Update: the maintainer review reached the same conclusion, so the earlier decline is superseded — the mapper now uses the plugin's `activeTurnSessionId` resolver as the chain's last step (explicit sessionId → toolCallId → active turn), and the mapper-side last-begun-turn copy is deleted. Session deletion clears the plugin-side state, so a late event for a deleted session resolves to null and is dropped.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Dropped `filename`; kept `filePath` (live-verified) and `path` (pre-PR test provenance), with a doc comment recording that provenance.

daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

`filename` is gone — agreed that a bare name resolving against the bridge process CWD is worse than surplus. The accepted keys are `filePath` + `path`, and the doc comment notes that if relative paths ever show up in a real trace they must resolve against the session's project directory, not the process CWD.

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Renamed to `CursorGeneratedImageReader` (method `read`) and moved it out of `repositories/mappers/` up to `repositories/`, so `mappers/` keeps only pure transformations. Kept read+map together in the one small class rather than splitting — the mapping half shrank anyway once the metadata builder and supported-type set moved to `AcpContentMapper` (sibling thread).

bridge/sesori_plugin_cursor/lib/src/repositories/mappers/cursor_generate_image_mapper.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

The unreadable-source log now carries the path, and the empty-path case logs its own distinct message. The zero-byte read no longer logs-and-drops at all — it degrades to a metadata chip like any other signature-less content (see the sibling thread on line 114).

daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

The three failures are now distinguishable: an empty path logs "empty source path", an unreadable file logs the path with error + stack, and a zero-byte file no longer drops silently — the `isEmpty` early-return is deleted so it flows into the no-signature branch and degrades to a metadata chip exactly like 4-byte garbage (test added asserting the chip and its extension-hint mime). The event-mapper drop log carries the path too.

bridge/sesori_plugin_cursor/lib/src/repositories/cursor_generated_image_reader.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

`AcpContentMapper` now exposes `supportedRasterMimeEssences` and a static `metadataImageBlock(...)`; the reader uses both, so the verbatim `_metadata` copy is gone. Its extension hint keeps only the `.jpg`→`jpeg` alias locally and derives membership from the shared set, so adding a sixth inline-able type is a one-place edit (byte-signature sniffing aside, tracked in its own thread).

bridge/sesori_plugin_cursor/lib/src/repositories/cursor_generated_image_reader.dart
daniil-shumko: [Sesori reply] Not addressed

Deferred as the fast-follow you suggested, to keep this PR bridge-only — collapsing the three signature tables means touching `module_core` and the app shell in the same change (`sesori_shared.sniffImageMime` + two client call sites). The bridge keeps exactly one copy in the meantime, now in `CursorGeneratedImageReader`.

bridge/sesori_plugin_cursor/test/cursor_event_mapper_test.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

Added the ordered-parts test: text chunk → generate_image → text chunk yields parts `-text`, `-image-1`, `-text-1` on the one `s1-t1-assistant-a0` message, with an explicit assertion that the image emits no new message envelope. Also added the unresolvable-session test: no `sessionId`, no `toolCallId`, resolver answering null → the payload is dropped.

bridge/sesori_plugin_cursor/test/cursor_event_mapper_test.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

The toolCallId test now uses a fresh mapper whose active-session resolver answers a different session (`s-other`), registers the tool call on `sA`, and asserts `part.sessionID == "sA"` — it fails if the chain is broken or deleted. The fallback test likewise asserts `part.sessionID == "s-active"` on a fresh mapper (the fallback is now the plugin-supplied resolver rather than last-begun-turn tracking, per the session-ownership thread). All temp files in both test files now get microsecond-stamped names via shared helpers.

bridge/sesori_plugin_cursor/test/cursor_generate_image_mapper_test.dart
daniil-shumko: [Sesori reply] Addressed (in commit 25998c6f)

All fixed temp names are gone — both test files create files through stamped `tempFile`/`writeTempPng` helpers, so concurrent worktree runs cannot collide.

bridge/sesori_plugin_cursor/lib/src/cursor_event_mapper.dart
```diff
@@ -1,27 +1,109 @@
...
+      if (fromTool != null) return fromTool;
+    }
+    return _activeSessionResolver();
+  }
+
+  /// `filePath` is the live-verified key; `path` has provenance from the
+  /// pre-PR wire tests. A bare filename is deliberately not accepted: it would
+  /// resolve against the bridge process CWD, not the session's project.
+  static String? _pathFromGenerateImageParams({required Map<String, dynamic> params}) {
+    for (final key in const ["filePath", "path"]) {
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 5a58422c)

Valid catch — the doc claimed the CWD hazard while only bare-filename keys were excluded. `_pathFromGenerateImageParams` now accepts only absolute paths (`p.isAbsolute` after trim), logs the rejected value, and a test proves an existing CWD-relative file is not read. Resolving relative paths against the session project was deliberately not built: every live trace shows absolute paths, so rejection is the smallest honest boundary rule until a real trace demands more.

PR #757, opened 2026-08-04
🚧 fix(bridge): bound graceful shutdown when agent work blocks teardown

AlexDochioiu:
## Complexity

🚧 Complex — shutdown-lifecycle coordination across four production files with strict ordering invariants (drain → plugin dispose → lifecycle → shared) and supervised exit-code discipline.

## What

Fixes a graceful-shutdown hang where Ctrl+C during/after an active Cursor (ACP) agent turn left the bridge stuck for 30-60s+ until a second SIGINT force-exited it, with the agent process killed only by the terminal's own SIGINT (exit 130), never by the bridge.

Three coordinated changes:

1. **Interrupt agent work at shutdown start** — `OrchestratorSession.beginShutdown()` now kicks off `PluginRuntime.interruptActiveWorkForShutdown()` (budgeted, best-effort per-plugin `interruptActiveWork`, same contract as the forced-stop path, no terminal handoff). In-flight turns get `session/cancel` and pending approvals resolve while the agent is still alive, so session-teardown drains (relay completions, session operations, plugin-event tails) are not held open by agent-coupled requests.
2. **Coordinator backstop covers the teardown** — the runner starts `shutdownCoordinator.shutdown()` the moment the session's new `shutdownRequested` future completes (any trigger: signal, supervised logout/restart, control-channel loss). The backstop now bounds the whole graceful path; before, it only started after the teardown finished.
3. **Emergency disposal before forced exit** — the coordinator's backstop runs an optional bounded (3s) `emergencyDisposal` before exiting; the runner wires it to `pluginRuntime.disposeStartedApis()` so backend agent processes get SIGTERM rather than being orphaned when the backstop fires. The backstop exit code also consults `restartService.supervisedRestartRequested` so a supervised-restart respawn sentinel (86) survives a backstop that fires mid-teardown.

## Why

The only mechanism that stopped the ACP agent (`AcpPlugin.dispose` → `AcpStdioClient.dispose`, SIGTERM → 5s → SIGKILL) ran in the coordinator's `pluginDispose` phase, which the runner invoked only after the full session teardown completed. A teardown blocked on agent-coupled work (in-flight `session/prompt` with a 30-min timeout, `session/load`/`session/resume` with 2-min timeouts, a lazy agent respawn) therefore hung indefinitely with no deadline and the agent never stopped gracefully.

## Risk and test focus

- Shutdown ordering invariants: drain events → relay close → plugin dispose → lifecycle → shared. The early coordinator start changes only the backstop timer start (cancel time), not phase order — the drain phase still awaits the full teardown before `pluginDispose`.
- Reentrancy: coordinator `shutdown()` is memoized; the early start and the runner's final join share one run, and failures still surface with the existing exit-code policy.
- Supervised exit codes: the sentinel slot stays authoritative; the restart flag is consulted only when the slot is empty (the mid-teardown backstop window).
- Verified: `dart analyze --fatal-infos` clean; full app suite passes (2435 tests); new tests cover the coordinator emergency-disposal cap/prompt-exit and the runtime interrupt (budget, isolation, hang bounding). Architecture implementation review: approved, no findings.

## Expected result

- Ctrl+C during an active agent turn stops the bridge cleanly in seconds: agent turn cancelled, teardown drains complete, plugin dispose reaps the agent.
- Any residual teardown blockage is bounded by the backstop (~40s) with agent processes SIGTERM'd instead of orphaned, and the second-Ctrl+C force-exit no longer needs to exist for the common case.
- No user-visible or database impact otherwise; no wire-contract changes.

---
## Summary by cubic
Fixes shutdown hangs when stopping the bridge during an active agent turn. Starts the shutdown coordinator as soon as teardown begins, cancels agent work up front, and bounds teardown with a backstop so we don’t orphan agent processes.

- **Bug Fixes**
  - Move the agent-work cancel to the shutdown coordinator’s signal phase (after the plugin-runtime fence) and await it before drain.
  - Start the coordinator on `session.shutdownRequested` after startup wiring; register the debug server’s shutdown phases only once it’s listening inside `startDebugServerIfRequested`.
  - Run a capped emergency disposal (6s) before forced exit, via `pluginRuntime.disposeStartedApis()`, so backend agents get SIGTERM/SIGKILL instead of being orphaned.
  - Record the supervised-restart sentinel at decision time via a `BridgeRestartService` callback, ensuring a single, consistent exit code across normal, error, and backstop paths.

<sup>Written for commit 19a4901584e88e78f3e33ce7874c44e6a9a977f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/757?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/app/lib/src/bridge/orchestrator.dart
+++ b/bridge/app/lib/src/bridge/orchestrator.dart
@@ -845,9 +845,9 @@ class OrchestratorSession {
   Future<void> get firstPhoneConnected => _firstPhoneConnectedCompleter.future;
   RoutedRequestDispatcher get routedRequestDispatcher => _routedRequestDispatcher;
 
-  /// Completes when [beginShutdown] starts the graceful shutdown, so
-  /// composition-root machinery (e.g. the shutdown coordinator's backstop)
-  /// can start bounding the session teardown at the same moment.
+  /// Completes when [beginShutdown] starts the graceful shutdown, so the
+  /// composition root can start the ordered shutdown coordinator (agent-work
+  /// interrupt, backstop deadline) the moment the session begins tearing down.
   Future<void> get shutdownRequested => _shutdownCompleter.future;
 
   Future<OrchestratorSessionStartResult> start() {
@@ -1265,11 +1265,6 @@ class OrchestratorSession {
         "[shutdown] cancel() requested"
         "${_inFlightRelayWorkDiagnostic(separator: " — ")}",
       );
-      // Interrupt in-flight agent work now (budgeted, best-effort) so the
-      // teardown drains below are not held open by agent-coupled requests
-      // (an in-flight turn, a resume-load, a lazy agent respawn) while the
-      // agent process is still alive and can answer a cancellation.
-      unawaited(_pluginRuntime.interruptActiveWorkForShutdown());
     } else {
       Log.v("[shutdown] cancel() again (already shutting down)");
     }
--- a/bridge/app/lib/src/bridge/orchestrator.dart
+++ b/bridge/app/lib/src/bridge/orchestrator.dart
@@ -845,6 +845,11 @@ class OrchestratorSession {
   Future<void> get firstPhoneConnected => _firstPhoneConnectedCompleter.future;
   RoutedRequestDispatcher get routedRequestDispatcher => _routedRequestDispatcher;
 
+  /// Completes when [beginShutdown] starts the graceful shutdown, so
+  /// composition-root machinery (e.g. the shutdown coordinator's backstop)
+  /// can start bounding the session teardown at the same moment.
+  Future<void> get shutdownRequested => _shutdownCompleter.future;
+
   Future<OrchestratorSessionStartResult> start() {
     if (_lifecycleFuture != null) {
       return Future.error(StateError("OrchestratorSession has already started"), StackTrace.current);
@@ -1260,6 +1265,11 @@ class OrchestratorSession {
         "[shutdown] cancel() requested"
         "${_inFlightRelayWorkDiagnostic(separator: " — ")}",
       );
+      // Interrupt in-flight agent work now (budgeted, best-effort) so the
+      // teardown drains below are not held open by agent-coupled requests
+      // (an in-flight turn, a resume-load, a lazy agent respawn) while the
+      // agent process is still alive and can answer a cancellation.
+      unawaited(_pluginRuntime.interruptActiveWorkForShutdown());
     } else {
       Log.v("[shutdown] cancel() again (already shutting down)");
     }
--- a/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
+++ b/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
@@ -858,21 +858,6 @@ class BridgeRuntimeRunner {
         );
       }
 
-      // Start the ordered shutdown the moment the session begins shutting
-      // down (any trigger: signal, supervised logout/restart, control-channel
-      // loss), so the coordinator's backstop bounds the session teardown
-      // itself — a teardown blocked on in-flight agent work must not hang the
-      // process with no deadline. The runner's finally joins the same
-      // (memoized) shutdown future, so a failure here is still surfaced there
-      // with the exit-code policy applied.
-      unawaited(
-        // The runner's finally awaits the same (memoized) shutdown future and
-        // applies the exit-code policy there, so the failure is already
-        // surfaced; this early-start copy must not raise an unhandled error.
-        activeRuntime.session.shutdownRequested
-            .then((_) => shutdownCoordinator.shutdown())
-            .catchError((Object _, StackTrace __) {}),
-      );
       registerSignalHandlers(session: activeRuntime.session, subscriptions: subscriptions);
       // start() synchronously subscribes local route-trigger listeners before
       // the debug server can expose mutation routes.
@@ -883,6 +868,24 @@ class BridgeRuntimeRunner {
         debugPort: options.debugPort,
         runtime: activeRuntime,
         shutdownCoordinator: shutdownCoordinator,
+      );
+      // Start the ordered shutdown the moment the session begins shutting down
+      // (any trigger: signal, supervised logout/restart, control-channel loss),
+      // so the coordinator's backstop bounds the session teardown itself — a
+      // teardown blocked on in-flight agent work must not hang the process
+      // with no deadline. Registered only after the debug server is assigned:
+      // the coordinator's signal phase reads `debugServer` (beginShutdown), so
+      // a shutdown racing the async server start would otherwise miss it, and
+      // no agent-coupled work can exist before startup completes anyway. The
+      // runner's finally joins the same (memoized) shutdown future, so a
+      // failure here is still surfaced there with the exit-code policy applied.
+      unawaited(
+        // The runner's finally awaits the same (memoized) shutdown future and
+        // applies the exit-code policy there, so the failure is already
+        // surfaced; this early-start copy must not raise an unhandled error.
+        activeRuntime.session.shutdownRequested
+            .then((_) => shutdownCoordinator.shutdown())
+            .catchError((Object _, StackTrace __) {}),
       );
       // Background: check + download + stage + apply-in-place on a 4h cadence.
       // The swap takes effect on the next launch (or a phone-triggered restart).
--- a/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
+++ b/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
@@ -204,9 +204,24 @@ class BridgeRuntimeRunner {
     // loss listener assigns with `??=` so a loss never overwrites an already
     // decided intentional exit.
     SupervisedExitCode? requestedSupervisedExit;
+    BridgeRestartService? restartService;
     final shutdownCoordinator = BridgeShutdownCoordinator(
       startAbortSignal: startAbortController.signal,
-      backstopExitCode: () => requestedSupervisedExit?.code ?? 0,
+      backstopExitCode: () {
+        final exit = requestedSupervisedExit;
+        if (exit != null) return exit.code;
+        // The supervised-restart respawn sentinel is folded into the slot in
+        // the try's finally, which has not run when the backstop fires
+        // mid-teardown; consult the handoff flag directly so a hung teardown
+        // still reports the sentinel and the GUI respawns.
+        if (restartService?.supervisedRestartRequested ?? false) {
+          return SupervisedExitCode.restart.code;
+        }
+        return 0;
+      },
+      // Last resort before a forced exit: stop plugin backends so their agent
+      // processes are not orphaned when the teardown (or a phase) hangs.
+      emergencyDisposal: () => pluginRuntime?.disposeStartedApis() ?? Future<void>.value(),
     );
     shutdownCoordinator
       ..addPhase(
@@ -762,7 +777,7 @@ class BridgeRuntimeRunner {
         shutdownCoordinator.add(disposable: controlStatusNotifier.dispose);
       }
 
-      final restartService = BridgeRestartService(
+      restartService = BridgeRestartService(
         processRepository: processRepository,
         commandBuilder: const BridgeRestartCommandBuilder(),
         binaryPath: managedRuntimePaths.binaryPath,
@@ -843,6 +858,21 @@ class BridgeRuntimeRunner {
         );
       }
 
+      // Start the ordered shutdown the moment the session begins shutting
+      // down (any trigger: signal, supervised logout/restart, control-channel
+      // loss), so the coordinator's backstop bounds the session teardown
+      // itself — a teardown blocked on in-flight agent work must not hang the
+      // process with no deadline. The runner's finally joins the same
+      // (memoized) shutdown future, so a failure here is still surfaced there
+      // with the exit-code policy applied.
+      unawaited(
+        // The runner's finally awaits the same (memoized) shutdown future and
+        // applies the exit-code policy there, so the failure is already
+        // surfaced; this early-start copy must not raise an unhandled error.
+        activeRuntime.session.shutdownRequested
+            .then((_) => shutdownCoordinator.shutdown())
+            .catchError((Object _, StackTrace __) {}),
+      );
       registerSignalHandlers(session: activeRuntime.session, subscriptions: subscriptions);
       // start() synchronously subscribes local route-trigger listeners before
       // the debug server can expose mutation routes.
--- a/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
+++ b/bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
@@ -100,7 +100,6 @@ import "../../updater/services/update_lifecycle_service.dart";
 import "../../updater/services/update_reconciliation_service.dart";
 import "../../updater/services/update_service.dart";
 import "../../version.dart";
-import "../debug_server.dart";
 import "../foundation/process_runner.dart";
 import "../foundation/process_runner_command_executor.dart";
 import "../log_failure_reporter.dart";
@@ -188,7 +187,6 @@ class BridgeRuntimeRunner {
     PluginRuntime? pluginRuntime;
     PluginLifecycleService? pluginLifecycleService;
     BridgeRuntime? runtime;
-    DebugServer? debugServer;
     CatalogImportConsoleListener? catalogImportConsoleListener;
     Future<void>? sessionRun;
     // The single typed slot for a deliberate supervised exit (restart /
@@ -204,24 +202,15 @@ class BridgeRuntimeRunner {
     // loss listener assigns with `??=` so a loss never overwrites an already
     // decided intentional exit.
     SupervisedExitCode? requestedSupervisedExit;
-    BridgeRestartService? restartService;
+    // Shared by the ordered pluginDispose phase and the backstop's emergency
+    // disposal so the two cannot drift.
+    Future<void> disposePluginApis() => pluginRuntime?.disposeStartedApis() ?? Future<void>.value();
     final shutdownCoordinator = BridgeShutdownCoordinator(
       startAbortSignal: startAbortController.signal,
-      backstopExitCode: () {
-        final exit = requestedSupervisedExit;
-        if (exit != null) return exit.code;
-        // The supervised-restart respawn sentinel is folded into the slot in
-        // the try's finally, which has not run when the backstop fires
-        // mid-teardown; consult the handoff flag directly so a hung teardown
-        // still reports the sentinel and the GUI respawns.
-        if (restartService?.supervisedRestartRequested ?? false) {
-          return SupervisedExitCode.restart.code;
-        }
-        return 0;
-      },
+      backstopExitCode: () => requestedSupervisedExit?.code ?? 0,
       // Last resort before a forced exit: stop plugin backends so their agent
       // processes are not orphaned when the teardown (or a phase) hangs.
-      emergencyDisposal: () => pluginRuntime?.disposeStartedApis() ?? Future<void>.value(),
+      emergencyDisposal: disposePluginApis,
     );
     shutdownCoordinator
       ..addPhase(
@@ -234,15 +223,22 @@ class BridgeRuntimeRunner {
       )
       ..addPhase(
         phase: BridgeShutdownPhase.signal,
+        // Cancel in-flight agent work (budgeted, best-effort) so the drain
+        // phase is not held open by agent-coupled requests — an in-flight
+        // turn, a resume-load, a lazy agent respawn — while the agent process
+        // is still alive and can answer a cancellation. Signal actions fire in
+        // registration order, so `beginShutdown` above has already fenced new
+        // lease acquisitions, and the phase is awaited before drain begins.
+        action: () => pluginRuntime?.interruptActiveWorkForShutdown() ?? Future<void>.value(),
+        budget: _pluginShutdownBudget,
+      )
+      ..addPhase(
+        phase: BridgeShutdownPhase.signal,
         action: () => runtime?.catalogImportService.beginShutdown(),
       )
       ..addPhase(
         phase: BridgeShutdownPhase.signal,
         action: () => runtime?.session.beginShutdown(),
-      )
-      ..addPhase(
-        phase: BridgeShutdownPhase.signal,
-        action: () => debugServer?.beginShutdown(),
       )
       ..addPhase(
         phase: BridgeShutdownPhase.signal,
@@ -257,12 +253,8 @@ class BridgeRuntimeRunner {
         action: () => sessionRun ?? Future<void>.value(),
       )
       ..addPhase(
-        phase: BridgeShutdownPhase.drain,
-        action: () => debugServer?.drain() ?? Future<void>.value(),
-      )
-      ..addPhase(
         phase: BridgeShutdownPhase.pluginDispose,
-        action: () => pluginRuntime?.disposeStartedApis() ?? Future<void>.value(),
+        action: disposePluginApis,
         budget: _pluginShutdownBudget,
       )
       ..addPhase(
@@ -777,7 +769,7 @@ class BridgeRuntimeRunner {
         shutdownCoordinator.add(disposable: controlStatusNotifier.dispose);
       }
 
-      restartService = BridgeRestartService(
+      final restartService = BridgeRestartService(
         processRepository: processRepository,
         commandBuilder: const BridgeRestartCommandBuilder(),
         binaryPath: managedRuntimePaths.binaryPath,
@@ -787,6 +779,12 @@ class BridgeRuntimeRunner {
         // exits with the sentinel code instead of spawning a successor (which
         // would replay --control-url with no off-argv secret and fail closed).
         isSupervised: options.isSupervised,
+        // Record the GUI-respawn sentinel the moment the handoff is decided —
+        // before the shutdown it triggers — so the normal return, the error
+        // paths, and a hung-teardown backstop all report the same code.
+        onSupervisedRestartRequested: () {
+          requestedSupervisedExit = SupervisedExitCode.restart;
+        },
       );
 
       // Run startup diagnostics before composing the runtime so the
@@ -864,28 +862,10 @@ class BridgeRuntimeRunner {
       final sessionStart = activeRuntime.session.start();
       sessionStart.ignore();
       sessionRun = activeRuntime.session.waitUntilStopped();
-      debugServer = await startDebugServerIfRequested(
+      await startDebugServerIfRequested(
         debugPort: options.debugPort,
         runtime: activeRuntime,
         shutdownCoordinator: shutdownCoordinator,
-      );
-      // Start the ordered shutdown the moment the session begins shutting down
-      // (any trigger: signal, supervised logout/restart, control-channel loss),
-      // so the coordinator's backstop bounds the session teardown itself — a
-      // teardown blocked on in-flight agent work must not hang the process
-      // with no deadline. Registered only after the debug server is assigned:
-      // the coordinator's signal phase reads `debugServer` (beginShutdown), so
-      // a shutdown racing the async server start would otherwise miss it, and
-      // no agent-coupled work can exist before startup completes anyway. The
-      // runner's finally joins the same (memoized) shutdown future, so a
-      // failure here is still surfaced there with the exit-code policy applied.
-      unawaited(
-        // The runner's finally awaits the same (memoized) shutdown future and
-        // applies the exit-code policy there, so the failure is already
-        // surfaced; this early-start copy must not raise an unhandled error.
-        activeRuntime.session.shutdownRequested
-            .then((_) => shutdownCoordinator.shutdown())
-            .catchError((Object _, StackTrace __) {}),
       );
       // Background: check + download + stage + apply-in-place on a 4h cadence.
       // The swap takes effect on the next launch (or a phone-triggered restart).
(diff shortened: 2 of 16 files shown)


bridge/app/test/bridge/runtime/plugin_runtime_test.dart
```diff
@@ -499,6 +499,62 @@ void main() {
...
+      startGate: Future<void>.value(),
+      pluginFactory: (_) => plugin = _FakePlugin(api: api),
+    );
+    final runtime = _runtime(factory: factory);
+    addTearDown(runtime.dispose);
+    await runtime.start(pluginId: "one");
+    plugin.interruptActiveWorkHandler = (_) => Completer<Set<String>>().future;
+
+    await runtime.interruptActiveWorkForShutdown();
+
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit cbb5a992)

Made the test helper _runtime(...) accept an injectable shutdownBudget and gave the hung-plugin test a 20ms budget instead of the default 1s, so the timeout path is still exercised (interrupt call count asserted) without a real one-second wait or real-timer flakiness.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -843,6 +858,21 @@ class BridgeRuntimeRunner {
...
       }
 
+      // Start the ordered shutdown the moment the session begins shutting
+      // down (any trigger: signal, supervised logout/restart, control-channel
+      // loss), so the coordinator's backstop bounds the session teardown
+      // itself — a teardown blocked on in-flight agent work must not hang the
+      // process with no deadline. The runner's finally joins the same
+      // (memoized) shutdown future, so a failure here is still surfaced there
+      // with the exit-code policy applied.
+      unawaited(
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit cbb5a992)

Moved the early coordinator registration to after the debug server is assigned (post `debugServer = await startDebugServerIfRequested(...)`). The coordinator's signal phase reads `debugServer` for beginShutdown, so a shutdown racing the async server start previously evaluated the null variable and never retried. Deferring is safe: no agent-coupled work can exist before startup completes, so a signal in that window falls back to the pre-existing finally-started coordinator.

bridge/app/test/bridge/runtime/plugin_runtime_test.dart
```diff
@@ -499,6 +499,62 @@ void main() {
...
+    final api = _FakeApi();
+    late _FakePlugin plugin;
+    final factory = _FakeGenerationFactory(
+      startGate: Future<void>.value(),
+      pluginFactory: (_) => plugin = _FakePlugin(api: api),
+    );
+    final runtime = _runtime(factory: factory);
+    addTearDown(runtime.dispose);
+    await runtime.start(pluginId: "one");
+    plugin.interruptActiveWorkHandler = (_) => Completer<Set<String>>().future;
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed (in commit cbb5a992)

Same as the related qodo thread: the hung-plugin interrupt test now runs with a 20ms shutdown budget via the injectable _runtime shutdownBudget, removing the real 1s wait and the timing dependence.

bridge/app/lib/src/bridge/runtime/bridge_shutdown_coordinator.dart
```diff
@@ -17,16 +17,22 @@ class BridgeShutdownCoordinator {
...
        _exitProcess = exitProcess ?? io.exit,
-       _startAbortSignal = startAbortSignal;
+       _startAbortSignal = startAbortSignal,
+       _emergencyDisposal = emergencyDisposal;
 
   static int _alwaysZero() => 0;
   static const Duration _backstopSlack = Duration(seconds: 10);
 
+  /// How long the backstop waits for [emergencyDisposal] before exiting.
+  static const Duration _emergencyDisposalCap = Duration(seconds: 3);
```
daniil-shumko: **The 3s cap can't contain the ACP client's SIGTERM→SIGKILL escalation (5s).**

`AcpStdioClient._teardownConnection` sends SIGTERM, waits `exitCode` up to 5s (`gracefulTimeout`), then SIGKILLs. With a 3s cap the backstop exits the bridge before the SIGKILL escalation can ever fire, so an agent that ignores SIGTERM is still orphaned — the exact outcome this hook exists to prevent. After a ~40s backstop, three more seconds is cheap.

Suggestion: raise the cap to ≥6s so the full escalation fits, or add a comment stating the truncation is a deliberate trade. The two constants live in different packages with nothing tying them together, so they will silently drift otherwise.

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
```diff
@@ -587,6 +587,37 @@ class PluginRuntime {
...
+  /// ongoing session teardown drains whatever events the cancellation
+  /// produces. Never throws; isolates and logs per-plugin failures.
+  Future<void> interruptActiveWorkForShutdown() {
+    return Future.wait([
+      for (final slot in _slots.values)
+        if (slot.plugin case final plugin?) _interruptPluginForShutdown(slot, plugin),
+    ]);
+  }
+
+  Future<void> _interruptPluginForShutdown(_PluginRuntimeSlot slot, BridgePlugin plugin) async {
```
daniil-shumko: **Convention: named `required` parameters.**

`_interruptPluginForShutdown(_PluginRuntimeSlot slot, BridgePlugin plugin)` uses positional parameters. The project rule — and this file's own siblings `_interruptActiveWork` and `_snapshotActiveSessionIds` — use named `required` parameters.

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
```diff
@@ -587,6 +587,37 @@ class PluginRuntime {
...
+    try {
+      await plugin.interruptActiveWork(budget: _shutdownBudget).timeout(_shutdownBudget);
```
daniil-shumko, line 617: **The `interruptActiveWork` interface contract is now stale at this call site.**

`BridgePlugin.interruptActiveWork`'s doc (`bridge_plugin.dart`) states: "The runtime has already fenced new acquisitions and drained pre-fence operations when this is called." The forced-stop path establishes that via `_drainForcedStopBarrier`; this new call site deliberately doesn't — interrupting still-in-flight operations is its whole purpose. All three current implementations tolerate it, but this is the contract future plugin authors code against. Suggest a one-sentence amendment in the interface doc describing both invocation modes (barrier-fenced force-stop vs best-effort shutdown cancel).

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -854,6 +869,24 @@ class BridgeRuntimeRunner {
...
+      // with no deadline. Registered only after the debug server is assigned:
+      // the coordinator's signal phase reads `debugServer` (beginShutdown), so
+      // a shutdown racing the async server start would otherwise miss it, and
+      // no agent-coupled work can exist before startup completes anyway. The
+      // runner's finally joins the same (memoized) shutdown future, so a
+      // failure here is still surfaced there with the exit-code policy applied.
+      unawaited(
+        // The runner's finally awaits the same (memoized) shutdown future and
+        // applies the exit-code policy there, so the failure is already
+        // surfaced; this early-start copy must not raise an unhandled error.
```
daniil-shumko: **Duplicated justification** — this inner comment restates the closing sentence of the block comment above ("the runner's finally joins the same (memoized) shutdown future, so a failure here is still surfaced there"). One copy suffices. Nit.

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -854,6 +869,24 @@ class BridgeRuntimeRunner {
...
+        activeRuntime.session.shutdownRequested
+            .then((_) => shutdownCoordinator.shutdown())
+            .catchError((Object _, StackTrace __) {}),
```
daniil-shumko: **`.ignore()` is the established idiom for this** — `sessionStart.ignore()` twenty lines up does exactly this (marks errors handled without a handler), and the orchestrator uses it too (`trackedCompletion.ignore()`).

```dart
activeRuntime.session.shutdownRequested
    .then((_) => shutdownCoordinator.shutdown())
    .ignore();
```

also drops the `unawaited(…)` wrapper. Nit.

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

bridge/app/lib/src/bridge/orchestrator.dart
```diff
@@ -1260,6 +1265,11 @@ class OrchestratorSession {
         "[shutdown] cancel() requested"
         "${_inFlightRelayWorkDiagnostic(separator: " — ")}",
       );
+      // Interrupt in-flight agent work now (budgeted, best-effort) so the
+      // teardown drains below are not held open by agent-coupled requests
+      // (an in-flight turn, a resume-load, a lazy agent respawn) while the
+      // agent process is still alive and can answer a cancellation.
+      unawaited(_pluginRuntime.interruptActiveWorkForShutdown());
```
daniil-shumko: **Optional: this wiring is the one behavioral seam left untested.**

A small test that `beginShutdown()` fires the interrupt exactly once (and not again on repeat cancels) looks feasible with the existing `OrchestratorSession` test harnesses (`orchestrator_error_recovery_test.dart` et al.) — unlike the runner wiring, which is composition-root-by-convention untested. Not blocking.

<sub>Found by Claude Fable 5 (`claude-fable-5`) · effort: max</sub>

daniil-shumko: **Main structural point: this hook is redundant after change 2.**

Before the coordinator started at cancel time, firing the interrupt from inside `beginShutdown()` was the only way to get it to happen early. Now that `shutdownRequested` starts the coordinator, this could simply be a signal-phase action next to the `pluginRuntime?.beginShutdown()` that's already registered at `bridge_runtime_runner.dart:231-234`. That version is better on four counts:

1. **Ordering stays visible in one place.** A reader of the phase list sees `pluginRuntime.beginShutdown()` but not the interrupt — it's hidden two files away, inside a method that is itself a phase action.
2. **Awaited instead of raced.** Today it runs concurrently with the drain phase and with `pluginDispose`; if dispose wins, the in-flight `interruptActiveWork` is talking to a process being SIGTERM'd and logs a warning.
3. **Fixes an order inconsistency.** Via SIGINT this runs *before* `pluginRuntime.beginShutdown()` sets `_shuttingDown`, so new lease acquisitions are not yet fenced. Via supervised logout / control-channel loss (`_shutdownThenExit` → coordinator) the fence runs first. Same operation, two different preconditions depending on trigger.
4. **Satisfies the documented plugin contract.** `BridgePlugin.interruptActiveWork` (`bridge_plugin.dart:49-56`) states the runtime has already fenced acquisitions and drained pre-fence operations when it is called. Neither holds here.

Trade-off, fairly stated: awaiting it in the signal phase can add up to `_pluginShutdownBudget` (10s) before drain for a plugin that ignores its budget. Since the drain is blocked on exactly that work anyway that isn't a real cost, and a shorter dedicated interrupt budget would be reasonable.

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
```diff
@@ -587,6 +587,37 @@ class PluginRuntime {
...
+  /// Uses the same per-plugin `interruptActiveWork` contract as the forced-
+  /// stop path ([_interruptActiveWork]) but emits no terminal handoff: the
+  /// ongoing session teardown drains whatever events the cancellation
+  /// produces. Never throws; isolates and logs per-plugin failures.
+  Future<void> interruptActiveWorkForShutdown() {
+    return Future.wait([
+      for (final slot in _slots.values)
+        if (slot.plugin case final plugin?) _interruptPluginForShutdown(slot, plugin),
+    ]);
+  }
```
daniil-shumko, line 610: The doc says this "uses the same per-plugin `interruptActiveWork` contract as the forced-stop path," but it also relaxes that contract in a way worth calling out: the interface doc (`bridge_plugin.dart:49-56`) promises implementors that *"the runtime has already fenced new acquisitions and drained pre-fence operations when this is called"*, and neither is true here — `_shuttingDown` isn't set yet on the SIGINT path, and there is deliberately no barrier drain.

Harmless for ACP/Codex/OpenCode (all recompute active sessions from live state on entry, and iterate a snapshot set, so a concurrently-arriving prompt can't corrupt them). But a future plugin author reading that doc could reasonably assume no new work can arrive concurrently. Either state the weaker precondition on the interface, or mention it here.

Also, minor: the closing sentence — "the ongoing session teardown drains whatever events the cancellation produces" — is a bit optimistic. `_teardown()` cancels `_subscriptions` (including the plugin-event subscription) as its very first step, so whether those events land is a race. Doesn't matter in practice (we're going away, and no terminal handoff is expected), but the comment reads as a guarantee.

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
daniil-shumko: Positional parameters. AGENTS.md: *"Use named parameters with `required`, including nullable parameters"* — the only exception being the primary message argument of logging APIs.

This is also the only two-positional-arg method in the file; its immediate neighbour `_snapshotActiveSessionIds({required slot, required plugin})` and `_interruptActiveWork({required slot, required plugin, ...})` both use named.

```suggestion
  Future<void> _interruptPluginForShutdown({
    required _PluginRuntimeSlot slot,
    required BridgePlugin plugin,
  }) async {
```

*(Opus 5 · effort: max)*

daniil-shumko (commented):
Deep review of the full shutdown path (orchestrator → runner → coordinator → plugin runtime → all three plugin implementations). Verdict: approve with minor findings — the design is genuinely elegant where it matters: it reuses three existing seams (the pre-existing `_shutdownCompleter`, the `interruptActiveWork` plugin contract, the coordinator's memoized `shutdown()`) instead of inventing machinery, and the claimed invariants (drain → pluginDispose order, reentrancy across the three shutdown triggers, sentinel precedence, cancellation events flowing during the drain window via still-routable slots) all hold under close reading. Both changed test suites pass locally (68/68).

Inline: 1 low-severity design gap (3s emergency cap vs 5s SIGKILL escalation), 1 convention break (positional params), 1 stale interface doc, 2 nits, 1 optional test suggestion.

Two notes without inline anchors:

- `interruptActiveWork(budget: X).timeout(X)` can log a spurious "could not interrupt" warning when the outer timer wins the equality race against the plugin's internal `.timeout(budget)` — but it mirrors the forced-stop idiom (`plugin_runtime.dart` line 1144) exactly, so no change requested.
- The PR body lists analyzer/suite/architecture verification but no live Ctrl+C-during-turn run. Given the symptom is wall-clock behavior with a real agent, recommend one live repro (Ctrl+C mid-Cursor-turn → bridge exits in seconds, agent process reaped) before merge if it hasn't happened.

<sub>Review by Claude Fable 5 (`claude-fable-5`), effort level: max.</sub>

daniil-shumko (commented):
## Deep review — design/elegance pass

Diagnosis is right and the fix is well-targeted: 200 lines, no new abstractions, three changes each closing a distinct hole. I found **no correctness bug**. Verified locally: `dart analyze --fatal-infos lib test` clean; `bridge_shutdown_coordinator_test.dart` + `plugin_runtime_test.dart` pass (68 tests). Also checked and found fine: `shutdown()` memoization shared by the early start and the `finally`; the session error surfacing twice (pre-existing — the drain phase already awaited the same `sessionRun`); every `shutdownCoordinator.add(...)` registration completing before the first `await` that can let the early start run, so no action is silently skipped today; `disposeStartedApis()` memoization making the emergency call safe when `pluginDispose` already ran; `exit(exitCode)` immediately after `runBridgeApp` so the unawaited 10s interrupt timer can't delay process exit.

The elegance criticism is one structural point plus a handful of small ones (inline).

**Changes 1 and 2 overlap.** Once the coordinator starts at cancel time (change 2), the interrupt hook (change 1) no longer needs to live inside `OrchestratorSession` — it could be a coordinator signal-phase action. Details on the `orchestrator.dart` comment; that's the finding I'd act on first.

**Two cross-cutting notes that don't anchor to a diff line:**

1. `BridgePlugin.interruptActiveWork` (`bridge_plugin.dart:49-56`) still documents the precondition *"The runtime has already fenced new acquisitions and drained pre-fence operations when this is called."* The new call site honours neither. ACP/Codex/OpenCode all recompute from live state so they're fine today, but this is a plugin-boundary contract — if the weaker precondition is intended, the interface doc should say so.

2. **Backstop arithmetic.** Phases contribute `pluginDispose 10s + lifecycle 10s = 20s`, `+10s` slack → the backstop fires at **30s**, not the ~40s the PR body states (nothing else registers a budget — no production `addOrdered` callers). More interesting: `drain`, now the phase most likely to hang, contributes **0** to the computed budget, so the bound is incidental rather than declared. Giving drain an explicit budget would make the number honest and self-documenting.

---
*Review produced with **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`) at **effort level: max**.*

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -854,6 +869,24 @@ class BridgeRuntimeRunner {
...
+      // Start the ordered shutdown the moment the session begins shutting down
+      // (any trigger: signal, supervised logout/restart, control-channel loss),
+      // so the coordinator's backstop bounds the session teardown itself — a
+      // teardown blocked on in-flight agent work must not hang the process
+      // with no deadline. Registered only after the debug server is assigned:
+      // the coordinator's signal phase reads `debugServer` (beginShutdown), so
+      // a shutdown racing the async server start would otherwise miss it, and
+      // no agent-coupled work can exist before startup completes anyway. The
+      // runner's finally joins the same (memoized) shutdown future, so a
+      // failure here is still surfaced there with the exit-code policy applied.
```
daniil-shumko: **The positional coupling this comment documents can be removed entirely.**

cbb5a992 fixed a real race, but by *moving a statement* — the early start must now sit after every mutable local the phase closures read, and nothing enforces that. The next person to add a `shutdownCoordinator.add(...)` below this line gets a silently-skipped disposable.

`debugServer` is referenced only by the two phase closures at `:245` and `:261`. Move both registrations into `startDebugServerIfRequested` (`bridge_runtime.dart:88`, which already registers `bandwidthTracker.dispose` and only registers when a server actually exists) and the outer `DebugServer? debugServer` local disappears along with this 10-line rationale. Signal-phase actions are fire-all-then-await-all, so relative order within the phase doesn't change anything.

That also makes the early start position-independent, so it could move up next to `session.start()` and close the startup window this comment currently has to argue away.

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -854,6 +869,24 @@ class BridgeRuntimeRunner {
...
+      unawaited(
+        // The runner's finally awaits the same (memoized) shutdown future and
+        // applies the exit-code policy there, so the failure is already
+        // surfaced; this early-start copy must not raise an unhandled error.
+        activeRuntime.session.shutdownRequested
+            .then((_) => shutdownCoordinator.shutdown())
+            .catchError((Object _, StackTrace __) {}),
+      );
```
daniil-shumko: Two nits on this block:

1. `unawaited(f.then(...).catchError((_, __) {}))` is exactly `f.then(...).ignore()` — precise intent, and the idiom already used 17 lines above (`sessionStart.ignore()`) and in `orchestrator.dart` (`trackedCompletion.ignore()`).

2. The comment inside the `unawaited(` and the last three lines of the comment above it say the same thing (the runner's `finally` joins the same memoized future, so this copy must not raise unhandled). One of them can go.

```suggestion
      activeRuntime.session.shutdownRequested.then((_) => shutdownCoordinator.shutdown()).ignore();
```

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -204,9 +204,24 @@ class BridgeRuntimeRunner {
...
+        if (exit != null) return exit.code;
+        // The supervised-restart respawn sentinel is folded into the slot in
+        // the try's finally, which has not run when the backstop fires
+        // mid-teardown; consult the handoff flag directly so a hung teardown
+        // still reports the sentinel and the GUI respawns.
+        if (restartService?.supervisedRestartRequested ?? false) {
+          return SupervisedExitCode.restart.code;
+        }
+        return 0;
+      },
```
daniil-shumko: This is the trickiest new decision in the PR (sentinel survival when the backstop fires mid-teardown) and it has no test. It's nearly pure — extracting it would make it coverable in `bridge_runtime_runner_test.dart`, which already tests two static helpers exactly this way:

```dart
@visibleForTesting
static int resolveBackstopExitCode({
  required SupervisedExitCode? requested,
  required bool supervisedRestartRequested,
}) => ...
```

Deeper point: the restart outcome is the *one* sentinel that isn't recorded at decision time, even though `:196-205` states that's the slot's whole purpose ("Set at the moment an outcome is decided — BEFORE any shutdown runs"). It's resolved in the inner `finally` instead, which is why the backstop now needs a second source of truth. Folding it into the slot when `performRestartHandoff()` sets `_supervisedRestartRequested` would delete both this special case and the hoisted nullable local — and `restartService.supervisedRestartRequested` at `:948` would stop relying on flow promotion of a closure-captured nullable.

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
```diff
@@ -204,9 +204,24 @@ class BridgeRuntimeRunner {
...
+      // Last resort before a forced exit: stop plugin backends so their agent
+      // processes are not orphaned when the teardown (or a phase) hangs.
+      emergencyDisposal: () => pluginRuntime?.disposeStartedApis() ?? Future<void>.value(),
```
daniil-shumko: `() => pluginRuntime?.disposeStartedApis() ?? Future<void>.value()` is byte-identical to the `pluginDispose` phase action at `:265`. Worth a single local so the two can't drift:

```dart
Future<void> disposePluginApis() => pluginRuntime?.disposeStartedApis() ?? Future<void>.value();
```

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/runtime/bridge_shutdown_coordinator.dart
daniil-shumko: **3s undercuts the escalation this hook exists to complete.**

ACP's teardown is SIGTERM → wait 5s → SIGKILL (`acp_stdio_client.dart:368, 414-425`). SIGTERM goes out in the first millisecond (the `_disposeApi` calls are in the first `Future.wait` alongside the start futures), so a well-behaved agent is reaped fine. An agent that *ignores* SIGTERM — the likelier one here, since we only reach the backstop because something is stuck — is abandoned at 3s, before the SIGKILL at 5s. Orphaned, which is the exact outcome the PR adds this for.

Sizing the cap above the plugin escalation window (~6s) closes it; adding 3s to a 30s force-exit is already accepted, and 6s isn't materially worse. If 3s is deliberate, the doc comment should say the escalation may be truncated.

*(Opus 5 · effort: max)*

bridge/app/lib/src/bridge/orchestrator.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Moved the interrupt out of `beginShutdown()` into a coordinator signal-phase action registered directly after `pluginRuntime?.beginShutdown()`, with `budget: _pluginShutdownBudget` so the backstop is sized honestly. That lands all four counts: the ordering is visible in the phase list, the interrupt is awaited before the drain phase instead of racing `pluginDispose`, every trigger now fences lease acquisitions before the interrupt starts (signal actions fire in registration order), and the call site matches the amended interface contract. On the trade-off: kept the full `_pluginShutdownBudget` rather than a shorter dedicated interrupt budget — as noted, the drain is blocked on exactly that work anyway, and a second budget constant would be machinery without a demonstrated need.

daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

The wiring this asked to test no longer exists: per the structural thread on this line, `beginShutdown()` no longer fires the interrupt — it is now a coordinator signal-phase action. The exactly-once property is the coordinator's memoized `shutdown()` (covered by the existing "repeated shutdown calls share one run" test), and the new "an async signal action is awaited before the drain phase starts" test covers the interrupt-before-drain ordering the move relies on. The remaining runner registration is composition-root-by-convention untested, same as the rest of the phase list.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Implemented the deeper fold rather than the extraction: `BridgeRestartService` now takes a required `onSupervisedRestartRequested` callback and invokes it at the moment `performRestartHandoff()` decides the supervised handoff — the runner records `SupervisedExitCode.restart` into the slot right there, before any shutdown runs, matching the slot's documented purpose. That deleted the backstop special case (back to `requestedSupervisedExit?.code ?? 0`), the hoisted nullable `restartService` local, the inner try/finally fold, and the flow-promotion read. The pull-side `supervisedRestartRequested` getter is gone too (the callback replaced it; service tests now assert the notification, and the four test fakes dropped their overrides). With the decision folded into the slot there is no backstop decision left to extract or test.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Extracted a `disposePluginApis` local; both the `emergencyDisposal` argument and the `pluginDispose` phase action now route through it.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Moved both registrations into `startDebugServerIfRequested` (registered once the server is listening, capturing the non-null server), which deletes the outer `DebugServer? debugServer` local, the 10-line rationale, and the runner's `debug_server.dart` import; the helper now returns `void`. One deliberate deviation: the early start did not move up next to `session.start()`. Its drain closure reads `sessionRun` (assigned two lines below `start()`), and more importantly a coordinator run during `HttpServer.bind` would execute the signal/drain phases before the helper's registrations land, leaving a just-started server with dead registrations — closing that fully would need `DebugServer.start()` to re-check shutdown state after bind, which is machinery for a SIGINT-during-loopback-bind window with no meaningful consequence. Instead the hook is now registered after all startup wiring (it previously sat above the `updateLifecycle.dispose` registration, which had the exact silently-skipped-disposable hazard this comment predicted), with the positional argument reduced to "after startup completes".

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

The block was rewritten as part of moving the early start after startup wiring; the justification now appears once, in the block comment, and the inner copy is gone (the statement also switched to the `.ignore()` idiom per the sibling threads).

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Switched to `.then(...).ignore()`, dropping the `unawaited(...)` wrapper and the `catchError` no-op, matching `sessionStart.ignore()` and the orchestrator usage.

bridge/app/lib/src/bridge/runtime/bridge_runtime_runner.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Both nits taken: the statement is now `activeRuntime.session.shutdownRequested.then((_) => shutdownCoordinator.shutdown()).ignore();` and the duplicated inner comment is gone — the finally-joins rationale appears once, in the block comment above.

bridge/app/lib/src/bridge/runtime/bridge_shutdown_coordinator.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Raised `_emergencyDisposalCap` to 6s so the ACP SIGTERM → 5s wait → SIGKILL escalation completes before the forced exit, and documented the sizing relationship on the constant (the two values live in different packages, so the comment is the tie — the bridge core must not import plugin-specific constants). The coordinator test now asserts exit has not happened 5s into the disposal and fires at 6s.

daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Cap raised to 6s (above the 5s SIGTERM→SIGKILL escalation window) with the rationale documented on `_emergencyDisposalCap`; the "capped" coordinator test asserts the full kill window fits inside the cap.

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Three parts: (1) the interface doc on `BridgePlugin.interruptActiveWork` now describes both invocation modes — the barrier-fenced forced stop and the best-effort bridge-shutdown cancel where pre-fence operations are deliberately not drained and implementations must tolerate concurrent in-flight work. (2) The fence half of the concern is now real rather than documented-away: with the interrupt moved into the coordinator signal phase after `pluginRuntime.beginShutdown()`, `_shuttingDown` is set before the interrupt on every trigger, SIGINT included. (3) The optimistic closing sentence is gone — the method doc now says cancellation events reach the concurrently tearing-down session only on a best-effort basis.

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

`BridgePlugin.interruptActiveWork` now documents the two invocation modes: forced generation stop (acquisitions fenced, pre-fence operations drained, returned IDs feed the terminal handoff) and bridge shutdown (best-effort cancel with teardown running concurrently — fenced acquisitions, deliberately no pre-fence drain, return value ignored), plus the requirement that implementations tolerate concurrently in-flight operations.

bridge/app/lib/src/bridge/runtime/plugin_runtime.dart
daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Converted `_interruptPluginForShutdown` to named `required` parameters, matching `_interruptActiveWork` and `_snapshotActiveSessionIds`.

daniil-shumko: [Sesori reply] Addressed (in commit 19a49015)

Applied as suggested — `_interruptPluginForShutdown({required _PluginRuntimeSlot slot, required BridgePlugin plugin})`, caller updated.

PR #759, opened 2026-08-05
🌿 Present harness settings per entry point

daniil-shumko:
## Complexity

Straightforward: one new route parameter threaded through the router and the
screen's top bar.

## What

Harness settings now present differently depending on where they are opened
from, instead of always being a modal:

- **Settings list** — pushes onto the settings stack as a plain `MaterialPage`
  and returns with the back chevron. No close button.
- **New-session harness menu** — rises as a modal
  (`CupertinoPage(fullscreenDialog: true)`) and closes with the X. No back
  chevron.

`AppRoute.settingsHarnesses` carries a `HarnessSettingsPresentation`
(`pushed` | `modal`), encoded as a `presentation` query parameter, so the
router picks the page type and the screen picks its dismissal affordance from
the same decoded route.

## Why

Previously both entry points shared one fullscreen-dialog page that showed a
back chevron and an X side by side — two dismissals in one bar, and the wrong
motion when the page is simply the next step in the settings stack. From the
settings list it belongs in that stack; from the new-session screen it is a
detour from an unrelated screen and a modal is right.

## Risk and test focus

Low. Presentation-only; no database, wire, or backend impact. The close
action itself is unchanged, including the fallback to Projects when nothing
sits underneath.

A URL that names no presentation — a deep link or a hand-typed address —
decodes as `modal`, so it always keeps a working way out rather than a back
chevron with nothing behind it.

Test focus: opening harness settings from both entry points, and dismissing
from each.

## Expected result

From Settings → Harnesses: slides in sideways, back chevron only. From the
new-session harness picker: slides up, X only, closing returns to the new
session screen.

## Verification

- New widget tests: each entry point shows exactly one dismissal affordance,
  and the pushed page's back button returns to the opener.
- Updated route round-trip tests for the `presentation` query parameter,
  including the unknown/absent-value fallback to `modal`.
- `flutter test` green for the harness settings, settings, new-session, and
  routing suites; `flutter analyze` clean for `app` and `module_core`.

### DIFF
--- a/client/app/lib/core/routing/app_router.dart
+++ b/client/app/lib/core/routing/app_router.dart
@@ -83,13 +83,21 @@ extension AppRouteToGoRoute on AppRouteDef {
   }
 
   Widget _buildScreen({required BuildContext context, required GoRouterState state}) {
-    final route = AppRoute.fromDef(
+    return AppRoute.fromDef(
       def: this,
       pathParams: state.pathParameters,
       queryParams: state.uri.queryParameters,
-    );
+    ).screen;
+  }
+}
 
-    return switch (route) {
+extension on AppRoute {
+  /// The screen this decoded route shows. Separate from [AppRouteToGoRoute._buildScreen]
+  /// so a page builder that already resolved its route — to choose a page type
+  /// from the route's own parameters — builds the screen from that same
+  /// instance instead of decoding the URL a second time.
+  Widget get screen {
+    return switch (this) {
       AppRouteSplash() => const SplashScreen(),
       AppRouteLogin() => const LoginScreen(),
       AppRouteProjects() => const ProjectListScreen(),
@@ -396,8 +404,8 @@ List<RouteBase> _buildAppRoutes({
           // Android too. The pushed page is the MaterialPage go_router would
           // have built itself, so it keeps the platform's push transition.
           pageBuilder: (context, state) {
-            final child = AppRouteDef.settingsHarnesses._buildScreen(context: context, state: state);
             final route = AppRouteSettingsHarnesses.fromParams(queryParams: state.uri.queryParameters);
+            final child = route.screen;
             return switch (route.presentation) {
               HarnessSettingsPresentation.modal => CupertinoPage<void>(
                 key: state.pageKey,
--- a/client/app/lib/core/routing/app_router.dart
+++ b/client/app/lib/core/routing/app_router.dart
@@ -95,7 +95,7 @@ extension AppRouteToGoRoute on AppRouteDef {
       AppRouteProjects() => const ProjectListScreen(),
       AppRouteSettings() => const SettingsScreen(),
       AppRouteSettingsNotifications() => const NotificationSettingsScreen(),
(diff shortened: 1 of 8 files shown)


client/app/lib/features/settings/harnesses_settings_screen.dart
```diff
@@ -48,6 +48,10 @@ class _HarnessesSettingsBody extends StatelessWidget {
         title: loc.settingsHarnessesTitle,
         titleMode: PregoTopNavigationTitleMode.inline,
         banner: ConnectionBanner.maybeFor(context),
+        // The page always rises as a modal, so closing is the only way out and
```
daniil-shumko: This is not right. When we navigate to the harness settings page through the app's settings page, it should be as a normal push navigation into the navigation stack and have a back button.

when we navigate to it from the new session screen then it is shown as a modal and with only the X close button on the right

daniil-shumko: [Sesori reply] Addressed (in commit 311535c3)

The route now carries the presentation instead of hard-coding one. `AppRoute.settingsHarnesses` takes a `HarnessSettingsPresentation` (`pushed` | `modal`), encoded as a query param so it survives a deep link:

- **From the settings list** — `pushed`: a plain `MaterialPage`, so it slides in like the other settings sub-pages and returns with the back chevron. No X.
- **From the new-session harness menu** — `modal`: the `CupertinoPage(fullscreenDialog: true)` as before, with only the X on the right. No back chevron.

A URL that names no presentation (deep link, hand-typed) decodes as `modal`, so it always has a way out rather than a back button with nothing behind it.

Covered by two new widget tests asserting each entry point shows exactly one dismissal affordance, plus round-trip tests for the new query param.

client/app/lib/core/routing/app_router.dart
```diff
@@ -385,16 +385,28 @@ List<RouteBase> _buildAppRoutes({
...
+          // new-session harness menu they are a detour from an unrelated
+          // screen, so they rise as a modal and close back onto it.
+          //
+          // The modal is a CupertinoPage for the same reason settings itself
+          // uses one: only the Cupertino route honours `fullscreenDialog` on
+          // Android too. The pushed page is the MaterialPage go_router would
+          // have built itself, so it keeps the platform's push transition.
+          pageBuilder: (context, state) {
+            final child = AppRouteDef.settingsHarnesses._buildScreen(context: context, state: state);
+            final route = AppRouteSettingsHarnesses.fromParams(queryParams: state.uri.queryParameters);
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Addressed (in commit 368215b0)

Valid — the presentation was parsed once for the page type and again inside `_buildScreen` for the child.

The route-to-screen switch now lives on the decoded route itself (`extension on AppRoute { Widget get screen }`), so the harness `pageBuilder` decodes once and derives both the page type and the child from that single instance. `_buildScreen` keeps its signature for the other routes and just delegates, so the mapping stays in one place.

PR #844, opened 2026-08-12
🌿 feat(notifications): send the device id when registering a push token

stanvladut:
Completes the client half of per-device notification filtering. Without this, the server-side filtering that already shipped cannot apply at all.

## The bug this fixes

Notification settings are stored **per device**, keyed by the `deviceId` this app already generates and sends to `/auth/settings/{deviceId}`. Push tokens were registered with only `{ token, platform }`.

So the server had two records describing the same phone and no shared key:

```
settingsConfiguration  { userId, deviceId: "de250236…", connectionStatus: false }   ← the opt-out
deviceTokens           { userId, token: "fMEp3x9…",     deviceId: null }            ← no link
```

Firebase is addressed per token — the server builds one message per token — so the send/skip decision is per token. With no `deviceId` on the token row it cannot tell which device it is, and falls back to delivering. **Result: a user who turned "Connection Status" off still received "Bridge Disconnected".** That was reproduced against a real account.

This sends the same `deviceId` on `register-token`, which is the only missing half of the join.

## Two details that matter

**A device id that cannot be read does not fail the registration.** `getOrCreate()` touches secure storage. If it throws, the token is registered without the id and logged. Losing the id costs *filtering* for that install; failing the call costs *push entirely*, which is strictly worse.

**A null id is omitted from the body, not serialized as `null`.** The server validates `deviceId` with an optional schema, which accepts an absent key but rejects an explicit null with a 400 — that would break every registration. Enforced with `@JsonKey(includeIfNull: false)` and covered by a test asserting the key is absent.

`deviceId` is `required String?` to match the codebase convention for nullable named parameters and satisfy `prefer_required_named_parameters`, so every call site states its intent.

## Verification

Ran the same steps as `mobile-ci.yml`:

- `dart analyze --fatal-infos` — clean on `client/app`, `module_core`, `module_auth`, `module_prego`
- `dart test` in `module_core` — **1068 passed**
- `flutter test` in `client/app` — **936 passed**

New coverage: the stored id is sent; a storage failure still registers the token; the key is omitted when there is none; `unregisterToken` never touches storage.

Generated sources were rebuilt with `build_runner` under the pinned Flutter 3.44.8-stable, not hand-edited. The only regenerated file beyond the model is `injection.config.dart`, because the repository constructor gained a dependency.

## Rollout

This is step 2 of three. After it ships and the install base rolls over, the server flag `AUTH_REQUIRE_DEVICE_ID_IN_TOKEN_REGISTRATION` can be flipped so a registration without a `deviceId` is rejected outright. Until then the server keeps failing open, so there is no ordering constraint between this and the server deploy.

---
## Summary by cubic
Registers push tokens with the device ID so the server can apply per‑device notification filtering. Previously tokens were registered without a device ID and disabled categories still delivered on that device; now tokens link to the device’s preferences. If the device ID can’t be read, registration proceeds unfiltered.

- Adds `deviceId` to `RegisterTokenRequest`; `module_core`’s `build.yaml` sets `include_if_null: false`, so the key is omitted when null to satisfy the server’s optional schema.
- `NotificationRepository` reads the ID from `NotificationPreferencesDeviceIdStorage`; on read failure it logs and sends the token without `deviceId`. DI updated in `injection.config.dart`.
- Tests cover sending the ID, omitting the key when absent, failure fallback, and that `unregisterToken` does not touch storage.
- Updates `docs/regression/notifications.md` to record device‑scoped filtering and coverage expectations.

<sup>Written for commit 18f0ad6ee8b25c937d2c90ef30c2b7f96a47ee8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


client/module_core/lib/src/repositories/notification_repository.dart
```diff
@@ -2,21 +2,42 @@ import "package:injectable/injectable.dart";
...
+  }) : _api = api,
+       _deviceIdStorage = deviceIdStorage;
 
-  Future<void> registerToken({required String token, required DevicePlatform platform}) {
+  /// Sends the same device ID the notification settings are stored under, so the
+  /// server can match this push token to the preferences for this device.
+  Future<void> registerToken({required String token, required DevicePlatform platform}) async {
     return _api.registerToken(
-      request: RegisterTokenRequest(token: token, platform: platform),
+      request: RegisterTokenRequest(token: token, platform: platform, deviceId: await _readDeviceId()),
```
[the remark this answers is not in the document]
stanvladut: [Sesori reply] Addressed (in commit 26f6977)

Valid, and citing a rule I should have followed without prompting — `AGENTS.md` requires the relevant `docs/regression/` document to move with a material behavior change, and this is one: a disabled category is now suppressed before delivery rather than only at foreground rendering.

`docs/regression/notifications.md` now records:

- **Required Behavior** — registration sends the same device ID the preferences are stored under, so the server can associate this token with this device; a device ID that cannot be read registers the token without one, leaving that install unfiltered rather than losing push, and recovers on the next registration.
- **L3 Release** — registration coverage now explicitly includes the device ID.
- **L4 Extended** — disabling a category on one device suppresses its remote delivery there while another device still receives it.

I put the cross-device scenario at L4 deliberately: it needs real delivery to two devices, which is exactly what the existing preference-gating coverage never exercised, since foreground rendering is single-device by construction.

Separately, your review also made me notice the PR title was missing the required complexity emoji. Fixed to `🌿`.

client/module_core/lib/src/capabilities/notifications/register_token_request.g.dart
```diff
@@ -10,13 +10,15 @@ _RegisterTokenRequest _$RegisterTokenRequestFromJson(Map json) =>
...
       platform: $enumDecode(_$DevicePlatformEnumMap, json['platform']),
+      deviceId: json['deviceId'] as String?,
     );
 
 Map<String, dynamic> _$RegisterTokenRequestToJson(
   _RegisterTokenRequest instance,
 ) => <String, dynamic>{
   'token': instance.token,
   'platform': _$DevicePlatformEnumMap[instance.platform]!,
+  'deviceId': ?instance.deviceId,
```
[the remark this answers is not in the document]
stanvladut: [Sesori reply] Not addressed

`'deviceId': ?instance.deviceId` is valid Dart. It is the null-aware element syntax introduced in Dart 3.12, which omits the entry entirely when the value is null — exactly what `@JsonKey(includeIfNull: false)` is supposed to produce. This repo pins Flutter 3.44.8-stable / Dart 3.12.2, so it compiles.

Evidence rather than assertion — both of these were run before pushing:

- `dart analyze --fatal-infos` on `client/app`, `module_core`, `module_auth`, `module_prego` — "No issues found!" in all four. Invalid Dart in a `part` file would fail here first.
- `dart test` in `module_core` — 1068 passed, including a new case asserting `request.toJson()` does **not** contain a `deviceId` key when it is null. That test exercises this exact generated line.

CI agrees: `analyze-test` and `test` are both green on this PR.

On the generated-file contract: the file was not hand-edited. I ran `dart run build_runner build --delete-conflicting-outputs` under the pinned toolchain, which is why `injection.config.dart` regenerated alongside it — the repository constructor gained a dependency.

client/module_core/lib/src/repositories/notification_repository.dart
```diff
@@ -2,21 +2,42 @@ import "package:injectable/injectable.dart";
...
   }
+
+  // A device ID that cannot be read must not cost the user push entirely. The
+  // server keeps delivering to a token registered without one, so registering
+  // unfiltered is strictly better than failing the registration.
+  Future<String?> _readDeviceId() async {
+    try {
+      return await _deviceIdStorage.getOrCreate();
+    } catch (error, stackTrace) {
+      logw("Failed to read notification device ID; registering token without it", error, stackTrace);
```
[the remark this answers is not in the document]
stanvladut: [Sesori reply] Not addressed

Deliberate, and narrowing it would invert the tradeoff this code exists to make.

The catch is broad because the consequence of *not* catching is worse than the failure it is hiding. `registerToken` is the only path that puts a push token on the server. If `getOrCreate()` throws anything — `Exception` or `Error` — and that propagates, the token is never registered and the user receives **no notifications at all** until the next registration attempt. Catching only `Exception` would let a `TypeError` in device-ID code silently cost a user their entire push channel, which is a much larger blast radius than losing per-device filtering for one install.

Debuggability is preserved rather than sacrificed: `logw("Failed to read notification device ID; registering token without it", error, stackTrace)` records the error *and* the stack trace, so a programmer error still surfaces with everything needed to diagnose it. It is swallowed for control flow, not for observability.

Covered by a test that makes the storage throw and asserts the token is still registered with a null device ID.

Happy to revisit if you would rather a programmer error hard-fail registration — but that is a product decision about what a bug in this path should cost the user, not a cleanup.

client/module_core/lib/src/repositories/notification_repository.dart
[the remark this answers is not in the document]
stanvladut: [Sesori reply] Not addressed

Real failure mode, but "permanently" overstates it — the recovery path already exists.

`NotificationRegistrationService.start()` is called during app startup (`client/app/lib/main.dart:209`), and syncs the *initial* auth state rather than only reacting to changes. For an authenticated user that reaches `_registerCurrentToken()`, and the early-return guard there is `_currentRegisteredToken`, a plain instance field that is empty on a fresh launch. So the registration call goes out again on **every app open**, re-reading the device ID.

A transient secure-storage failure therefore leaves the token unfiltered until the next app launch, not indefinitely. Given the trigger is secure storage being unavailable — rare, and typically resolved by the time the process restarts — a session-scoped retry would be machinery for a window that already closes on its own.

The one case it does not cover is a session that stays alive for a very long time after a failed read, since `_currentRegisteredToken` suppresses re-registration within the session. Making the repository report "registered without a device ID" so the service could retry mid-session is doable, but it means threading that signal through the repository return type and holding pending state in the service — a meaningful amount of coupling for a rare, self-healing case.

I have documented the behavior instead: `docs/regression/notifications.md` now states the install stays unfiltered and recovers on the next registration, so it is a known bound rather than a silent one. Happy to add the mid-session retry if you would rather not depend on the restart.

client/module_core/lib/src/capabilities/notifications/register_token_request.dart
```diff
@@ -9,6 +9,9 @@ sealed class RegisterTokenRequest with _$RegisterTokenRequest {
   const factory RegisterTokenRequest({
     required String token,
     required DevicePlatform platform,
+    // Omitted rather than sent as null: the server treats deviceId as optional,
+    // which accepts an absent key but rejects an explicit null with a 400.
+    @JsonKey(includeIfNull: false) required String? deviceId,
```
AlexDochioiu: Get rid of `@JsonKey(includeIfNull: false)`. It's already default behaviour

client/module_core/lib/src/repositories/notification_repository.dart
```diff
@@ -2,21 +2,42 @@ import "package:injectable/injectable.dart";
...
+  Future<void> registerToken({required String token, required DevicePlatform platform}) async {
     return _api.registerToken(
-      request: RegisterTokenRequest(token: token, platform: platform),
+      request: RegisterTokenRequest(token: token, platform: platform, deviceId: await _readDeviceId()),
     );
   }
 
   Future<void> unregisterToken({required String token}) {
     return _api.unregisterToken(token: token);
   }
```
AlexDochioiu, line 30: I think token + device id get wiped on logout? and then new token and device id gets generated on re-login?

If this is corect, please check at server level that we're purging from server's DB both this token and device id (and any device settings) we stored.

client/module_core/lib/src/capabilities/notifications/register_token_request.dart
stanvladut: [Sesori reply] Not addressed

I checked this by running the generator both ways rather than trusting either of our assumptions, and the annotation turns out to be load-bearing here.

**With** `@JsonKey(includeIfNull: false)`:
```dart
'deviceId': ?instance.deviceId,   // null-aware element: key omitted when null
```

**Without** it:
```dart
'deviceId': instance.deviceId,    // serializes "deviceId": null
```

`json_serializable` defaults `includeIfNull` to **true**, and our `build.yaml` only sets `any_map` and `explicit_to_json`, so nothing overrides it package-wide. No existing generated file in `module_core` uses the null-aware form, which matches.

It matters because of the server contract: `deviceId` is validated with an optional schema, which accepts an **absent** key but rejects an explicit `null` with a 400. `deviceId` is only null when secure storage failed, and that is precisely the case we degrade gracefully — the token should still register, unfiltered. Dropping the annotation would turn "registered but unfiltered" into "registration 400s, no push at all", which is the opposite of the intent.

Covered by a test asserting `toJson()` does not contain a `deviceId` key when it is null, so a future removal fails rather than shipping quietly.

Happy to drop it if you would rather the server accept an explicit null instead — that is a one-line schema change on the auth side, and then the annotation genuinely is redundant.

client/module_core/lib/src/repositories/notification_repository.dart
stanvladut: [Sesori reply] Question / no change

Traced both sides. The premise is half right, and I think the current behaviour is what we want — but flagging one real gap at the end.

**Token: purged on both sides already.** The client calls `unregisterCurrentDevice()` before local auth is cleared (`DELETE /notifications/tokens/{token}`), and server-side `logoutUser()` also runs `deviceTokenRepo.deleteAllForUser(userId)`. `revoke()` goes through the same path. So no token row survives a logout.

**Device ID: not wiped, and not regenerated on re-login.** `AuthManager.logoutCurrentDevice()` clears tokens, the PKCE verifier, the auth provider and the OAuth session — it does not touch `notification_preferences_device_id_v1`, and nothing else deletes that key. So the same device ID comes back on re-login.

**Device settings: not purged server-side**, and I think deliberately so. Because the device ID survives, logging out and back in on the same phone restores that device's notification preferences. Purging them would silently reset a user's toggles on every logout, which seems worse than keeping them. There is also no cross-account leak: `settingsConfiguration` is keyed by `(userId, deviceId)`, so a different account signing in on the same hardware gets its own row and cannot read the previous account's.

**The gap worth naming:** this is right for *logout*, but it means a `settingsConfiguration` row outlives an account that is deleted rather than logged out. If there is an account-deletion or privacy-erasure path, that collection should be on its purge list — it is not today. That is auth-server work rather than something this PR can fix, so tell me if you want it and I will raise it separately.

No change here unless you would prefer preferences to reset on logout, which is a product call rather than a correctness one.

client/module_core/lib/src/capabilities/notifications/register_token_request.dart
AlexDochioiu: Last PR merged just made this configuration global. No longer needed here.

stanvladut: [Sesori reply] Addressed (in commit 18f0ad6)

You're right, and my earlier reasoning was out of date rather than wrong at the time. `client/module_core/build.yaml` didn't exist at this branch's merge base; the merged PR added it across every package with `include_if_null: false`, which is exactly the behaviour the per-field annotation was there to force. Removed it.

Verified rather than assumed, since the annotation was load-bearing before: regenerating without it produces a **byte-identical** `register_token_request.g.dart`, still emitting the null-aware `'deviceId': ?instance.deviceId`. The key is still omitted rather than serialized as null, so the server contract is untouched. The api test asserting the body is exactly `{token, platform}` when `deviceId` is null still passes.

That also settles qodo's separate "invalid Dart" flag on the same line: the toolchain now generates that form repo-wide.

Picked up in the same push: this branch had gone conflicting against main. The only conflict was `register_token_request.freezed.dart`, resolved by regenerating under the new pinned toolchain (`.tool-versions` moved to `3.47.0-0.4.pre-beta`) rather than by hand. Merge and annotation removal are separate commits. Analyze is clean on app, module_core, module_auth and module_prego, with 1068 module_core and 945 client/app tests passing.

PR #985, opened 2026-08-19
⚙️ feat(hermes): surface Hermes model in the session picker

eexxio:
## Complexity

⚙️ Moderate — touches the shared ACP protocol layer (additive: parse `models`, new method constant, repo writer) plus the Hermes plugin, with unit tests. No wire-contract breakage; existing behaviour is preserved when a backend doesn't advertise a model catalog.

## What

Makes the Hermes Agent harness surface its model in the Sesori new-session picker instead of rendering an empty list.

- `HermesPlugin.captureSessionConfig` seeds the ACP configuration tracker (process default + per-session override) from Hermes's ACP `SessionModelState` — `availableModels[]` + `currentModelId` — so `AcpSessionOptionsService` synthesises a populated provider.
- `HermesPlugin.applyTurnSelection` issues the standard ACP `session/set_model` to switch a session to a user-selected model; when no model is requested it preserves the Hermes-configured default (no-op).
- ACP layer (additive): parse the `models` field on `AcpNewSessionResult` into `AcpModelInfo`/`AcpSessionModelState`; carry the provider catalog on `AcpSessionConfigurationTracker`; derive picker providers from the catalog (falls back to the single-model shape when only `modelId` is set); add `AcpMethods.sessionSetModel` + `AcpSessionConfigRepository.setModel`.

## Why

Hermes is policy-only over ACP and never populated the configuration tracker, so the new-session model picker was empty (confirmed in the 2026-08-18 iOS E2E). This restores visibility of the active model and enables switching, matching the other menu-bearing harnesses (Cursor, OMP).

## Risk and test focus

- If Hermes returns no `models` field (`models == null` or empty catalog), capture is a no-op and the picker stays empty exactly as before — no regression for non-advertising backends.
- Model id splitting assumes Hermes's encoded `provider:model`; ids without a separator fall back to the plugin id.
- Test focus: capture surfaces the active model, empty-catalog leaves the picker empty, and a selected model writes `session/set_model`.

## Expected result

- `dart analyze --fatal-infos` clean in `sesori_plugin_hermes` and `sesori_plugin_acp`.
- 28/28 unit tests green (3 new: capture surfaces model, no-models leaves picker empty, switch writes `session/set_model`).
- Live against the real `hermes` binary, `getProviders()` returns the full populated catalog with `isAvailable=true` (VERIFY: PASS).

## Verification

- ✅ `dart analyze --fatal-infos` — clean (hermes + acp)
- ✅ `dart test` — 28/28 pass
- ✅ Bridge rebuilt (`dart build cli`) and runs connected to `wss://relay.sesori.com`
- ✅ Live probe: real `hermes` binary → `getProviders()` returns the populated catalog (e.g. `opencode-go:deepseek-v4-flash`, `grok-4.5`, `gemini-3.5-flash-lite`, …)

No user-facing or database schema impact beyond the now-populated model picker.

Closes #1 (owner issue: "Hermes harness: cannot see/select the models provided by the Hermes integration")

---
## Summary by cubic
Makes the Hermes model catalog visible in the new-session picker and enables model switching. Previously the picker was empty and sessions could not switch; now it is populated from ACP SessionModelState and selections are applied via session/set_model. Backends that do not advertise models behave unchanged.

- ACP: parse the models field in AcpNewSessionResult into AcpModelInfo/AcpSessionModelState; add AcpMethods.sessionSetModel; add AcpSessionConfigRepository.setModel.
- Tracker and options: store availableModels on AcpSessionConfigurationTracker; AcpSessionOptionsService builds providers from the catalog and falls back to a single-model shape when only modelId exists.
- Hermes plugin: captureSessionConfig seeds defaults and the catalog; applyTurnSelection issues session/set_model and preserves the default when no model is selected; provider parsed from provider:model (ids without a colon default to "hermes").
- Safety: if models is null or empty, the picker stays empty; no wire-contract breakage; non-advertising backends unchanged.
- Tests and tooling: unit tests cover capture, empty catalog, and switching; live probes added in `sesori_plugin_hermes` tools. No migration required.

<sup>Written for commit 7f9d18ef045932c5bb40a5d6aa481ad5dc7e5321. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.plan/active/hermes-model-visibility/PLAN.md
+++ b/.plan/active/hermes-model-visibility/PLAN.md
@@ -1,0 +1,192 @@
+# Surface the Hermes Configured Model in Sesori (Model Visibility)
+
+## Status
+
+- **Plan slug:** `hermes-model-visibility`
+- **Status:** active (plan authored; awaiting owner approval before implementation)
+- **Plan date:** 2026-08-19
+- **Owner issue:** `eexxio/sesori_apps_monorepo#1` — "Hermes harness: cannot see/select the models provided by the Hermes integration"
+- **Implementation base:** current `origin/main` (`983d5b44`)
+- **Delivery:** small PR series (plan PR first, then implementation PRs)
+
+## Goal
+
+When a user creates a new session for the **Hermes Agent** harness in Sesori, the
+new-session screen's model picker is currently **empty** — the user cannot see
+which model the session will run, or choose one. Make Hermes surface its models
+so the picker shows the **active model** (visibility) **and lets the user pick a
+different model** (switching), per owner decision 2026-08-19 ("both").
+
+Non-goal: as owner approved, switching is in scope, but only within what Hermes's
+ACP surface supports (verify; see Scope below). Hermes remains authoritative over
+its model config — Sesori reads the available models and issues the switch, it
+does not claim to own/install models.
+
+## Current Context (verified against `origin/main`, 2026-08-19)
+
+- `bridge/sesori_plugin_hermes/lib/src/hermes_plugin_impl.dart` is a policy-only
+  `AcpPlugin` subclass. Its doc comment states *"Version 1 uses Hermes's
+  configured model/mode rather than exposing a picker."* It does **not** override
+  `captureSessionConfig` (base `AcpPlugin.captureSessionConfig` is a no-op).
+- `bridge/sesori_plugin_acp/lib/src/acp_session_options_service.dart`
+  `getSessionOptions()` synthesizes a single `PluginProvider` (with one
+  `PluginModel`) **only when** `_configurationTracker.processDefaults.modelId`
+  is non-null. When the tracker has no model, `providers:` is `[]`.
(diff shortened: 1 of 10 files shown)


AlexDochioiu:
Thanks @eexxio for investigating this and putting together the initial implementation. The feature has now landed through #995 and #999, which keep the Hermes-specific model discovery and switching behavior within the Hermes plugin and cover the required lifecycle and cleanup behavior. Since those PRs supersede this implementation, I'm closing this PR. We really appreciate the contribution and the work that helped validate the feature.

PR #991, opened 2026-08-20
🌿 fix(hermes): stop streamed response text disappearing after reconnect

eexxio:
## Complexity

🌿 straightforward — one focused change in a single client method plus a regression test and a regression-doc note.

## What

Fixes streamed Hermes response text disappearing from the committed transcript when the end-of-turn finalize snapshot is lost.

- `session_detail_cubit.dart` `_onPartDelta`: the delta is now also appended into the committed text part (alongside the transient streaming buffer), instead of only feeding the buffer.
- Adds `session_detail_streaming_text_regression_test.dart` covering both the in-order happy path and the lost-finalize path.
- Notes the fix in `docs/regression/session-turns.md`.

## Why

Hermes response text streams to the client as part deltas, but the cubit only wrote those deltas into the transient `StreamingTextBuffer` and relied on the bridge's end-of-turn finalize snapshot to commit the text into the message part. When that snapshot is lost (disconnect/reconnect mid-turn, event loss), the committed part stayed empty — the streamed text vanished once the streaming overlay cleared. With multiple tool calls the turn is long and splits the response into several separate text parts (one per text segment between tool groups), each depending on its own finalize snapshot, so a missed snapshot empties that whole segment: "the text in the Hermes agent's response disappears when there are multiple tool calls."

## Risk and test focus

- Low risk: the fix only makes `_onPartDelta` also commit into the part; the existing finalize-snapshot path still overwrites with identical text, so no doubling on the happy path (covered by the "exactly once" test).
- Client session-detail streaming + reconnect behavior: run `dart test test/cubits/session_detail/` in `client/module_core`.
- Analyzer: `dart analyze` clean on the changed cubit and test.

## Expected result

Response text for a multi-tool Hermes turn is committed into the message parts as it streams, so it survives a mid-turn disconnect/reconnect and no longer disappears from the rendered transcript. The in-order happy path and behavior around existing suites are unchanged.

## Verification (additional)

- `session_detail_streaming_text_regression_test.dart` passes with the fix and fails without it (verified).
- Full `test/cubits/session_detail/` suite (115 tests) passes.
- Reproduction was driven against a real captured Hermes multi-tool turn over ACP.

---
## Summary by cubic
Prevents streamed Hermes assistant text from disappearing after reconnects by committing accumulated text deltas into the committed message part on the render throttle. Previously only the transient buffer updated; if the end‑of‑turn finalize snapshot was lost the committed part stayed empty.

- Review focus: `_emitStreamingSnapshot` now appends buffered text into committed parts (one coalesced emit per flush) using a per‑part `_streamingTextCommitWatermark` to append only the uncommitted tail; `_doSilentRefresh` re-arms the buffer and seeds the watermark from the refreshed snapshot to avoid doubling; `_onPartUpdated` (finalize) clears the watermark; `_onPartDelta` continues to buffer deltas.
- Behavior and risk: coalesced updates improve render performance; finalize snapshots still overwrite with identical text (no duplication on the happy path); protects against replayed prefixes after reconnect; no API/schema changes.
- Verification: `session_detail_streaming_text_regression_test.dart` covers lost-finalize and happy paths; run `dart test test/cubits/session_detail/` in `client/module_core`.

<sup>Written for commit 226c3690ca21e0fd5ad4973766aefcf03f60ba09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
+++ b/client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
@@ -1280,6 +1280,37 @@ class SessionDetailCubit(
 
   void _onPartDelta({required String partId, required String delta}) {
     _streamingBuffer.appendDelta(partId: partId, delta: delta);
+
+    // Keep the committed part's text in sync with the streamed delta. Streamed
+    // deltas are the live source of the response text; the bridge's end-of-turn
+    // snapshot (_onPartUpdated) is a duplicate commit, not the only one. Without
+    // this, text for parts written before their finalize snapshot is lost if that
+    // snapshot never arrives (disconnect/reconnect mid-turn, event loss), leaving
+    // an empty bubble — the "response text disappears" symptom. The finalize
+    // snapshot later overwrites the part with identical complete text, so this is
+    // idempotent on the happy path.
+    final current = state;
+    if (current is! SessionDetailLoaded) return;
+    final messageIndex = current.messages.indexWhere((item) =>
+        item.parts.any((part) => part.id == partId));
+    if (messageIndex < 0) {
+      // The message/part envelope hasn't arrived yet — keep the text only in the
+      // streaming buffer; the envelope + snapshot will commit it when they land.
+      return;
+    }
+    final message = current.messages[messageIndex];
+    final partIndex = message.parts.indexWhere((item) => item.id == partId);
+    if (partIndex < 0) return;
+    final part = message.parts[partIndex];
+    if (part.type != MessagePartType.text) return;
+    final existing = part.text;
+    if (existing == null) return;
+    final messages = List<MessageWithParts>.from(current.messages);
+    final parts = List<MessagePart>.from(message.parts);
+    parts[partIndex] = part.copyWith(text: existing + delta);
+    messages[messageIndex] = message.copyWith(parts: parts);
+    if (isClosed) return;
+    emit(current.copyWith(messages: messages, streamingText: _streamingBuffer.snapshot()));
   }
 
   void _onPartUpdated(MessagePart part) {
--- a/client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
+++ b/client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
@@ -85,6 +85,13 @@ class SessionDetailCubit(
   late final StreamSubscription<void> _staleSubscription;
   late final StreamSubscription<LifecycleState> _lifecycleSubscription;
   late final StreamingTextBuffer _streamingBuffer;
+  /// Tracks, per streaming text part id, how many characters of the buffered
+  /// deltas have already been committed into the message part. Used for two
+  /// things: (1) committing only the *new* tail of the buffer on each throttled
+  /// flush, and (2) not re-appending a replayed prefix after a reconnect — a
+  /// fresh snapshot may already carry some of the streamed text, so appending
+  /// the full buffered value would double it.
+  final Map<String, int> _streamingTextCommitWatermark = {};
   Future<void>? _activeRefresh;
   int _commandCatalogGeneration = 0;
   Timer? _eventRefreshCooldown;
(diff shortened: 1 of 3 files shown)


client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -511,6 +518,22 @@ class SessionDetailCubit(
...
+          // snapshot (already holding some of the streamed prefix) is not
+          // doubled when the remaining deltas are appended.
+          for (final entry in streamingText.entries) {
+            _streamingBuffer.appendDelta(partId: entry.key, delta: entry.value);
+            final existingLength = _partTextLength(
+              messages: snapshot.messages,
+              partId: entry.key,
+            );
+            _streamingTextCommitWatermark[entry.key] =
+                existingLength > entry.value.length ? entry.value.length : existingLength;
```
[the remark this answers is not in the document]
AlexDochioiu: Confirmed on the current head with a focused `SessionDetailCubit` refresh test. The local buffer contained `I’ll inspect the repo. ` and the reload snapshot contained `I’ll inspect the repository completely. `. After `_doSilentRefresh`, `messages` held the complete snapshot but `streamingText` still held the shorter local value; `AssistantMessageCard` therefore rendered the truncated text. With the finalize event already lost, that stale overlay can persist indefinitely. Treating length as prefix identity is not sufficient, so this is blocking.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -1369,11 +1402,68 @@ class SessionDetailCubit(
     }
   }
 
+  /// Commits the buffer's accumulated streamed text into the committed message
+  /// parts and emits once, coalesced to the buffer's render throttle. Only the
+  /// not-yet-committed tail of each part is appended (tracked by
+  /// [_streamingTextCommitWatermark]), so a replayed prefix after a reconnect
+  /// cannot double the text. The overflow parts that are still buffered (no
+  /// message envelope yet) are surfaced as [streamingText] as before.
   void _emitStreamingSnapshot() {
```
AlexDochioiu, line 1411: This turns a presentation buffer into a second transcript authority. The Cubit now writes transient SSE deltas into `messages` while refresh snapshots also replace `messages`, so it owns prefix reconciliation, replay deduplication, and commit watermarks for every backend. That is lifecycle/business state rather than Cubit orchestration, and it survives only this Cubit instance. Since the reported behavior is Hermes/ACP and `AcpEventMapper` already accumulates complete text for final snapshots, please first repair the authoritative ACP/history snapshot boundary. If generic client reconciliation is genuinely required, put that invariant in one explicit tracker/service behind the Cubit and test snapshot/delta ordering there rather than keeping this split source of truth in `SessionDetailCubit`.

client/module_core/test/cubits/session_detail/session_detail_streaming_text_regression_test.dart
```diff
@@ -0,0 +1,279 @@
...
+    String committedText(SessionDetailCubit cubit) {
+      final state = cubit.state as SessionDetailLoaded;
+      return state.messages
+          .expand((m) => m.parts)
+          .where((p) => p.type == MessagePartType.text && p.text != null)
+          .map((p) => p.text!)
+          .join("|");
+    }
+
+    test("streamed text is committed even when end-of-turn finalize snapshots are lost", () async {
```
AlexDochioiu, line 243: These tests never drive the behavior described in the PR: connection status never changes and `reload` is never called. Omitting a finalize event from one uninterrupted event stream exercises only the new commit path, not `_doSilentRefresh`, watermark reseeding, missed deltas, or snapshot reconciliation. The current truncation bug at line 535 passes both tests. Please add a reconnect/refresh case where the authoritative snapshot is ahead of the local buffer and assert the resolved/rendered text after refresh.

AlexDochioiu (changes requested):
The current implementation is not safe to merge. A focused Cubit refresh reproduction confirms that the local streaming overlay can override a newer authoritative snapshot with truncated text, while the added tests never execute reconnect or refresh.

The broader reconciliation state also belongs at one authoritative producer/reconciler boundary rather than split between transient `streamingText` and canonical `messages` inside the shared Cubit. Please address the inline findings and rebase onto current `main`, which now conflicts with this branch.

AlexDochioiu:
Thanks @eexxio for investigating this and sharing the captured Hermes behavior. We reproduced a correctness problem in this implementation's reconnect reconciliation: the transient client buffer can override a newer authoritative snapshot, and the added tests do not exercise the reconnect path. The shared Cubit also becomes a second transcript authority for every backend. We are going to re-implement the fix at the authoritative Hermes/ACP boundary with focused reconnect coverage, so I am closing this PR in favor of the maintainer implementation. We appreciate the report and the work that narrowed down the symptom.

AlexDochioiu:
The maintainer replacement is now open as #1006. It fixes the authoritative ACP event boundary by finalizing completed assistant text before each tool call, with focused multi-tool regression coverage and no client-side reconciliation state.

PR #1069, opened 2026-08-24
🌿 Fix OpenCode prompt reservation failures

daniil-shumko:
## Complexity

🌿 Straightforward: narrow OpenCode plugin request-shaping fix.

## What

- Reserve normal OpenCode message IDs without agent, model, or variant fields.
- Keep those selections on the real prompt or slash-command dispatch.
- Add regression coverage and document the reservation boundary.

## Why

The synchronous reservation request unnecessarily resolves and applies supplied session options before the real dispatch. OpenCode can report failures from that preparatory request as a generic 500, preventing the actual asynchronous prompt request from owning option selection and reporting its outcome.

## Risk and test focus

Low risk. The reservation remains an empty, no-reply message and still obtains OpenCode's ordered server ID. Test focus is request serialization and prompt/command correlation. There is no database impact.

## Expected result

OpenCode prompt and slash-command reservations no longer fail while applying options intended for the real dispatch. User-visible behavior is otherwise unchanged.

## Verification

- `dart pub get` from `bridge/`
- `dart test test/opencode_repository_test.dart test/opencode_service_test.dart test/opencode_plugin_impl_test.dart`
- `dart analyze --fatal-infos` from `bridge/sesori_plugin_opencode/`

### DIFF
--- a/bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
+++ b/bridge/sesori_plugin_opencode/lib/src/opencode_plugin_impl.dart
@@ -427,9 +427,6 @@ class OpenCodePlugin._({
     final messageId = await _call(
       () => _service.reserveMessage(
         sessionId: sessionId,
-        agent: agent,
-        variant: variant,
-        model: model,
       ),
     );
     _promptMessages.record(messageId: messageId, promptId: promptId);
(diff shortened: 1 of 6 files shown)


AlexDochioiu:
Closing this in favor of #1073.

The affected user's OpenCode 1.18.20 log mapped the bundled stack location to `SessionPrompt.createUserMessage`'s **Agent not found** branch. Omitting options from the reservation would let that preparatory request use the default agent, but the unchanged missing agent would still be sent to `prompt_async`. Because that endpoint returns 204 before background processing succeeds, the failure would bypass Sesori's existing typed stale-options refresh/retry flow.

#1073 instead keeps the requested options on the synchronous reservation. After OpenCode's ambiguous 500, it checks the live project catalog and throws `PluginStaleOptionsException` only when the requested agent, model, or variant is actually absent or unavailable; unrelated 500s remain unchanged.

PR #1081, opened 2026-08-24
🌱 Sharpen the compatibility marker rules and drop the agent archive

AlexDochioiu:
## Complexity

Trivial. Two instruction bullets, one file deletion, and two stale references.

## What

**Compatibility markers.** Two rules the existing bullet left unstated:

- `vX.Y.Z` is always the **current** product version, read from the code at the
  time of writing, never a guess at the release the change will ship in. The
  rule names where to read it — `bridge/app/pubspec.yaml` or
  `client/app/pubspec.yaml`, both `1.8.1` — and rules out a library package's
  own version, such as `shared/sesori_shared` at `0.3.0`, even when the marker
  sits in that package. The cleanup is stated as the condition that retires it
  ("once every supported bridge sends it") rather than a future version.
- The marker attaches to the smallest thing it explains — the field, parameter,
  enum value, or branch — inside the enclosing declaration, not above the
  enclosing function, factory, constructor, class, or its doc comment. The
  enclosing declaration gets it only when that whole declaration is what
  compatibility retains.

**Archive.** Deletes `AGENTS.bak.md` (693 lines), the pointer paragraph in
`AGENTS.md` that introduced it, and the stale entry in the codebase-cleanup plan
that still listed it as deliberately retained.

## Why

Both marker rules come from getting them wrong in #1074, in the same comment.
The version was invented as `v1.9.0` — a release that does not exist — when the
current version is `1.8.1`; a marker dated to a guessed future release cannot be
matched against anything. Naming the source explicitly matters because the
marker in question lives in `shared/sesori_shared`, whose own package version is
`0.3.0`; reading the nearest `pubspec.yaml` would have been the obvious move and
would also have been wrong. And the marker was placed above the enclosing
`const factory` and its doc comment, which reads as though the whole completion
variant exists for compatibility rather than the single nullable field that
does. That is misleading exactly where the marker exists to be unambiguous, and
it obscures the cleanup boundary: a reader retiring the baseline needs to know
whether to delete a field or a declaration.

The placement convention was already being followed — see
`plugin_management.dart:68`, where the marker sits directly above
`authenticationState` inside the factory — it just was not written down.

The archive deletion is an owner decision. The codebase-cleanup plan had listed
`AGENTS.bak.md` among things it deliberately kept; that entry now records that
the owner had it removed, since Git history retains the content anyway and
`AGENTS.md` had to carry a standing "do not follow this file" warning to keep it
from being read as guidance.

## Risk and test focus

None. Instruction and documentation only; no code, no behaviour, no consumer.

Existing markers already comply with the placement rule, so nothing needs
migrating. A repository-wide search confirms no reference to `AGENTS.bak.md`
survives except the cleanup plan's note that it was deleted.

## Expected result

No user-visible change. No database change. No code change.

Future compatibility markers carry the current version and attach to the
smallest thing they explain. `AGENTS.md` no longer opens by warning about a file
that does not exist.

## Verification

- `git diff --check`
- documentation only in the diff: 2 insertions, 697 deletions across three files
- `grep -rn "AGENTS.bak"` returns only the cleanup plan's deletion note
- `docs/parallel-plugins/PLAN.md`, the remaining declared archive, is untouched
- no Dart or Flutter suites run: instruction-only change


AGENTS.md
```diff
@@ -70,6 +67,21 @@ eagerly "just in case."
...
+  Do not use a library package's own version, such as
+  `shared/sesori_shared/pubspec.yaml`, even when the marker sits in that
+  package. State the cleanup as the condition that retires the marker, such as
+  "once every supported bridge sends it", rather than naming a future version.
+- Attach that marker to the smallest thing it explains. A field, parameter,
+  enum value, or branch that exists for compatibility carries the marker
+  directly above itself, inside the enclosing declaration, not above the
+  enclosing function, factory, constructor, class, or its doc comment. Mark the
+  enclosing declaration only when that whole declaration is what compatibility
+  retains.
```
AlexDochioiu: This is too long. Compress this in a shorter and clear text explaining those

AlexDochioiu: [Sesori reply] Addressed (in commit c28629df8)

Agreed — 14 lines for two rules in a file whose bullets run 2-5. Down to 7.

The version rule now folds into the existing bullet, directly after the format line it describes, rather than standing alone:

```
  `// COMPATIBILITY YYYY-MM-DD (vX.Y.Z): ...`
  `vX.Y.Z` is the current product version from `bridge/app/pubspec.yaml` (or
  `client/app`) — never a library package's own version, never an unreleased
  one. Word the cleanup as its retiring condition, not a future version.
- Put the marker directly above the field, parameter, or branch it explains,
  inside the enclosing declaration, not above the enclosing function, factory,
  class, or doc comment. Mark the enclosing declaration only when it is itself
  what compatibility retains.
```

Kept all four facts and dropped the prose around them: current product version, where to read it, not a library version, not an unreleased one, cleanup by condition, and placement on the smallest thing with its one exception. Cut the worked example of `shared/sesori_shared` at `0.3.0` — "never a library package's own version" carries it without the illustration.

AlexDochioiu: [Sesori reply] Follow-up (in commit 6961640f5)

Dropped "never an unreleased one" from the quoted text above, per your note — during development the pubspec version normally *is* the unreleased one, so that clause contradicted the instruction to read the version from the code. The rule now reads:

```
  `vX.Y.Z` is the current product version from `bridge/app/pubspec.yaml` (or
  `client/app`), never a library package's own version. Word the cleanup as its
  retiring condition, not a future version.
```

"Not a future version" survives on the cleanup clause, which is where it actually mattered — the original mistake was naming `v1.9.0` as the retirement target, not the marker date.

PR #1120, opened 2026-08-25
⚙️ Fix Claude queued image prompts

AlexDochioiu:
## Complexity
⚙️ Moderate — localized Claude stream correlation plus queued-message presentation.

## What
- Correlate Claude image echoes even when the CLI omits `isReplay` or decorates image metadata.
- Stamp the resulting user message with the queued prompt ID so it replaces the queued row.
- Show locally staged image thumbnails in queued bubbles and a bounded image count for bridge-owned queued prompts.

## Why
Claude Code can surface an image user echo without the usual replay marker. The bridge then rendered the image as an uncorrelated message while leaving its queued copy visible.

## Risk and test focus
User-visible session transcript and queue handoff behavior only. No database impact. The image fallback remains content-exact and applies to unmarked image echoes only; unmarked text remains uncorrelated.

## Expected result
An image prompt appears once as its sent message, its queued bubble clears as soon as the echo is rendered, and queued image state visibly indicates its attachment.

## Verification
- `cd bridge && dart test sesori_plugin_claude/test/claude_session_process_repository_test.dart sesori_plugin_claude/test/claude_plugin_impl_test.dart`
- `cd bridge && dart analyze sesori_plugin_claude`
- `cd client && flutter test app/test/features/session_detail/widgets/session_detail_body_test.dart app/test/features/session_detail/widgets/session_detail_message_list_test.dart`
- `cd client && dart analyze app`

### DIFF
--- a/bridge/sesori_plugin_claude/lib/src/repositories/claude_session_process_repository.dart
+++ b/bridge/sesori_plugin_claude/lib/src/repositories/claude_session_process_repository.dart
@@ -1,5 +1,6 @@
 import "dart:async";
 import "dart:collection";
+import "dart:convert";
 
 import "package:sesori_plugin_interface/sesori_plugin_interface.dart";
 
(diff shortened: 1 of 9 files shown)


AlexDochioiu (commented):
No blocking findings.

I reviewed the Claude echo-correlation state flow, semantic image-source matching, queue consumption ordering, and the local-versus-bridge-owned attachment presentation. The focused bridge/client tests and analyzers pass, and the PR checks are green.

PR #1147, opened 2026-08-27
🚧 [default-new-session-prefill] prefill new sessions from plugin defaults [step 3/3]

AlexDochioiu:
## Complexity

🚧 Complex — this completes a backend-neutral flow across the shared wire contract, bridge creation/options services, client repository/service resolution, and New Session UI behavior.

## What

- Persist the complete requested agent/model/effort selection per plugin after a session is created successfully and durably.
- Add nullable `lastUsedPromptDefaults` to the shared session-options response.
- Overlay remembered defaults after loading the option catalog, keeping preference state separate from the catalog cache.
- Map the preference through the client repository and use it to prefill New Session controls.
- Preserve deliberate in-memory choices by dimension and validate all remembered values against the current agent/model/variant catalogs.
- Add focused wire, bridge service, client service/repository, and widget coverage and complete the regression documentation.

## Why

Developers commonly create consecutive sessions with the same harness choices. Remembering the last successful selection removes repeated setup while keeping defaults isolated per plugin and avoiding stale or removed catalog values.

## Risk and test focus

- **Risk:** moderate, centered on cross-version transport behavior and selection fallback precedence.
- **User-visible impact:** the next New Session screen can preselect the last successfully used agent, model, and effort for the selected plugin.
- **Database impact:** no new schema in this step; it begins populating the v14 plugin-default table introduced in step 2 after successful creations. The preference intentionally survives retiring an individual session.
- **Wire compatibility:** older bridges omit the nullable field and newer clients fall back normally; older clients ignore the added response field.
- **Failure isolation:** preference read/write failures are logged but do not hide usable option catalogs or turn an already-created session into a retryable failure.
- **Analytics:** no new event; this convenience default does not introduce a distinct authoritative product outcome.

## Expected result

After a successful creation, reopening New Session with the same plugin preselects the prior agent, model, and effort when they remain available. Different plugins retain independent defaults. An unavailable remembered value falls back through existing catalog defaults, and active deliberate edits continue to win.

## Verification

- `dart analyze` in `shared/sesori_shared`, `client/module_core`, and `client/app`
- `dart analyze --fatal-infos` in `bridge/app`
- Shared session-options tests: 3 passed
- Focused bridge options/creation/route/listener suites: 131 passed
- Focused module-core API/repository/options/plugin/session-detail suites: 119 passed
- Mobile New Session widget tests: 41 passed
- `git diff --check`
- Architecture implementation review: approved

## Series

1. [Clear retired prompt defaults](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1140) — merged.
2. [Add plugin-default persistence](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1142) — merged.
3. **This PR:** connect successful creation, bridge options, and client prefill behavior.

---
## Summary by cubic
Remembers the last successful agent, model, and effort selection per plugin after a session is created, and prefills the next New Session screen with it. Previously each plugin's choices reset to catalog defaults on every open.

- Adds nullable `lastUsedPromptDefaults` to the session-options response; older bridges omit it and decode as null.
- Persists the selection asynchronously after a successful creation, so a slow or failed write never delays the durable session; overlays it after loading the option catalog.
- Merges remembered defaults with any restored in-route selection, so deliberate per-dimension choices win.
- Validates remembered values against the current catalogs; removed values fall back to existing defaults, and an unavailable remembered model also drops its stored effort.

<sup>Written for commit e9c28dfebac7068d2483bdf010d6e63484db2f44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/app/lib/src/services/session_creation_service.dart
+++ b/bridge/app/lib/src/services/session_creation_service.dart
@@ -6,6 +6,7 @@ import "package:sesori_plugin_interface/sesori_plugin_interface.dart"
 import "package:sesori_shared/sesori_shared.dart";
 
 import "../repositories/models/session_operation.dart";
+import "../repositories/new_session_defaults_repository.dart";
 import "../repositories/session_metadata_repository.dart";
 import "../repositories/session_repository.dart";
 import "session_mutation_dispatcher.dart";
@@ -16,6 +17,7 @@ class SessionCreationService({
   required final SessionMetadataRepository _sessionMetadataRepository,
   required final WorktreeService _worktreeService,
   required final SessionRepository _sessionRepository,
+  required final NewSessionDefaultsRepository _newSessionDefaultsRepository,
   required final SessionMutationDispatcher _sessionMutationDispatcher,
 }) {
   final PendingOperations _lateMetadataWork = PendingOperations();
@@ -27,7 +29,17 @@ class SessionCreationService({
     // The stored path is authoritative; unknown ids are not directories.
     final projectDirectory = await _sessionRepository.resolveProjectDirectory(projectId: request.projectId);
     final normalizedCommand = request.command?.normalize();
-    final agentModel = request.model;
+    final requestedModel = request.model;
+    final promptDefaults = SessionPromptDefaults(
+      agent: request.agent,
+      model: requestedModel == null
+          ? null
+          : AgentModel(
+              providerID: requestedModel.providerID,
+              modelID: requestedModel.modelID,
+              variant: request.variant?.id,
+            ),
+    );
     final userTexts = _extractTexts(parts: request.parts);
     final firstText = userTexts.firstOrNull;
     final userVisibleText = userTexts.isEmpty ? null : userTexts.join("\n\n");
@@ -60,14 +72,8 @@ class SessionCreationService({
       branchName: worktreeState.branchName,
       baseBranch: worktreeState.baseBranch,
       baseCommit: worktreeState.baseCommit,
-      lastAgent: request.agent,
-      lastAgentModel: agentModel != null
-          ? AgentModel(
-              providerID: agentModel.providerID,
-              modelID: agentModel.modelID,
-              variant: request.variant?.id,
-            )
-          : null,
+      lastAgent: promptDefaults.agent,
+      lastAgentModel: promptDefaults.model,
     );
     await _maybeSendCommand(
       session: created,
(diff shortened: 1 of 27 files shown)


client/module_core/lib/src/services/new_session_options_service.dart
```diff
@@ -250,7 +255,7 @@ class NewSessionOptionsService({
                   ),
           ) ??
           defaultAgentModel,
-      variantIntent: restoredSelection?.variant,
+      variantIntent: effectiveSelection?.variant,
```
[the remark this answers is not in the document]
AlexDochioiu: [Sesori reply] Addressed in 048cd2271. The option resolver now reapplies a remembered variant only when the remembered model itself validates; when that model is unavailable, the fallback model keeps its own default-first variant. The focused test now uses an overlapping variant ID to cover this case.

AlexDochioiu (commented):
[Sesori reply] Addressed in 048cd2271. The option resolver now reapplies a remembered variant only when the remembered model itself validates; when that model is unavailable, the fallback model keeps its own default-first variant. The focused test now uses an overlapping variant ID to cover this case.

bridge/app/lib/src/services/session_creation_service.dart
```diff
@@ -81,6 +87,11 @@ class SessionCreationService({
       agent: request.agent,
       model: request.model,
     );
+    await _recordNewSessionDefaults(
```
AlexDochioiu: Can be done async. No need to wait (low impact if it fails). Also we should catch the error and log it, but not fail the request if this fails.

AlexDochioiu: [Sesori reply] Addressed in e9c28dfeb. The plugin-default write is now launched with `unawaited`, while `_recordNewSessionDefaults` still catches and logs any failure without affecting the creation response. Added a gated test proving `createSession` returns before persistence finishes, and kept explicit failure-log coverage.

PR #1169, opened 2026-08-28
⚙️ [grok-harness] feat(grok): compose ACP sessions and turns [step 4/9]

AlexDochioiu:
## Complexity

⚙️ Moderate — this composes the Grok ACP plugin across shared authentication, lifecycle, replay, and selection hooks while preserving existing harness defaults.

## What

- Added an optional advertised-auth allowlist to standard ACP initialization; every existing harness explicitly retains unrestricted behavior.
- Added production `GrokPlugin` composition over the shared ACP process, mapper, command, configuration, and approval machinery.
- Wired Grok identity validation, live initialize/new/load catalog capture, exact pre-turn model/reasoning selection, and one Grok provider/agent surface.
- Enabled fail-closed selection and standard stop-and-send cancellation without Grok-specific permission or event paths.
- Kept replay-process initialize validation pure and materialized replay with one atomic loaded model/provider/reasoning tuple.
- Reused shared deletion, command, reconnect, and disposal ownership rather than adding duplicate Grok cleanup paths.

## Why

Step 3 added Grok-local model contracts and collaborators but deliberately left them outside production ACP lifecycle hooks. This step connects those collaborators to the proven shared ACP implementation while keeping Grok's deprecated model API package-local and headless authentication fail-closed.

## Risk and test focus

Risk is moderate and currently dormant because Grok is not registered in the bridge app until Step 6. The highest-value checks cover auth-method filtering without regressions, live-versus-replay state ownership, stored-session effort validation after load, selection-before-prompt ordering, stop-and-send cancellation, replay identity stamping, connection reset, deletion, and idempotent disposal. There is no current user-visible or database impact, and existing ACP wire behavior is unchanged when no allowlist is supplied.

## Expected result

The Grok package now provides a complete headless ACP session/turn implementation with exact model and reasoning selection, standard permissions/events/history, safe reconnect behavior, and no activation in production routing yet.

## Verification

- `bridge/sesori_plugin_acp`: `dart test` — 270 passed
- `bridge/sesori_plugin_grok`: `dart test` — 42 passed
- `bridge/sesori_plugin_deepseek`: `dart test` — 41 passed
- ACP, Grok, DeepSeek, Copilot, Cursor, Hermes, and OMP: `dart analyze --fatal-infos` — no issues
- Dart LSP diagnostics — 0 diagnostics across 14 changed Dart files
- `git diff --check` and authored 120-column check — passed
- Two architecture implementation reviews — approved
- Post-rebase general correctness review — approved
- Scope — 1,143 changed lines against the Step 3 merge base, below the 1,500-line target

### DIFF
--- a/bridge/sesori_plugin_acp/lib/src/acp_session_loader.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_session_loader.dart
@@ -18,8 +18,8 @@ class AcpReplayCollector({
   required final String sessionId,
   required final String agentId,
 
-  /// Model/provider stamped on replayed assistant messages. Mutable so the
-  /// plugin can set the loaded session's real model after `session/load`
+  /// Model/provider/variant stamped on replayed assistant messages. Mutable so the
+  /// plugin can set the loaded session's real selection after `session/load`
   /// returns its catalog (the collector is created before the load runs).
   var String? modelId,
   var String? providerId,
@@ -39,6 +39,7 @@ class AcpReplayCollector({
 }) {
   static const AcpContentMapper _contentMapper = AcpContentMapper();
 
+  String? variant;
   final List<_Draft> _drafts = [];
   int _seq = 0;
   bool _hasUserDraft = false;
@@ -232,7 +233,7 @@ class AcpReplayCollector({
             agent: agentId,
             modelID: modelId,
             providerID: providerId,
-            variant: null,
+            variant: variant,
             errorName: halt.errorName,
             errorMessage: assistantText,
             time: draft.time,
@@ -340,7 +341,7 @@ class AcpReplayCollector({
       agent: agentId,
       modelID: modelId,
       providerID: providerId,
-      variant: null,
+      variant: variant,
       sender: PluginMessageSender.agent,
       time: draft.time,
     );
--- a/bridge/sesori_plugin_acp/lib/src/acp_session_loader.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_session_loader.dart
@@ -7,6 +7,7 @@ import "repositories/trackers/acp_tool_content_tracker.dart";
 
 typedef AcpReplayUserMessageIdOverride = String? Function({required String acpMessageId});
 typedef AcpReplayMessageTimeResolver = PluginMessageTime? Function({required Map<String, dynamic> params});
+typedef _AcpReplayAssistantSelection = ({String? modelId, String? providerId, String? variant});
 
 /// Accumulates the `session/update` notifications replayed by `session/load`
 /// into ordered [PluginMessageWithParts] for `getSessionMessages`.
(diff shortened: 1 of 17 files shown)


AlexDochioiu:
[Sesori reply] Both Cursor notices report only that Bugbot could not run because its usage limit was reached; they contain no code finding or requested change. Local verification and the independent architecture/general correctness reviews remain the actionable evidence for this PR.

bridge/sesori_plugin_acp/lib/src/acp_session_loader.dart
```diff
@@ -39,6 +39,7 @@ class AcpReplayCollector({
 }) {
   static const AcpContentMapper _contentMapper = AcpContentMapper();
 
+  String? variant;
```
AlexDochioiu: Is this ever set to anything? Seems to always be null?

AlexDochioiu: [Sesori reply] Yes. `AcpPlugin.getSessionMessages` sets it from `replayVariantForSession(sessionId:)` when constructing the collector, then sets it again after `session/load` captures the loaded session state. `GrokPlugin` overrides that hook to return the tracked reasoning effort. It remains null for stock ACP harnesses that do not expose a session variant; the Grok integration test asserts replayed `variant == "max"`.

AlexDochioiu: [Sesori reply] Agreed—the public mutable fields were the wrong API. Reworked in eb02f83ffc: `AcpReplayCollector` now only accumulates replay content, and callers supply the complete model/provider/variant tuple atomically to `buildWithAssistantSelection(...)` after `session/load` settles. The public `modelId`, `providerId`, and `variant` mutation paths are gone; ACP, Grok, and DeepSeek replay tests remain green.

AlexDochioiu:
[Sesori reply] This latest Cursor notice is again only a Bugbot usage-limit failure and contains no code finding. The replay-selection refactor was verified locally across ACP, Grok, and DeepSeek tests/analyzers.

bridge/sesori_plugin_grok/lib/src/services/grok_session_options_service.dart
```diff
@@ -76,6 +80,48 @@ class GrokSessionOptionsService({
     }
   }
 
+  void validateTurnSelection({
```
AlexDochioiu, line 83: Prevalidation failures here are labeled with `GrokSessionConfigRepository.selectionOperation` (`session/set_model`), but this runs from the base `validateTurnSelection` hook for `createSession`/`sendPrompt`/`sendCommand`, before any `session/set_model` RPC is attempted. The base hook passes `operation` for exactly this, and `GrokPlugin.validateTurnSelection` currently drops it; Copilot threads the real operation into its `PluginStaleOptionsException`s. Suggest accepting `operation` here and using it for these throws so a stale selection on e.g. `sendPrompt` doesn't log/report as a failed `session/set_model`. (The `applyTurnSelection` throws are accurate as-is, since that path really is about the set_model write.)

AlexDochioiu: [Sesori reply] Fixed in a9830bf325. `GrokPlugin.validateTurnSelection` now threads the caller operation into `GrokSessionOptionsService`, and every prevalidation `PluginStaleOptionsException` uses that operation. `applyTurnSelection` still uses `session/set_model`, since that path describes the actual selection write. The stale-selection integration test now asserts `sendPrompt`, and all 42 Grok tests plus fatal-info analysis and LSP pass.

AlexDochioiu:
[Sesori reply] This Cursor notice is another Bugbot usage-limit failure with no code finding or requested change. The operation-label correction is covered by the passing Grok integration test and analyzer/LSP checks.

PR #1175, opened 2026-08-28
⚙️ [grok-harness] feat(grok): add direct-CLI setup and lifecycle [step 5/9]

AlexDochioiu:
## Complexity

⚙️ **Moderate** — this adds bounded runtime discovery and composes Grok into the shared host-process ACP lifecycle without introducing a managed installer.

## What

- Added the `GrokPluginDescriptor`, `--grok-bin`, and frozen Grok Build `1.0.5` minimum/target version.
- Added explicit-path-before-PATH selection plus bounded, ANSI-safe `grok --version` probing through `HostProcessCommandExecutor`.
- Classified missing, malformed, outdated, current, timed-out, and failed probes without inspecting credentials or starting ACP.
- Revalidated the runtime in `ensureRuntime` and provided local `grok login`/headless credential guidance; no install or update capability is advertised.
- Composed `GrokPlugin` through `hostProcessAcpFactory` and `AcpBridgePlugin`, with bounded eager connection, abort support, reconnect behavior, and owned-process shutdown.
- Added descriptor coverage for setup, provisioning, authentication isolation, crashes, reported versions, aborts, and idempotent disposal.

## Why

Initial Grok Build support deliberately uses xAI's official user-installed CLI instead of duplicating its installer or updater. The descriptor supplies the bridge lifecycle seam while keeping setup inspection read-only, credential-blind, and isolated from other harnesses.

## Risk and test focus

**Moderate risk.** The sensitive areas are command/path authority, bounded process output and timeout behavior, version evidence, authentication failure classification, and exact process ownership. Tests focus on explicit paths never falling through to PATH, malformed or unrelated version output not being trusted, ACP startup using the shared process host, Grok-only degradation, and clean shutdown/reconnect behavior.

## Expected result

Once Grok is activated in Step 6, a supported user-installed binary can be discovered or selected with `--grok-bin`, validated without reading credentials, and launched through the shared ACP lifecycle. Missing or unsupported runtimes receive actionable local guidance, while Grok failures do not affect other harnesses.

There is no database or persisted-data impact. Grok is not yet registered in the bridge app, so this step alone adds no user-visible built-in harness availability.

## Verification

- `bridge/sesori_plugin_grok`: `dart test` — 52 passed
- `bridge/sesori_plugin_grok`: `dart analyze --fatal-infos` — no issues
- Dart LSP diagnostics — 0 across the four changed Dart implementation/test files
- `git diff --check origin/main...HEAD` — passed
- Authored line-width check — no lines over 120 characters
- Scope — 845 changed lines against `origin/main`, below the 1,500-line target
- Architecture implementation review over Step 5 against Step 4 — approved with no findings

### DIFF
--- a/.plan/active/grok-harness/TRACKER.md
+++ b/.plan/active/grok-harness/TRACKER.md
@@ -3,13 +3,13 @@
 ## Current State
 
 - **Plan:** `.plan/active/grok-harness/PLAN.md`
-- **Status:** Steps 1-3/9 merged; Step 4/9 in PR
-- **Current branch:** `grok-harness-step-4-core`
-- **Base:** `origin/main` after merged PR #1160
+- **Status:** Steps 1-3/9 merged; Step 4/9 in PR; Step 5/9 implemented locally
+- **Current branch:** `grok-harness-step-5-lifecycle`
+- **Base:** `grok-harness-step-4-core` (PR #1169)
 - **Architecture plan review:** approved 2026-08-27 after catalog ownership and auth-policy corrections
 - **Merged predecessor:** #1160 — <https://github.com/sesori-ai/sesori_apps_monorepo/pull/1160>
 - **Open PR:** #1169 — <https://github.com/sesori-ai/sesori_apps_monorepo/pull/1169>
-- **Current step:** `grok-harness-step-4-core`; monitor CI and review, then advance the local Step 5 successor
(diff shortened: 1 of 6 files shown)


AlexDochioiu:
[Sesori reply] Both Cursor comments report only that Bugbot could not run because its usage limit was reached. They contain no code finding or requested change. Step 5 remains covered by 52 Grok tests, fatal-info analysis, LSP diagnostics, and an approved architecture review.

PR #1198, opened 2026-08-28
⚙️ Update dependencies and retire the go_router page workaround

AlexDochioiu:
## Complexity

Moderate. Mostly mechanical constraint bumps and lockfile regeneration, plus two
major-version upgrades that needed changelog and call-site verification, and the
removal of a routing workaround that one of those upgrades retires.

## What

Weekly dependency update across all three resolution units, plus native and Ruby
dependency graphs.

- **Toolchain**: Flutter 3.47.1 → 3.47.2, Dart 3.13.1 → 3.13.2. Environment
  constraints raised in all 25 in-scope pubspecs (`sdk: ^3.13.2`,
  `flutter: ">=3.47.2"`).
- **Dart**: `freezed` 4.0.0-dev.3 → `^4.0.0` (stable) everywhere;
  `go_router` `^17.5.0` → `^18.0.0`; `liquid_glass_widgets` `^0.30.1` → `^1.1.0`.
- **Constraint minimums** raised to resolved versions: `material_ui` `^1.1.0`,
  `cupertino_ui` `^1.0.1`, `firebase_core` `^4.14.0`, `firebase_analytics`
  `^12.5.0`, `firebase_crashlytics` `^5.3.0`, `firebase_messaging` `^16.6.0`,
  `firebase_remote_config` `^6.6.0`, `sign_in_with_apple` `^8.2.0`.
- **Native (SwiftPM)**: Firebase 12.17.0 → 12.18.0, GoogleUtilities 8.1.2 →
  8.1.3, across all four `Package.resolved` copies.
- **Ruby**: iOS and Android `Gemfile.lock` regenerated. `fastlane` stays at
  2.238.0 — `~> 2.238` is already current.
- **Cleanup enabled by the upgrades**: retired the go_router MaterialApp
  workaround, and dropped the now-redundant `_fe_analyzer_shared` override.

## Why

Routine weekly currency. Two items are worth calling out:

`go_router` 18.0.0 migrates to `material_ui`/`cupertino_ui`. Its `isMaterialApp`
check now resolves `MaterialApp` from `material_ui`, which is what the app
actually builds — so builder routes get real `MaterialPage`s again instead of
`NoTransitionPage`s. That is exactly the condition `app_router.dart` was waiting
on, so the `_materialPage` helper and its explicit `pageBuilder`s are gone.
go_router's own `_buildPlatformAdapterPage` supplies the same `key`, `name`,
`arguments`, and `restorationId` the helper did, so the pages are unchanged.

The `_fe_analyzer_shared: 103.0.0` override no longer does anything: resolution
lands on 103.0.0 without it, because `freezed` 4.0.0 still caps `analyzer` at
`^13.0.0`. Its TODO condition (analyzer 14 support) is still unmet, but keeping a
no-op pin only risks masking the resolution once that support lands.

## Risk and test focus

**User-visible impact**: none intended. The routing change is the only one that
touches runtime behavior, and it is page-for-page equivalent. Worth a manual pass
on push/pop transitions: projects → sessions → session detail, the settings
full-screen modal, and harness settings reached both from the settings list
(pushes) and from the new-session harness menu (rises as a modal).

`liquid_glass_widgets` 1.x removes a number of APIs (`GlassBottomBar`,
`GlassTabBarCollapseConfig`, `enablePeek`, `respectsAccessibility`,
`usesBackdropFilter`, `LiquidGlassScope.stack`, and others). None are referenced
anywhere in `client/`, so the bump is a no-op at the call sites.

**Database impact**: none. No schema, migration, or persisted-shape changes.
`drift` is deliberately unchanged.

## Expected result

Same behavior on a current toolchain and native SDK set, with one workaround and
one redundant override removed.

## Deferred

| Dependency | Current | Latest | Reason |
|---|---|---|---|
| `drift_dev` | 2.34.3 | 2.34.5 | `drift` runtime is still 2.34.3. Both are pinned to the same patch on purpose — skew breaks the Drift schema verifier. |
| `test` (client only) | 1.31.1 | 1.31.2 | `flutter_test` from the SDK pins `test_api` 0.7.12; 1.31.2 needs 0.7.13. Bridge is already on 1.31.2. |
| `analyzer` | 13.3.0 | 14.1.0 | `freezed` 4.0.0 requires `analyzer ^13.0.0`; `analyzer_buffer` 0.3.3 caps at `<14.0.0`. |

## Verification

- `make analyze` — clean in `shared`, `bridge`, `client` (client runs with `--fatal-infos`).
- `make test` — `shared` 208 passed, `bridge` 2798 passed, `client` all members
  passed (`app` 914; one fewer than before, the deleted guard test).
- `make codegen` — clean in all three; only real output change is
  `GeneratedPluginRegistrant.swift`, from the `firebase_crashlytics` 5.3.0 plugin
  class rename.
- SwiftPM project and workspace pins verified identical per platform.
- `bundle exec fastlane --version` reports 2.238.0 for both iOS and Android.

## Skill maintenance

`.ai/skills/update-dependencies/SKILL.md` was stale and is updated:

Three bridge plugins were missing from every per-package step —
`sesori_plugin_copilot`, `sesori_plugin_grok`, `sesori_plugin_deepseek`. The
Phase 0 preflight is what caught them.

### DIFF
--- a/.ai/skills/update-dependencies/SKILL.md
+++ b/.ai/skills/update-dependencies/SKILL.md
@@ -36,6 +36,9 @@ The repo has two Dart workspaces and two standalone packages. Workspace members 
 - `bridge/sesori_plugin_claude/pubspec.yaml`
 - `bridge/sesori_plugin_pi/pubspec.yaml`
 - `bridge/sesori_plugin_hermes/pubspec.yaml`
+- `bridge/sesori_plugin_copilot/pubspec.yaml`
+- `bridge/sesori_plugin_grok/pubspec.yaml`
+- `bridge/sesori_plugin_deepseek/pubspec.yaml`
 - `bridge/app/pubspec.yaml` (CLI relay server)
 
 **Standalone packages** (NOT in any workspace — resolve independently with their own lockfile):
@@ -119,6 +122,16 @@ flutter --version
 ```
 
 If `flutter --version` still reports the old version after `asdf set`, run `asdf reshim flutter` (or open a new shell) and re-run `flutter --version`.
+
+Two recurring environment traps on this machine (both hit 2026-08-28):
+
+- **The shell profile puts a pinned `~/.asdf/installs/flutter/<old>-stable/bin` on `PATH` ahead of `~/.asdf/shims`**, so `which flutter` keeps resolving the old SDK no matter what `asdf set` writes. `asdf current flutter` will correctly report the new version while `flutter --version` reports the old one. Prefix every `flutter`/`dart` command in this workflow with `export PATH="$HOME/.asdf/shims:$PATH"` (the Makefile targets included) and verify with `which flutter` before trusting any version output.
+- **`asdf install flutter latest` reports "already installed" for a half-extracted install.** If the version dir contains only `flutter_macos_*.zip`, or `flutter --version` fails with `fatal: not a git repository`, the extraction was interrupted. `asdf uninstall` alone is not enough — the plugin's `mv`-based install then fails with "Directory not empty". Fully remove both dirs and reinstall:
+
+  ```bash
+  rm -rf ~/.asdf/installs/flutter/<version> ~/.asdf/downloads/flutter/<version>
+  asdf install flutter <version>
+  ```
 </step>
 
 <step name="1.2">Check and update environment constraints in all pubspec.yaml files **except `shared/no_slop_linter/pubspec.yaml`** (manually managed).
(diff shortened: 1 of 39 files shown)


.ai/skills/update-dependencies/SKILL.md
```diff
@@ -119,6 +122,16 @@ flutter --version
...
+
+Two recurring environment traps on this machine (both hit 2026-08-28):
+
+- **The shell profile puts a pinned `~/.asdf/installs/flutter/<old>-stable/bin` on `PATH` ahead of `~/.asdf/shims`**, so `which flutter` keeps resolving the old SDK no matter what `asdf set` writes. `asdf current flutter` will correctly report the new version while `flutter --version` reports the old one. Prefix every `flutter`/`dart` command in this workflow with `export PATH="$HOME/.asdf/shims:$PATH"` (the Makefile targets included) and verify with `which flutter` before trusting any version output.
+- **`asdf install flutter latest` reports "already installed" for a half-extracted install.** If the version dir contains only `flutter_macos_*.zip`, or `flutter --version` fails with `fatal: not a git repository`, the extraction was interrupted. `asdf uninstall` alone is not enough — the plugin's `mv`-based install then fails with "Directory not empty". Fully remove both dirs and reinstall:
+
+  ```bash
+  rm -rf ~/.asdf/installs/flutter/<version> ~/.asdf/downloads/flutter/<version>
+  asdf install flutter <version>
+  ```
```
AlexDochioiu: remove this entirely, not needed here

AlexDochioiu: [Sesori reply] Removed the whole block in 329f526796 — the surrounding `asdf reshim` line already covers the stale-version case, and the rest was specific to one machine rather than to this workflow.

PR #1217, opened 2026-08-31
🌿 Fix registered-bridge state leaking across logout

AlexDochioiu:
## Complexity

🌿 Straightforward: add auth-generation fences to existing asynchronous latch reads and regression coverage.

## What

- Fence the reactive startup seed and store-backed resolution so results started before logout cannot update the service.
- Fence the persisted store cache so a delayed read cannot rehydrate the previous account's latch.
- Add deterministic logout-race tests for the service and store.
- Document the logout invariant in the account and onboarding regression contract.

## Why

The registered-bridge service starts a persisted-latch read from its app-wide builder. If logout completes while that read is pending, the old implementation emitted `true` afterward, allowing the signed-out or next account to enter the bridge-offline recovery flow. The store could also retain the delayed result in its in-memory cache.

## Risk and test focus

There is no wire-contract, database-schema, or migration impact. The change only discards asynchronous registered-bridge results whose auth generation is no longer current. Focus is on logout ordering, persisted-latch reads, cache state, and preserving existing positive-latch behavior.

## Expected result

Logging out while a registered-bridge read is in flight leaves the service and store unlatched for the next account. Reads completed within the current auth generation retain their existing behavior.

## Verification

- The new startup-race test failed before the fix with `Expected: false; Actual: true` and passes afterward.
- `dart test` from `client/module_core`: 1,466 tests passed.
- `dart analyze` from `client/module_core`: passed.
- `flutter test` from `client/app`: 913 tests passed.
- `flutter analyze` from `client/app`: passed.

---
## Summary by cubic
Fixes a logout race where a registered-bridge read that finishes after logout could restore the previous account's latch, causing the signed-out or next account to enter the bridge-offline recovery flow.

- Discards persisted-latch reads started before logout in the service and store.
- Adds regression tests for the service and store logout races.
- Documents the logout invariant in the account and onboarding regression contract.

<sup>Written for commit 82f7cb715c9d15994e32c9698faefe6acbad56b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/module_core/lib/src/repositories/registered_bridges_store.dart
+++ b/client/module_core/lib/src/repositories/registered_bridges_store.dart
@@ -43,11 +43,18 @@ class RegisteredBridgesStore({
   /// registered bridge; `false` means "not yet known" — never "no bridges".
   bool _knownRegistered = false;
 
+  /// Bumped on logout so a storage read started for the previous account
+  /// cannot repopulate this cache after the account changes.
+  int _authGeneration = 0;
+
   this {
     _authSubscription = authSession.authStateStream.listen((state) {
       // Logout: drop the latch so the next account starts from scratch.
       // clear() handles its own errors, so this stays fire-and-forget.
-      if (state is AuthUnauthenticated) unawaited(clear());
+      if (state is AuthUnauthenticated) {
+        _authGeneration++;
+        unawaited(clear());
+      }
     });
   }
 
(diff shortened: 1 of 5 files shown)


AlexDochioiu:
[Sesori reply] Cursor Bugbot could not run because its usage limit was reached; this comment contains no actionable review feedback.

AlexDochioiu:
[Sesori reply] The Codex review is still running; I will address any findings when it completes.

PR #1251, opened 2026-09-01
🌱 Leave the macOS native spinner in its system colour

AlexDochioiu:
## Complexity

🌱 trivial — removes the tint block from the macOS native activity indicator view and adjusts one regression-doc sentence.

## What

The macOS `NSProgressIndicator` inside `NativeActivityIndicatorView` no longer receives a `CIColorMonochrome` content filter, the pinned dark appearance, or layer backing; it draws in its normal system appearance colour. The platform-view factory still validates the shared colour creation argument (iOS keeps using it via `UIActivityIndicatorView.color`) but ignores it on macOS. The now-unused `CoreImage` import is dropped, and the regression doc records that the macOS native spinner intentionally keeps its system colour.

## Why

After #1246 made the tint actually apply in the light appearance, the brand-blue spinner turned out to be barely visible and looked worse than the system-grey indicator that macOS users expect. The owner prefers the untinted native look.

## Risk and test focus

Low. Impacted: every native spinner on macOS (session detail busy state, pickers, loading views). Not impacted: iOS tinting, the Android/web/desktop stepped Flutter spinner, sizing, reduce-motion handling, semantics. Test focus: on macOS in both system appearances the spinner renders in the system colour and still animates; no build warnings from the removed import.

## Expected result

User-visible: macOS spinners return to the standard system-grey appearance. No database change; no wire change (the creation argument is unchanged, only unused on macOS).

## Verification

- `flutter build macos --debug` in `client/app` succeeds with the change.
- Visual confirmation on a Mac is the authoritative check.

---
## Summary by cubic
Removes the tint applied to the macOS native activity indicator so it renders in the system default grey instead of the brand blue. The blue spinner was barely visible and looked worse than the native look, so the shared colour argument is validated but not used on macOS; iOS tinting is unaffected.

<sup>Written for commit 61d2ed745b94f6de3f9f579f26dcb174274c01ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1251?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/module_prego/darwin/theme_prego/Sources/theme_prego/ThemePregoPlugin.swift
+++ b/client/module_prego/darwin/theme_prego/Sources/theme_prego/ThemePregoPlugin.swift
@@ -5,7 +5,6 @@ import Foundation
   import UIKit
 #elseif os(macOS)
   import AppKit
-  import CoreImage
   import FlutterMacOS
 #endif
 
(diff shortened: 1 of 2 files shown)


AlexDochioiu:
Superseded: the intent is to stop using the tint everywhere while keeping the capability, not to remove macOS tinting.

PR #1277, opened 2026-09-03
⚙️ Keep active ACP prompts alive

AlexDochioiu:
## Complexity

⚙️ Moderate — this changes shared NDJSON request timeout bookkeeping while keeping the new behavior opt-in for long-running ACP prompts.

## What

- Add an activity-aware timeout mode to the shared NDJSON process client.
- Make ACP `session/prompt` use a 30-minute inactivity deadline that resets only for inbound frames attributed to the same session.
- Keep ordinary RPC requests on their existing fixed wall-clock timeout.
- Add regression coverage and document the behavior and failure signals.

## Why

An OMP prompt remained active and continued emitting tool/reasoning updates beyond 30 minutes, but the bridge applied a fixed 30-minute response timeout and persisted a false prompt failure before OMP later completed normally.

## Risk and test focus

The activity matcher must not let traffic from another concurrently running session keep a silent request alive. Response correlation, true inactivity, and immediate process-exit failures must continue to settle pending requests correctly.

User-visible impact: active ACP turns no longer show a false timeout merely because total turn duration exceeds 30 minutes. Database/schema impact: none.

## Expected result

An ACP prompt may run longer than 30 minutes while its own session continues producing protocol activity. It still times out after 30 minutes without matching activity, and unrelated session traffic does not extend that deadline.

## Verification

- `cd bridge/sesori_plugin_acp && dart test test/acp_stdio_client_test.dart`
- `cd bridge/sesori_plugin_acp && dart test test/acp_turn_serialization_test.dart`
- `cd bridge/sesori_plugin_acp && dart analyze`
- `cd bridge/sesori_plugin_runtime && dart test test/ndjson_process_client_test.dart`
- `cd bridge/sesori_plugin_runtime && dart analyze`
- `git diff --check`

---
## Summary by cubic
Fixes ACP session prompts falsely timing out after 30 total minutes even when the turn is still producing activity. The `session/prompt` deadline now measures same-session inactivity instead of wall-clock time, and unrelated concurrent session traffic does not reset it.

- Adds an inactivity-based dispatch mode to the shared NDJSON process client; ordinary RPC requests keep their fixed timeout.
- Adds regression coverage and updates the regression docs with the new failure signals.

<sup>Written for commit c0f7faa2bd6f897d2866a296a8e9e3e97b2f0b1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/sesori_plugin_acp/lib/src/acp_plugin.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_plugin.dart
@@ -1380,14 +1380,15 @@ abstract class AcpPlugin({
         queuedPrompt.phase = _QueuedAcpPromptPhase.writing;
       }
       final meta = outboundPromptMeta(sessionId: sessionId, messageId: turn.messageId);
-      final dispatched = await client.dispatchRequest(
+      final dispatched = await client.dispatchSessionRequest(
         method: AcpMethods.sessionPrompt,
         params: {
           "sessionId": sessionId,
           "prompt": turn.blocks,
           "_meta": ?meta,
         },
-        timeout: const Duration(minutes: 30),
+        sessionId: sessionId,
+        inactivityTimeout: const Duration(minutes: 30),
       );
       _markTurnDispatched(sessionId: sessionId, state: state, turn: turn);
       final raw = await dispatched.response;
--- a/bridge/sesori_plugin_acp/lib/src/acp_stdio_client.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_stdio_client.dart
@@ -82,14 +82,57 @@ class AcpStdioClient({
     required String method,
     Object? params,
     Duration timeout = const Duration(seconds: 60),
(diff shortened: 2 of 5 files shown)


AlexDochioiu:
[Sesori reply] Reviewed the lint suppressions. They all cover the intentional opaque JSON-RPC correlation-ID boundary: protocol IDs may be different non-null JSON scalar types, so narrowing them would be incorrect. Each suppression is paired with the boundary rationale; no code change is needed.

PR #1329, opened 2026-09-05
⚙️ [periodic-cleanup] auth: share response and interactive login completion [step 17/25]

AlexDochioiu:
## Complexity

⚙️ Moderate. Two local folds inside the auth HTTP client and `AuthManager`, in
security-sensitive code whose generation fencing and persistence ordering must
survive unchanged.

## What

- `HttpApiClient._toApiResponse` becomes the single HTTP-response-to-`ApiResponse`
  branch for both ordinary and multipart requests. Non-2xx keeps its status code
  and raw body, an empty 2xx body still reaches `fromJson(null)`, and a decode or
  parse failure still logs the typed `_JsonResponseParsingException` — now under a
  per-operation label — before returning `ApiError.jsonParsing`.
- `AuthManager._completeInteractiveLogin` takes the captured generation and the
  parsed `AuthLoginResponse`, performs the existing `_persistAuthenticatedResult`
  call (OAuth state cleared, no OAuth ownership requirement) and returns
  `AuthLoginResult`. Email and Apple keep their own requests; email keeps its 401
  message and input validation.
- The shared `_parseLoginResponse` throws a file-local
  `_AuthResponseParsingException(innerError:)` instead of wrapping the cause in a
  string. Its presentation names only the cause's runtime type, because the body
  may carry tokens.

OAuth polling, token refresh and OAuth ownership are untouched.

## Why

Step 17 of the periodic-cleanup series. Both response branches and both
interactive login tails were duplicated line for line, so a change to empty-body
handling, parse-failure reporting or persistence ordering had to be made twice to
stay correct. The typed inner cause also replaces string-only wrapping, matching
the repository's error-translation rule.

## Risk and test focus

High sensitivity, small surface. Focus on: superseded-login fencing still throwing
and leaving no tokens; secure persistence ordering unchanged; OAuth state clearing
and returned account status unchanged; email's 401 path; non-2xx and malformed
bodies on both the ordinary and multipart client paths.

No authentication endpoint, credential format, trust posture, public API, wire
contract or database change. No user-visible change beyond identical failures now
carrying a typed cause in local logs.

## Expected result

Email and Apple sign-in behave exactly as before, including their failure and
supersession behavior. Voice's multipart upload path is unchanged.

## Verification

- `dart analyze` in `client/module_auth`: no issues.
- `dart test` in `client/module_auth`: 109 tests pass, including the new
  "rejects a malformed success body without persisting tokens" case.
- `dart test test/capabilities/voice/voice_api_test.dart` in `client/module_core`
  (multipart consumer): 8 tests pass.
- `docs/regression/account-and-onboarding.md` gains the matching failure signal.
- Architecture implementation review (required by the plan because a production
  exception type is introduced): approved, no findings.

---
## Summary by cubic
Deduplicates auth response handling and interactive login completion so email and Apple logins share code paths. Behavior is unchanged for users; the only difference is that parse failures now carry a typed cause in local logs.

- Both ordinary and multipart HTTP responses now go through one `_toApiResponse` branch; empty 2xx bodies still reach `fromJson(null)`, and non-2xx keeps its status and raw body.
- Email and Apple logins share `_completeInteractiveLogin`, which persists credentials with OAuth state cleared and no OAuth ownership requirement.
- Auth parse failures now throw a file-local `_AuthResponseParsingException` whose presentation names only the cause's runtime type, because the body may contain tokens.
- OAuth polling, token refresh, and OAuth ownership are untouched.

<sup>Written for commit 5b2c1d8fa6e5b2b3f948ce306df0dfa4443f278e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.plan/active/periodic-cleanup/TRACKER.md
+++ b/.plan/active/periodic-cleanup/TRACKER.md
@@ -23,8 +23,8 @@ asked to start working the plan; steps execute in order from step 2.
 | 13/25 | ⚙️ [periodic-cleanup] runtime: share provisioning and bounded cold-start waiting [step 13/25] | Merged | [#1323](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1323) |
 | 14/25 | 🌿 [periodic-cleanup] bridge: fold repeated worktree and Codex algorithms [step 14/25] | Merged | [#1324](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1324) |
 | 15/25 | 🌿 [periodic-cleanup] bridge: preserve caught errors and stacks in logs [step 15/25] | Merged | [#1326](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1326) |
-| 16/25 | ⚙️ [periodic-cleanup] client: share shell cubit composition [step 16/25] | In review | [#1328](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1328) |
-| 17/25 | ⚙️ [periodic-cleanup] auth: share response and interactive login completion [step 17/25] | In progress (local, awaiting step 16 merge) | — |
+| 16/25 | ⚙️ [periodic-cleanup] client: share shell cubit composition [step 16/25] | Merged | [#1328](https://github.com/sesori-ai/sesori_apps_monorepo/pull/1328) |
+| 17/25 | ⚙️ [periodic-cleanup] auth: share response and interactive login completion [step 17/25] | In review | PENDING |
 | 18/25 | 🌿 [periodic-cleanup] tests: consolidate substantial bridge fixtures [step 18/25] | Proposed | — |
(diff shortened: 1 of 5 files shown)


AlexDochioiu:
[Sesori reply] No change. These three are not new suppressions; they moved out of the two duplicated response branches into the single `_toApiResponse` helper that replaced them. The file's total goes from 14 to 13, because `postMultipart` and `_execute` no longer each carry their own `json parsing` and `json decoding` pair. Each remaining one is the pre-existing `dynamic` JSON boundary the `SafeApiClient` contract requires, with its original justification comment intact.

PR #1332, opened 2026-09-05
🌿 bridge: stop repeating the gh identity warning on hosts without gh login

AlexDochioiu:
## Complexity

Straightforward. One early return, one moved log call, one deleted redundant check.

## What

`PrSyncService.verifyGithubIdentity()` now returns null when gh is missing or
unauthenticated, without running `gh api user`, and reports a failed
verification once per failing streak instead of once per request.

`_runRefreshCycle` drops its own capability check, which the shared identity
verification now performs.

## Why

`GET /sessions` calls `verifyGithubIdentity()` on every request, and that path
never consulted the gh capability check that the refresh cycle used. Identity
failures are deliberately not cached, so a bridge on a host without `gh auth
login` spawned `gh api user`, failed, and logged a full warning plus stack trace
for every sessions poll:

```
[PrSyncService] Failed to verify the active GitHub identity; PR refresh is skipped
ProcessException: To get started with GitHub CLI, please run:  gh auth login
```

`GhCliApi` already emits one deduplicated, actionable warning for a missing
install or missing login, so the generic per-request message added noise without
adding information.

## Risk and test focus

No user-visible change and no database impact: PR-free responses on an
unauthenticated host behave exactly as before. Failed logins are still never
cached, so a fresh `gh auth login` is picked up on the next refresh — now via the
capability cache, which also only caches successes.

Focus on hosts where gh is installed and `gh auth status` passes but `gh api
user` fails (expired token, no network): the first failure still logs the error
and stack, and the streak resets on the next success.

## Expected result

A bridge without gh, or with gh not logged in, logs one console warning naming
the command to run and then stays quiet, instead of one warning and stack trace
per sessions request.

## Verification

- `dart test test/bridge/services/pr_sync_service_test.dart test/bridge/routing/get_sessions_handler_test.dart` — 60 passing, including a new test that the identity query is skipped while gh is unavailable or unauthenticated.
- `dart analyze` on the touched files — clean.

---
## Summary by cubic
Stops `verifyGithubIdentity()` from running `gh api user` when gh is missing or unauthenticated, and now reports identity verification failure only once per failing streak instead of once per request. This removes the repeated warning and stack trace on hosts without a gh login. The shared identity verification now checks gh capability first, so the refresh cycle no longer needs its own check. Failed logins are still never cached, so a fresh `gh auth login` is picked up on the next refresh.

<sup>Written for commit a1bb756526b6dcf4bffd01ec8725984d57a62134. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/app/lib/src/services/pr_sync_service.dart
+++ b/bridge/app/lib/src/services/pr_sync_service.dart
@@ -54,6 +54,9 @@ class PrSyncService({
   /// reusing a successful result for [_githubCliCacheTtl]. Each call is a
   /// network round trip, and an explicit refresh otherwise pays for several in
   /// series. Failures are never cached so a fixed login is picked up at once.
+  ///
+  /// Returns null when gh is missing or unauthenticated, without running the
+  /// identity query.
   Future<VerifiedGithubLogin?> verifyGithubIdentity() async {
     final inFlight = _identityVerification;
     if (inFlight != null) return await inFlight;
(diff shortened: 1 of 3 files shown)


AlexDochioiu:
[Sesori reply] Codex review quota was exhausted, so no findings were produced. Nothing to address from this comment; CI is the remaining gate.

AlexDochioiu:
[Sesori reply] cubic approved with no issues across the 3 changed files, and CI is green (15/15). Nothing to address.

PR #1338, opened 2026-09-05
🌿 [harness-auth-detection] omp, grok: report a logged-out install as authentication required

AlexDochioiu:
## Complexity

Straightforward. Follows #1334 exactly: one probe added to each descriptor's `inspectSetup`, reusing the existing `authenticationRequired` seam. No new types, no contract or DI changes.

## What

Extends the Pi setup probe from #1334 to the remaining harnesses whose CLI can answer the question.

**OMP** follows its version probe with `omp models --json`:
- listing parses and reports `{"models":[]}` → `PluginSetupAuthenticationRequired` with the resolved version
- listing is unparsable or exits non-zero → `PluginSetupReady` (unchanged from today)
- listing cannot be run at all → `PluginSetupUnknown`

**Grok** follows its version probe with `grok models`:
- output contains `not authenticated` → `PluginSetupAuthenticationRequired` with the resolved version
- any other wording → `PluginSetupReady`
- listing cannot be run → `PluginSetupUnknown`

**OpenCode** is reclassified in the matrix from ⬜ to 🚫 — see below.

## Why

Installing a runtime never authenticates it. Both harnesses returned ready as soon as they resolved a runtime, so an install with no credentials advertised ready and then failed every turn.

The two probes are asymmetric on purpose, and that asymmetry is the main review point:

- **OMP** treats only a *parsed empty array* as logged out. The pinned target is 17.3.8 and supported releases reach back that far, but `models --json` was only verified on 18.1.10. Mapping a non-zero exit to unknown — as Pi does — would show "undetermined" on every older install that lacks the flag. Leaving those ready preserves today's behavior.
- **Grok** treats only the explicit `not authenticated` line as logged out. The model list is byte-identical whether or not you are signed in, so the account line is the only signal, and it is English prose xAI may reword. Unrecognized wording degrades to ready.

Both follow the same principle as Pi: **only positive evidence of being logged out downgrades setup.** A false `authenticationRequired` blocks a working install, which is worse than the stale ready state being fixed.

**OpenCode is not a gap.** Per maintainer decision, its bundled free `opencode/…` models make a credential-less install genuinely usable — `opencode models` lists them identically with five credentials and with zero — so ready is the correct state there regardless of stored credentials. Marked 🚫 with that rationale rather than left as unimplemented work.

## Risk and test focus

No user-visible impact for an authenticated install of either harness. No database impact. No wire contract change — `PluginSetupState.authenticationRequired` already exists and is handled on both sides.

The deliberate behavior change: an OMP or Grok install with no credentials moves from ready to authentication-required, and `startAllowed` becomes false for it.

Worth reviewing closely:
- The Grok `inspectSetup` switch expression became a switch statement so the ready branch can await the probe. Exhaustiveness is preserved; the other three arms are unchanged.
- Grok now spawns a second process during setup inspection. The regression doc previously stated setup runs *only* a `--version` probe; that line is updated. The constraints that matter — never create a session, start ACP, or invoke login — still hold.
- OMP's fallback-to-ready on a failed listing is intentional, not an oversight.

## Expected result

A freshly installed OMP or Grok with no provider credentials reports authentication required with an action hint pointing at local login, instead of reporting ready and failing at session start. Authenticated installs are unaffected, as are older OMP releases without `models --json`.

## Verification

- `dart test` in `sesori_plugin_omp` — 59/59 pass, including four cases covering ready, empty listing, unparsable listing, and non-zero listing.
- `dart test` in `sesori_plugin_grok` — 85/85 pass, including three cases covering logged in, not authenticated, and unrecognized wording.
- `dart analyze` on both packages — no errors or warnings.
- Signals verified against real CLIs on 2026-09-05: `omp models --json` on 18.1.10 (`{"models":[]}` under an isolated profile, populated otherwise) and `grok models` on 1.0.5, which is the minimum supported version.

## Remaining after this PR

- **DeepSeek** already probes readiness in `inspectSetup` but maps a negative result to `unknown` rather than `authenticationRequired`. Not changed here; it needs the CLI to confirm the distinction is safe.
- **Copilot** has not been probed — the CLI was not available locally.

---
## Summary by cubic
Makes OMP and Grok report a logged-out install as `authenticationRequired` instead of ready, so an unauthenticated backend no longer advertises ready and then fails at session start. OpenCode is documented as intentionally not supported because its bundled free models make an unauthenticated install usable.

- OMP probes `omp models --json`; only a parsed empty model array downgrades setup.
- Grok probes `grok models`; only the explicit "not authenticated" line downgrades setup, since the model list is identical either way.
- Unrecognized output falls back to ready on both harnesses to avoid blocking working installs.
- Authenticated installs are unaffected; older OMP releases without `models --json` stay ready.

<sup>Written for commit e54c56d7232f24efa056ec1fa50dd4a5f14fffb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1338?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bridge/sesori_plugin_grok/lib/src/runtime/grok_plugin_descriptor.dart
+++ b/bridge/sesori_plugin_grok/lib/src/runtime/grok_plugin_descriptor.dart
@@ -109,22 +109,66 @@ class const GrokPluginDescriptor() extends BridgePluginDescriptor {
       environment: environment,
     );
 
-    return switch (runtime) {
-      _GrokRuntimeReady(:final version) => PluginSetupReady.versioned(runtimeVersion: version),
-      _GrokRuntimeMissing() => PluginSetupRuntimeMissing(
-        actionHint: explicitBin == null
-            ? "Install Grok Build with xAI's official installer, then restart the bridge."
-            : "Fix the configured Grok Build binary path, then restart the bridge.",
-      ),
-      _GrokRuntimeOutdated() => PluginSetupUnavailable(
-        actionHint: explicitBin == null
-            ? "Update Grok Build with xAI's official installer, then restart the bridge."
-            : "Update the configured Grok Build binary or fix `--grok-bin`, then restart the bridge.",
-      ),
-      _GrokRuntimeUnknown() || _GrokRuntimeUnrecognized() => const PluginSetupUnknown(
-        actionHint: "Grok Build setup could not be verified. Run `grok --version` locally, then retry.",
-      ),
-    };
+    switch (runtime) {
+      case _GrokRuntimeReady(:final version):
+        return await _inspectAuthentication(
+          executablePath: explicitBin ?? GrokBinary.defaultBinary,
+          processes: processes,
+          environment: environment,
+          runtimeVersion: version,
+        );
+      case _GrokRuntimeMissing():
+        return PluginSetupRuntimeMissing(
+          actionHint: explicitBin == null
+              ? "Install Grok Build with xAI's official installer, then restart the bridge."
+              : "Fix the configured Grok Build binary path, then restart the bridge.",
+        );
+      case _GrokRuntimeOutdated():
+        return PluginSetupUnavailable(
+          actionHint: explicitBin == null
+              ? "Update Grok Build with xAI's official installer, then restart the bridge."
+              : "Update the configured Grok Build binary or fix `--grok-bin`, then restart the bridge.",
+        );
+      case _GrokRuntimeUnknown() || _GrokRuntimeUnrecognized():
+        return const PluginSetupUnknown(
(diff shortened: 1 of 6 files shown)


AlexDochioiu:
[Sesori reply] cubic approved with no issues across 6 files, CI 15/15. The Codex comment is a usage-limit notice rather than review feedback, so there is nothing to address. Marking ready for human review.

PR #1379, opened 2026-09-08
🌿 Prevent concurrent NDJSON dispatch from crashing the bridge

AlexDochioiu:
## Complexity
🌿 Straightforward transport-local correction: remove unsafe per-request flush and observe write failures at the existing owner. No queue, lock, registry, or API expansion.

## What
- Admit requests and notifications synchronously through the existing FIFO IOSink instead of binding it with per-request `flush()`.
- Observe the pending response before sending, preventing unhandled completer errors on synchronous write failure.
- Propagate/log original asynchronous stdin failures, with existing attachment-generation isolation.
- Add real-SDK IOSink regressions; replace obsolete fake-flush ACP assertions with stdin-admission coverage and correct documentation.

## Why
Live DeepSeek phone QA reproduced a bridge process crash when Stop dispatched multiple native scopes. The first request's flush bound the sink; the next write threw `Bad state: StreamSink is bound to a stream`. Its pending completer also failed before an error observer was attached.

## Risk and test focus
Shared transport serves ACP harnesses and Codex. Focus is FIFO request/notification admission, synchronous and asynchronous write errors, prompt/event ordering, and old-sink isolation after reset. Dispatch now acknowledges sink admission rather than OS drain; later pipe errors remain observable and fail pending requests.

No database or wire-contract impact. User-visible impact: concurrent Stop no longer crashes the bridge. Native Stop policy and adapter version are unchanged.

## Expected result
Multiple scoped STOP requests can be admitted before later prompts without illegal concurrent IOSink writes or unhandled pending errors. The bridge remains usable afterward.

## Verification
- Runtime package tests passed, including 17 focused transport tests using a real SDK IOSink.
- 45 targeted ACP tests and the full DeepSeek package (116 tests) passed.
- Runtime, ACP, and DeepSeek fatal-info analysis clean; whitespace checks clean.
- CI exposed one remaining DeepSeek fake-flush assumption. Its regression now injects old-input/cancel/reused-input frames at actual prompt admission and checks user-message publication precedes the question, retaining cancellation and replacement-input assertions. No production changes in that follow-up.
- Independent correctness/simplicity review approved; stale comment corrected.
- Real iOS simulator, relay, and DeepSeek 0.1.4 / Flash: independently resumed child's main turn settled while its grandchild ran; ancestor Stop from the phone settled all three sessions idle without exiting bridge/native runtime.
- A subsequent prompt produced the expected assistant marker and settled idle without a session error.
- Native root-owned background shell jobs remained after agent-turn cancellation; this PR does not change that behavior. Desktop and the broader question-cancellation matrix were not tested in this bug-fix run.

### DIFF
--- a/bridge/sesori_plugin_acp/lib/src/acp_plugin.dart
+++ b/bridge/sesori_plugin_acp/lib/src/acp_plugin.dart
@@ -148,10 +148,9 @@ abstract class AcpPlugin({
   /// decisions live on this class — the state object only holds fields.
   final Map<String, _SessionTurnState> _turnStates = {};
 
-  /// Agent updates and server requests that raced the stdin flush for an
-  /// accepted existing-session prompt. The user message cannot publish before
-  /// the flush succeeds, so hold both until that message and any preceding tool
-  /// update have entered the event stream.
+  /// Agent updates and server requests that raced stdin admission for an
+  /// accepted existing-session prompt. Hold both until the admitted user
+  /// message and any preceding tool update have entered the event stream.
   final Map<String, List<AcpNotification>> _promptWriteNotifications = {};
   final Map<String, List<AcpServerRequest>> _promptWriteServerRequests = {};
   final Set<String> _cancelledPromptWriteSessions = {};
--- a/bridge/sesori_plugin_acp/test/acp_turn_serialization_test.dart
+++ b/bridge/sesori_plugin_acp/test/acp_turn_serialization_test.dart
@@ -97,35 +97,6 @@ class _PromptOrderedChildMapper({
   }
 }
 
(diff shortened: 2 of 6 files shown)


AlexDochioiu:
[Sesori reply] Fixed the CI failure in e94031d. The remaining DeepSeek test assumed a held fake flush delayed input handling. It now injects old question → cancellation → reused-ID question synchronously at prompt stdin admission, asserts user-message publication precedes the question, and retains the original old-input rejection/cancel ACK/new-input survival checks. Full DeepSeek package: 116 tests passed; fatal-info analysis clean. No production changes in this follow-up. Inspected Codex completion and Cubic approval: no actionable review findings.

PR #1436, opened 2026-09-11
⚙️ Synchronize AI sparkles with a 40 fps Flutter loop

AlexDochioiu:
## Complexity
⚙️ Moderate — shared Flutter animation timing plus synchronized iOS/macOS Core Animation, with focused lifecycle and presentation tests.

## What
- Synchronize working AI sparkles using a common two-second device-clock phase instead of per-session/project offsets.
- Drive the steady Flutter loop at 40 fps (25 ms) through one shared timer, with no repeating ticker or inter-step frame scheduling.
- Keep native animation cadence unchanged and preserve the finite 700 ms completion and 150 ms restart/rejoin transitions.
- Remove the internal row-phase API and update regression documentation and preview copy. Cupertino/activity spinners and the caller-owned Deep Scan timeline remain unchanged.

## Why
Multiple working sessions should rotate together rather than distract with staggered indicators. The lower Flutter loop cadence reduces scheduled animation work while keeping a higher cadence than the initially proposed 30 fps.

## Risk and test focus
UI-only, medium regression risk around staggered mounts, interrupted completion, background/resume, reduced motion, and native view lifecycle. No database, transport, backend, or account-state impact.

Device-level visual smoothness, native compositing/lifecycle behavior, and battery savings were not measured here. The existing thread-state motion playbook is the manual review target; the 40 fps cap applies only to the steady Flutter loop, not finite transitions.

## Expected result
Working sparkles in session/project lists visibly stay in sync, including rows mounted later. Flutter's steady loop updates at 40 fps; Apple keeps unthrottled Core Animation. Unread, reduced-motion, and inactive indicators retain their existing meaning. Other indicator types remain independent.

## Verification
Using repository-pinned Flutter 3.47.3 / Dart 3.13.3:
- `client/module_prego`: AI loader and activity indicator widget suites passed.
- `client/module_app_ui`: session tile state and catalog scan row widget suites passed.
- `flutter analyze`: `module_prego`, `module_app_ui`, and the app thread-state motion playbook passed.
- Native `ThemePregoPlugin.swift` type-check passed for macOS and iOS Simulator against the matching Flutter frameworks (not a full app build or GUI run).
- Architecture plan and implementation reviews approved; `git diff --check` passed.

---
## Summary by cubic
Synchronizes working AI sparkles so rows rotate together instead of at staggered per-row phase offsets.

**Refactors**
- Flutter's steady loop now runs at 40 fps from one shared device-clock-aligned timer; no repeating ticker or inter-step frames, and the 700 ms completion and 150 ms restart/rejoin transitions are unchanged.
- Native iOS/macOS Core Animation keeps its unthrottled cadence.
- Removes `PregoAiLoader.phase` and `phaseFor`; session and project tiles no longer seed a phase.
- Adds the `clock` package for deterministic device-clock sampling in tests.

<sup>Written for commit c62b104458c1eb51aae46027fb535f36a20ea467. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sesori-ai/sesori_apps_monorepo/pull/1436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/module_prego/darwin/theme_prego/Sources/theme_prego/ThemePregoPlugin.swift
+++ b/client/module_prego/darwin/theme_prego/Sources/theme_prego/ThemePregoPlugin.swift
@@ -70,11 +70,10 @@ private struct ActivityIndicatorCreationParams {
   }
 }
 
-/// Theme colours, initial state, and phase supplied by the Flutter decoration.
+/// Theme colours and initial state supplied by the Flutter decoration.
 private struct AiLoaderCreationParams {
   let solid: CGColor
   let outline: CGColor
-  let phase: Double
   let loading: Bool
 
   init?(from args: Any?) {
@@ -82,12 +81,10 @@ private struct AiLoaderCreationParams {
       let dictionary = args as? [String: Any],
       let solid = dictionary["solid"] as? NSNumber,
       let outline = dictionary["outline"] as? NSNumber,
-      let phase = dictionary["phase"] as? NSNumber,
       let loading = dictionary["loading"] as? Bool
     else { return nil }
     self.solid = ColorComponents(fromARGB: solid.int64Value).cgColor
     self.outline = ColorComponents(fromARGB: outline.int64Value).cgColor
-    self.phase = phase.doubleValue
     self.loading = loading
   }
 }
@@ -98,6 +95,7 @@ private enum AiLoaderSparkle {
   static let viewBox: CGFloat = 20
   static let period: CFTimeInterval = 2
   static let completionDuration: CFTimeInterval = 0.7
+  static let resumeDuration: CFTimeInterval = 0.15
   static let clear = ColorComponents(fromARGB: 0x00b2d1ff).cgColor
   static let completionHighlight = ColorComponents(fromARGB: 0xffb2d1ff).cgColor
   static let colourTiming = CAMediaTimingFunction(controlPoints: 0.5, 0, 0.5, 1)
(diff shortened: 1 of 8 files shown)


client/module_prego/darwin/theme_prego/Sources/theme_prego/ThemePregoPlugin.swift
```diff
@@ -111,57 +109,91 @@ private enum AiLoaderSparkle {
...
+    strokeAnimation.duration = loading ? resumeDuration : completionDuration
     strokeAnimation.timingFunction = colourTiming
     layer.add(strokeAnimation, forKey: "stroke")
     CATransaction.commit()
   }
 
+  /// Start an unthrottled Core Animation loop at current Unix-clock phase.
+  /// Sampling here, rather than from Flutter creation arguments, also aligns
+  /// views mounted later and views resuming after suspension.
+  static func startSynchronizedLoading(layer: CAShapeLayer, params: AiLoaderCreationParams) {
```
AlexDochioiu, line 171: we should NEVER throttle the animation for the native implementation of any widget. only the flutter implementation must be throttled

AlexDochioiu: [Sesori reply]
Agreed. Only Flutter is throttled in this PR: the 40 fps / 25 ms timer lives in Dart’s `_AiLoaderLoopClock`, and native-rendered loaders never subscribe to it.

iOS/macOS still use continuous, unthrottled `CABasicAnimation`. `beginTime` here aligns the starting phase; the unchanged two-second duration is the rotation speed, not a frame-rate cap. No native throttling was introduced, so no code change is needed for this point. I will preserve this rule for every widget.

PR #1479, opened 2026-09-14
🚧 [queued-composer] Make queued prompt cancellation truthful [step 2/2]

daniil-shumko:
## Complexity
🚧 Complex — clarifies dispatch ownership across plugin, wire, client-state and UI boundaries without adding another queue or lifecycle owner. 706 changed lines: 613 authored and 93 generated.

Step 2/2 of `queued-composer`; #1478 is merged. This PR now targets `main` and contains only the cancellation fix.

## What
- Derive queued/dispatched state from the existing Pi, Claude and ACP dispatch flags, and publish it through the shared prompt contract.
- Keep best-effort trash for bridge-pending and unknown/legacy entries. Explicitly dispatched messages show Sending without trash.
- Remove client queue rows only after confirmed cancellation. Refusals report a localized notice without hiding still-live prompts; existing queue events and reconnect snapshots remain authoritative.
- Coalesce repeated cancellation taps per prompt while the request is pending, releasing the guard on every outcome so failures remain retryable.
- Derive Pi presentation directly from the existing turn dispatch flag rather than storing a stale presentation state.
- Keep immediate steering, existing send ordering and the 3.5-row composer layout unchanged.
- Document the native per-item cancellation limitation and add ownership/refusal coverage.

## Why
The user found that cancelling a Pi queued message hid it, but it still executed later. Same-selection follow-ups had already been handed to Pi's native steering queue while still looking cancellable. Pi refused individual cancellation; the client treated that 404 refusal as successful deletion. This fixes the misleading controls and outcome without removing the user-approved steering behavior.

## Risk and test focus
Cross-layer risk: dispatch-boundary updates, prompt correlation, repeated taps, refused versus successful cancellation, and mixed client/bridge versions. The wire field is additive; missing or future values decode to unknown. Public v1.8.0–v1.8.3 bridges retain best-effort cancellation rather than losing an existing feature. Cancellation outcomes, not a missing dispatch hint, determine success.

No native clear-all/replay, new queue, persisted state, dependency, database, or analytics changes. One short-lived in-flight ID set suppresses duplicate requests; no reconciliation read, generation counter, or extra lifecycle owner.

## Expected result
- A prompt still waiting in Sesori can be cancelled and never reaches the backend.
- A prompt already forwarded to a harness shows Sending without trash until its user message appears. Pi continues to steer immediately.
- A stale/racing cancellation reports failure rather than silently disappearing; existing bridge events update its state. Repeated taps during one request do not produce a false warning.
- Older bridges retain their cancellation action. If an entry is already dispatched, the refusal is surfaced honestly without claiming deletion.

## Verification
- Shared model/legacy decoding: 10 tests passed.
- Pi session service: 58 passed, including accepted native steering and true pre-dispatch cancellation.
- Claude session service: 40 passed; ACP turn serialization: 31 passed.
- Bridge queued-prompt routing/mapping: 10 passed.
- Client core queue/cancellation: 41 passed, including repeated taps, retry after returned/thrown failures, authoritative dispatch/removal events, and no redundant refresh.
- Prego 3.5-row geometry/text scaling: 2 passed.
- Mobile body: all 120 tests passed, including actionable legacy/unknown rows, read-only dispatched rows, cancellation notices, draft/focus preservation, and unchanged composer geometry.
- Status tooltips retain their visual hints without duplicating row semantics; focused mobile coverage passes.
- All ten owning package analyzers pass; the three affected client analyzers and Pi analyzer were rerun after human review fixes.
- Independent architecture plan and implementation reviews approved the existing ownership boundaries; follow-up changes stay within those owners.
- `git diff --check` passes. No live authenticated harness/device E2E run claimed.


client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -1495,16 +1495,26 @@ class SessionDetailCubit(
...
     if (result case ErrorResponse(:final error)) {
-      if (error is! NonSuccessCodeError || error.errorCode != 404) return;
+      logw("Queued prompt cancellation was not confirmed for $_sessionId/$promptId", error);
+      _noticeStream.add(const SessionDetailQueueCancellationFailed());
+      if (error is NonSuccessCodeError && error.errorCode == 404) {
+        final refreshed = await _sessionRepository.getQueuedPrompts(sessionId: _sessionId);
+        if (isClosed) return;
+        switch (refreshed) {
+          case SuccessResponse(:final data):
+            _onBridgeQueueUpdated(data.data);
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Fixed in 29124983d. The refresh captures the existing immutable queue snapshot and applies its result only if the current queue still matches. Newer dispatch/removal events therefore win, while unrelated streaming transcript updates do not discard useful reconciliation. No persistent generation counter or extra queue owner was added. Both delayed-GET regression scenarios failed before the fix; the complete core queue suite now passes 41 tests.

client/module_app_ui/lib/src/features/session_detail/widgets/session_detail_prompt_queue.dart
```diff
@@ -30,18 +31,38 @@ class const SessionDetailPromptQueue({
...
+          deliveryStatus?.label ??
+          (unavailable
+              ? loc.sessionDetailUnavailableCommand
+              : isCommand
+              ? loc.sessionDetailQueuedCommand
+              : loc.sessionDetailQueuedMessage),
       warning: unavailable ? loc.sessionDetailUnavailableCommand : null,
-      removeLabel: unavailable ? loc.sessionDetailRemoveQueued : loc.sessionDetailCancelQueued,
-      onRemove: onRemove,
+      trailing: deliveryStatus != null
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Fixed in 29124983d with excludeFromSemantics on delivery-status tooltips. The row remains the one status-label source; the visual tooltip remains available and cancellation-action tooltips are unaffected. Focused mobile coverage verifies the retained tooltip widget, excluded duplicate semantic tooltip, and Sending row label.

daniil-shumko:
[Sesori reply] Both review findings are addressed in 29124983d with regression evidence: newer live queue state wins over a delayed reconciliation read, and status tooltips no longer duplicate accessibility labels. Core queue suite: 41 passed; focused mobile status test passed; core/shared UI/mobile analyzers pass. Diff remains 673 lines (570 authored, 103 generated). CI is rerunning for this head.

daniil-shumko:
[Sesori reply] Reviewed the latest approval: no further findings. After #1478 merged, GitHub retargeted this PR to main. Merged current main without rewriting published history and resolved the squash-related conflicts while preserving the independently merged attachment-button/test updates. The cancellation diff is unchanged at 673 lines (570 authored, 103 generated). All 120 mobile body tests and both affected UI/mobile analyzers pass after the merge. CI will reassess the new merge head.

daniil-shumko:
[Sesori reply] Rebased onto latest main (55bf80145) as requested and pushed a0472dae3 with an exact-SHA force-with-lease. Dropped only the two design commits already squash-merged in #1478. Both cancellation commits are unchanged according to git range-diff, and the complete tree delta exactly matches the three incoming main files. No manual source conflict resolution was needed; the PR remains 673 changed lines. Existing focused test evidence is unchanged; CI will reassess the rebased head.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -1495,16 +1495,37 @@ class SessionDetailCubit(
...
   Future<void> cancelBridgeQueuedPrompt({required String promptId}) async {
     if (_refuseWhenInteractionBlocked(action: "cancel a bridge-queued prompt")) return;
     final result = await _sessionRepository.cancelQueuedPrompt(sessionId: _sessionId, promptId: promptId);
-    // Only not-found means the entry is gone (dispatched or removed
-    // elsewhere); any other failure proves nothing about the bridge queue.
+    if (isClosed) return;
     if (result case ErrorResponse(:final error)) {
-      if (error is! NonSuccessCodeError || error.errorCode != 404) return;
+      logw("Queued prompt cancellation was not confirmed for $_sessionId/$promptId", error);
+      _noticeStream.add(const SessionDetailQueueCancellationFailed());
```
AlexDochioiu: **Change before merge: a double tap on trash now surfaces a false warning.**

The trash button stays on the row until the success response comes back, and `session_detail_composer_controls.dart:115` calls this method with no in-flight guard. A second tap during that round trip reaches the bridge after the first cancel already removed the entry, gets a 404, and this line turns it into "Cancellation was not confirmed. The message may already have been sent." while the first tap actually succeeded. Before this PR the second 404 was silent, so this is a new regression and easy to hit on a laggy relay.

Smallest fix: keep a `Set<String>` of in-flight bridge cancellations in the cubit, return early when the id is already present, and remove it in a `finally`.

client/module_app_ui/lib/src/features/session_detail/widgets/session_detail_prompt_queue.dart
```diff
@@ -53,7 +75,20 @@ class const SessionDetailPromptQueue({
...
+                QueuedPromptDispatchState.dispatched => (
+                  label: loc.sessionDetailSendingMessage,
+                  indicator: const SizedBox.square(dimension: 14, child: PregoActivityIndicator(color: null)),
+                ),
+                QueuedPromptDispatchState.unknown => (
+                  label: loc.sessionDetailQueueCancellationUnavailable,
+                  indicator: Icon(TablerRegular.info_circle, size: 20, color: context.prego.colors.textTertiary),
+                ),
+              },
+              onRemove: prompt.dispatchState == QueuedPromptDispatchState.queued && onCancelBridge != null
```
AlexDochioiu: **Change before merge (product call): `unknown` removes cancellation for every public bridge.**

The queued-prompt contract shipped in v1.8.0 through v1.8.3, so a 1.8.4 app against any of those bridges renders no trash at all until the user updates the bridge. Those bridges cancel bridge-pending prompts correctly and already emit queue updates on cancel and on echo; they only lack the per-row dispatch state.

With the refusal-is-not-success fix in this PR, offering trash on `unknown` rows is honest: a row that was already dispatched yields the failure notice instead of vanishing, which is exactly the outcome this PR wants. Suggest `onRemove` for `dispatchState != dispatched`, render `unknown` like `queued`, and drop `sessionDetailQueueCancellationUnavailable` plus the info-icon branch. Judgment call, but the current behaviour takes a working feature away from the whole installed base of older bridges.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
```diff
@@ -1495,16 +1495,37 @@ class SessionDetailCubit(
...
+      if (error is NonSuccessCodeError && error.errorCode == 404) {
```
AlexDochioiu: Optional simplification: the reconciliation read after a 404, with the list-equality race guard and its four tests, is redundant in every flow I can trace. New bridges emit the dispatch queue event before the 404 can even be produced, older bridges carry no dispatch state so the read returns nothing new, and removal by another client already arrives over SSE. Only a lost SSE event benefits, and the next queue event or reconnect snapshot self-corrects, which AGENTS.md says to accept. Log, notice, return would be enough.

AlexDochioiu (commented):
Reviewed the full diff against `main` at the rebased head, plus the plugin, cubit and widget code around it.

**Verdict: approve once the two inline items marked "change before merge" are addressed.**

Verified:
- Pi, Claude and ACP each flip their existing dispatch flag synchronously before the backend write and now emit a queue update at that exact moment. Their cancel paths already refused entries past that flag, so the Sending row and the refusal finally agree. The emitted events are built from the same derived list, so they carry the correct state.
- The wire field is genuinely additive in both directions. The queued-prompt contract shipped publicly in v1.8.0 through v1.8.3, so the `unknown` default is required, and the v1.8.3 generated decoder ignores unrecognised keys, so a newer bridge does not break an older app.
- The delayed-read race from the first review round is covered by tests. Layering is clean and no backend concepts leak past the plugins.

Inline: one new UX regression (double tap), one product call on `unknown` rows, one optional simplification, one nit.

bridge/sesori_plugin_pi/lib/src/services/pi_session_service.dart
```diff
@@ -192,7 +192,12 @@ final class PiSessionService({
     if (state == null) return const [];
     return [
       for (final turn in state.turns)
-        if (turn is _PiQueuedPromptTurn && turn.queueState == _PiQueueState.visible) turn.presentation,
+        if (turn is _PiQueuedPromptTurn && turn.queueState == _PiQueueState.visible)
+          turn.presentation.copyWith(
```
AlexDochioiu: Nit: the presentation is stored with `dispatchState: queued` at creation (line 336) and overridden on every read here, so `turn.presentation.dispatchState` is never authoritative and a future reader of the stored value gets a stale `queued`. ACP derives it in a getter on `_QueuedAcpPrompt`; mirroring that here would keep one source of truth.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
daniil-shumko: [Sesori reply] Fixed in a83053c2a. The cubit now suppresses a repeated cancellation of the same prompt while its first request is pending, using one ID set released in finally. The laggy-response regression reproduced two requests before the fix and now proves one request with no false notice. Retry is covered after both returned errors and thrown repository failures.

client/module_app_ui/lib/src/features/session_detail/widgets/session_detail_prompt_queue.dart
daniil-shumko: [Sesori reply] Agreed and fixed in a83053c2a. I verified the public v1.8.0–v1.8.3 release metadata and v1.8.3 cancellation implementation. Unknown remains the honest wire default, but now retains best-effort trash like queued; only explicit dispatched state suppresses cancellation. Refusals still show the notice without deleting the row. Removed the info branch and unused localization via flutter gen-l10n. Mobile coverage proves legacy/queued actions work and dispatched remains read-only.

client/module_core/lib/src/cubits/session_detail/session_detail_cubit.dart
daniil-shumko: [Sesori reply] Simplified in a83053c2a. Removed the cancellation-triggered GET, list-equality guard, and read-specific tests. Refusal now logs, emits the notice, and returns; existing queue events and reconnect snapshots remain authoritative. Replacement tests cover later dispatch/removal events and verify no extra refresh. A lost event may remain stale until the next event/reconnect, as discussed.

bridge/sesori_plugin_pi/lib/src/services/pi_session_service.dart
daniil-shumko: [Sesori reply] Fixed in a83053c2a. The turn now stores only its immutable display metadata; presentation is a getter derived from the existing promptDispatched flag. No stale queued presentation is stored and queuedPrompts reads the getter directly. All 58 Pi session-service tests pass, including pre-dispatch cancellation and native steering ownership.

daniil-shumko:
[Sesori reply] All four review items are addressed in a83053c2a: repeated cancellation taps are coalesced, public older bridges retain best-effort cancellation, the redundant 404 refresh is removed, and Pi presentation derives from its current turn state. Both required-item regressions failed before the fixes. Current validation: 41 core queue tests, 58 Pi session-service tests, all 120 mobile body tests, and all four affected analyzers pass. The full PR is 703 changed lines (610 authored, 93 generated). CI is reassessing this head; immediate steering is unchanged.

client/module_app_ui/lib/src/features/session_detail/widgets/session_detail_prompt_queue.dart
```diff
@@ -53,7 +75,16 @@ class const SessionDetailPromptQueue({
...
               unavailable: false,
-              onRemove: onCancelBridge == null ? null : () => onCancelBridge?.call(prompt.id),
+              deliveryStatus: switch (prompt.dispatchState) {
+                QueuedPromptDispatchState.queued || QueuedPromptDispatchState.unknown => null,
+                QueuedPromptDispatchState.dispatched => (
+                  label: loc.sessionDetailSendingMessage,
+                  indicator: const SizedBox.square(dimension: 14, child: PregoActivityIndicator(color: null)),
+                ),
+              },
+              onRemove: prompt.dispatchState != QueuedPromptDispatchState.dispatched && onCancelBridge != null
```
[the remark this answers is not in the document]
daniil-shumko: [Sesori reply] Retaining best-effort cancellation for unknown is intentional, as requested in the human review: https://github.com/sesori-ai/sesori_apps_monorepo/pull/1479#discussion_r4012893678. Public v1.8.0–v1.8.3 bridges support pre-dispatch cancellation despite omitting this metadata. The false-success claim does not apply to the current cubit: every ErrorResponse, including 404, emits the failure notice and returns before row removal; only a successful cancellation removes the row. Those refusal and legacy-action paths are covered by passing core/mobile tests. The stale model comment was valid feedback and is corrected in 91bb902fa; the PR body already describes the updated legacy policy. No runtime behavior was changed in this follow-up.

daniil-shumko:
[Sesori reply] Assessed the new review finding. Kept the human-requested best-effort legacy cancellation behavior: failed responses cannot remove the row as cancelled. Fixed the genuinely stale model documentation in 91bb902fa. This follow-up changes only documentation; runtime code and generated output are unchanged from the passing 41 core / 58 Pi / 120 mobile tests and four analyzers. Fresh-head CI will reassess normally.

